PR #9, opened 2016-10-31
Add toString method to some datatypes

kabl:


src/main/java/org/web3j/abi/datatypes/NumericType.java
```diff
@@ -43,4 +47,9 @@ public int hashCode() {
         result = 31 * result + (value != null ? value.hashCode() : 0);
         return result;
     }
+
+    @Override
```
conor10: We need more specific Numeric type toString implementations - there are fixed point types, as well as signed versus unsigned that need to be taken into account.

src/main/java/org/web3j/abi/datatypes/Utf8String.java
```diff
@@ -1,5 +1,7 @@
 package org.web3j.abi.datatypes;
 
+import static java.util.Objects.requireNonNull;
```
conor10: Please can you add requireNonNull no all types so it's consistent?

PR #13, opened 2016-11-11
execute transaction without waiting for it to be mined.

MaxBinnewies:
Hi Conor,

first of all, I have to say that I am not used to working with github or gradle and dont know the entire web3j code. I couldn't get the fork to properly work in my eclipse workspace, so unfortunately I couldn't test it either. So this pull request is more a crude suggestion, than a proper contribution, please DONT merge it blindly without checking it, sorry about that. 

I am playing around with the live network through infura and it takes forever for transactions to get mined (like several minutes). That itself is not really a problem for me, but I dont want to wait that long to get the transaction hash. Right now Contract.executeTransactionAsync(Function function) seems to be doing two things: (1) it sends the transaction async and waits for the response from the node in the form of a transaction hash and (2) it waits for the transaction to get mined by repeatedly requesting a transaction receipt for that transaction hash.
Basically, I am looking for a way to only do (1). I cant override the function properly, because its using private fields (GAS_PRICE, GAS_LIMIT, and contractAddress). I have already figured out how to do (2) on my own and want to check for the transaction receipt at my own/the users discretion.

Do you think something like that is possible?
Max

### DIFF
--- a/src/main/java/org/web3j/abi/Contract.java
+++ b/src/main/java/org/web3j/abi/Contract.java
@@ -129,6 +129,35 @@ public abstract class Contract extends ManagedTransaction {
 
         return signAndSend(rawTransaction);
     }
+    
+    /**
+     * Given the duration required to execute a transaction, asyncronous execution is strongly
+     * recommended via {@link Contract#executeTransactionAsync}.
+     * This executes the transaction and waits for the transaction to be processed by the node, but does not wait for the transaction receipt (i.e. the transaction getting mined) 
+     *
+     * @param function to transact with
+     * @return Transaction hash
+     * @throws ExecutionException if the computation threw an
+     * exception
+     * @throws InterruptedException if the current thread was interrupted
+     * while waiting
+     * @throws TransactionTimeoutException if the transaction was not processed while waiting
+     */
+    protected TransactionReceipt executeTransactionWithoutWaiting(
+            Function function) throws ExecutionException, InterruptedException,
+            TransactionTimeoutException {
+        BigInteger nonce = getNonce(credentials.getAddress());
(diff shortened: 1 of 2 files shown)


MaxBinnewies:
The long mining time comes from the gas limit of 2million. I have cloned the methods locally for now and set it to 600000 -> mining time went down from about 15minutes to about 30 seconds.

conor10:
Agree it makes sense to be able to specify gas price & limit parameters for generated smart contracts. I'm thinking they should be additional constructor parameters in the smart contract create/deploy methods.

Regarding the transaction receipt - I provided the async methods so that people could carry on with stuff & get notified when the transaction completes. However, if you want to work with the transaction receipt, I'll have a think on how best to do this, as I don't want to risk any further complication to the API depending on if you require the transaction receipt or a future.

MaxBinnewies:
I agree that being able to choose the gas limit seems to be the core issue here. Because with 2million it takes a really long time for transactions to get mined on the main network, mostly so long that the method runs into its timeout and I get neither transaction receipt nor transaction hash. With my custom limit (600000 atm, which is still pretty high) it is a lot quicker.

My current approach is more of a pull-concept. The user has a dashboard with all relevant information and can refresh it through a button click. But I am not sure how relevant this would be to other developers. I actually extended Contract with a CustomContract-class which basically contains the methods in this pull request. So I am fine with that for now, feel free to reject the transaction hash idea if you dont think its relevant.

conor10:
I've pushed out release 1.0.6 which enhances the smart contract API to require gas limits & gas prices at construction.

For the time being I hope that fixes your issues? I'm not keen currently on breaking down the smart contract API further - although I may start making transaction nonce & transaction hashes available in the event of failure with requests.

PR #23, opened 2016-11-29
Allow leading zeros when parsing hex strings.

vethan:
Currently Numeric.decodeQuanitity will fail if called with a hex value with a leading zero, such as "0x01".  This doesn't make sense, leading zeroes are valid for hex numbers, and it's fairly common for a hex number to be padded. Check remove, and tests changed

### DIFF
--- a/src/main/java/org/web3j/utils/Numeric.java
+++ b/src/main/java/org/web3j/utils/Numeric.java
@@ -51,10 +51,6 @@ public final class Numeric {
             return false;
         }
 
-        if (value.length() > 3 && value.charAt(2) == '0') {
-            return false;
-        }
-
         return true;
     }
 
--- a/src/test/java/org/web3j/utils/NumericTest.java
+++ b/src/test/java/org/web3j/utils/NumericTest.java
@@ -44,14 +44,15 @@ public class NumericTest {
                 equalTo(new BigInteger("204516877000845695339750056077105398031")));
     }
 
-    @Test(expected = MessageDecodingException.class)
-    public void testQuantityDecodeMissingPrefix() {
-        Numeric.decodeQuantity("ff");
+    @Test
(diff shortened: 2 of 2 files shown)


conor10:
Thanks for the submission. My issue with supporting zero padded quantity values is that it goes against the [JSON-RPC specification](https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding). What's the specific use case you need this functionality for?

vethan:
Ah. Interesting, this means that testrpc isn't obeying the specification. Even so, I wouldn't expect an exception on parsing, I'd expect an error  but for execution to continue

vethan:
Also, thanks for the link to the specification I'd been looking for something like that to figure out which of testrpc and web3j was off spec. I'll update my testrpc pull request.

conor10:
I'd rather fail if an invalid value is provided in the payload.

Would you be able to reference your testrpc pull request here please? I'd like to keep this change out of web3j. However, if it seems like it may take a while for the pull request to be accepted, happy to reconsider.

vethan:
So, gonna have to re-do my testrpc PR. What seemed like a quick and easy change over there is having unforseen consequences because they don't differentiate between treating buffers as addresses and treating them as values >_> 

Give me the weekend to investigate this in more detail

vethan:
@conor10 So, I can't link my PR because my work so far still isn't a 100% fix, it's gonna take me a little while to get to grips with their codebase enough to do the fix properly. Opened issue /ethereumjs/testrpc#220

conor10:
Hey @vethan are you still looking into submitting a pull request to testrpc or have you put it on hold?

vethan:
@conor10 unfortunately I've had to put it on hold for now, not sure when I'll have time to get around to it properly

conor10:
Merged - makes sense to have this working with testrpc.

PR #28, opened 2016-12-02
more robustness to Protocol changes

h2mch:
I had the issue, that the Transaction Receipt had an attribute in the json object called logsBloom. Therefore the object could not be parsed to the Java Class.
Do be a bit more robust, I think we can ignore such mapping errors.


conor10:
Logs bloom should already be supported, was this using an older version of web3j?

h2mch:
Yes, it was in version 1.0.6. I was "too late" to send you my logs blooms extension (or you too fast). But for other changes in the future, i think it is still better to be a bit more robust by the json -> java parsing.

PR #39, opened 2017-01-09
Fix bug where boolean values are decoded incorrectly

ermyas:
[`TypeDecoder.decode()`](https://github.com/web3j/web3j/blob/master/src/main/java/org/web3j/abi/TypeDecoder.java#L42) ignores the offset value specified for Boolean
types. This leads to incorrect decoding by the [`TypeDecoder.decodeBool()`](https://github.com/web3j/web3j/blob/master/src/main/java/org/web3j/abi/TypeDecoder.java#L128) function,
which attempts to resolve the entire HEX input string instead of the portion of the
string identifying the boolean value.

e.g. In the HEX input given below, where the second value is the desired boolean value to decode, [`TypeDecoder.decode()`](https://github.com/web3j/web3j/blob/master/src/main/java/org/web3j/abi/TypeDecoder.java#L42) attempts to resolve the entire input as a Boolean value instead of the
64 characters starting at offset 64. See [`TestDecoderTest.testBoolDecodeGivenOffset()`](https://github.com/web3j/web3j/pull/39/files#diff-652add185382ec214d38ee44d802daf3R29)

**input:** 
_"0000000000000000000000000000000000000000000000007fffffffffffffff
0000000000000000000000000000000000000000000000000000000000000001
0000000000000000000000000000000000000000000000007fffffffffffffff"_
**expected:** true
**result:** false

This PR is an attempt at fixing this bug

### DIFF
--- a/src/main/java/org/web3j/abi/TypeDecoder.java
+++ b/src/main/java/org/web3j/abi/TypeDecoder.java
@@ -39,7 +39,7 @@ class TypeDecoder {
         if (NumericType.class.isAssignableFrom(type)) {
             return (T) decodeNumeric(input.substring(offset), (Class<NumericType>) type);
         } else if (Bool.class.isAssignableFrom(type)) {
-            return (T) decodeBool(input);
+            return (T) decodeBool(input.substring(offset, offset + MAX_BYTE_LENGTH_FOR_HEX_STRING));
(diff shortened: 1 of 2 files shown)


conor10:
Thanks for highlighting - I'll make the updates I suggested before I release.

PR #42, opened 2017-01-11
Remove blank spaces in each element of the list passed to the Collection#join

fooock:
The blank spaces are removed from the element of the list in each iteration. Included test


conor10:
Thanks.

PR #45, opened 2017-01-14
Change return type of generated wrapper methods to CompletableFuture<T>

ermyas:
Hi @conor10 

This PR proposes returning `CompletableFuture<T>`s, instead of `Future<T>`, from generated solidity wrapper methods. This would offer better flexibility to the calling client, in terms of composability, pipelining, transforming etc. of asynchronous operations.
Given that the underlying transaction execution methods in the super class `Contract` already use `CompletableFuture<T>`s the change in this PR is primarily to the generator, and associated tests.

Thanks!

### DIFF
--- a/src/main/java/org/web3j/abi/Contract.java
+++ b/src/main/java/org/web3j/abi/Contract.java
@@ -7,8 +7,6 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Future;
-
 import org.web3j.abi.datatypes.Event;
 import org.web3j.abi.datatypes.Function;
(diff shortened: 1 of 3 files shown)


conor10:
Thanks for taking the time to submit this pull request.

The reason I didn't use CompletableFutures in the code generator was to keep the generated code Android compatible. For the time being I don't intend to change that, but longer term depending on how much traction the Android port has that may change.

ermyas:
Sure, no worries.
From what I understand, `CompletableFuture` is available from [Android API Level 24](https://developer.android.com/reference/java/util/concurrent/CompletableFuture.html) onwards, but I suppose this would be a problem for earlier versions which would require a separate compatibility library, [Android streamssupport](https://developer.android.com/reference/java/util/concurrent/CompletableFuture.html).

PR #46, opened 2017-01-15
Replace Java internal API and make file seprator OS-specific

matthiaszimmermann:
noticed that path expressions for wallet files on windows include "/" as separators instead of "\". while fixing this, eclipse hinted that the use of ParameterizedTypeImpl is discouraged as this is considered to be Java internal API.


conor10:
Thanks Matthias.

matthiaszimmermann:
you're welcome and tx for the great web3j library :-)

PR #52, opened 2017-02-02
a High Throughput TransactionManager (first commit)

ferOnti:
This is the first version of the HighThroughput TransactionManager,
Example:

			//create an instance of your smart contract wrapper
			MySmartContract contract;

                        //use the loadHT instead the load 
			contract = EPCISRegistry.loadHT(contractAddress, web3, credentials, gasPrice, gasLimit);
			//set the attempts to 0, because we don't want to wait for receipts 
                        contract.setAttempts(0);

			Future<TransactionReceipt> futureReceipt = contract.myfunction(param1, param2);

			HighThroughputTransactionManager transactionManager = (HighThroughputTransactionManager)contract.getTransactionManager();
			TransactionReceipt tr = futureReceipt.get();
			transactionHash = tr.getTransactionHash();
                        //currently the function futureReceipt.get is returning a fake receipt with only the transaction hash
                        // todo: save this transactionHash and then process receipts for all transactions sent by HT
			log.info(transactionManager.getCurrentNonce());

### DIFF
--- a/src/main/java/org/web3j/tx/HighThroughputTransactionManager.java
+++ b/src/main/java/org/web3j/tx/HighThroughputTransactionManager.java
@@ -1,39 +1,70 @@
 package org.web3j.tx;
 
 import org.web3j.crypto.Credentials;
-import org.web3j.crypto.Hash;
-import org.web3j.crypto.TransactionEncoder;
 import org.web3j.protocol.Web3j;
 import org.web3j.protocol.core.DefaultBlockParameterName;
 import org.web3j.protocol.core.methods.request.RawTransaction;
-import org.web3j.protocol.core.methods.response.EthGetTransactionCount;
+import org.web3j.protocol.core.methods.response.EthBlock;
 import org.web3j.protocol.core.methods.response.EthSendTransaction;
-import org.web3j.protocol.exceptions.TransactionTimeoutException;
-import org.web3j.utils.Numeric;
+import org.web3j.protocol.core.methods.response.TransactionReceipt;
+import rx.Subscription;
 
+import java.io.IOException;
 import java.math.BigInteger;
 import java.util.HashMap;
 import java.util.concurrent.ExecutionException;
 
 /**
- * Created by fernando on 2/1/17.
+ * High Throughput transaction manager.
  */
+
 public class HighThroughputTransactionManager extends RawTransactionManager {
 
-	private static HashMap<String,HighThroughputTransactionManager> map = new HashMap<>();
+	//this static hashmap stores web3j objects and their accounts
+	// ---> web3j ---> account ---> HQtransactionManager
+	private static HashMap<Web3j, HashMap<String, HighThroughputTransactionManager>> map = new HashMap<>();
+
 	private String accountAddress;
 	private BigInteger nonce = BigInteger.valueOf(-1);
-	private HashMap<BigInteger, Object> transactions;
 	private Web3j web3j;
 	private byte chainId;
 
+	//this private hashmap stores receipts for each block
+	// ---> block number ---> txindex ---> TransactionReceipt
+	private HashMap<BigInteger, HashMap<Long, TransactionReceipt>> transactions;
+
 	public static HighThroughputTransactionManager getInstance(Web3j web3j, Credentials credentials)
 	{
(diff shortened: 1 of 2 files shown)


src/main/java/org/web3j/tx/HighThroughputTransactionManager.java
```diff
@@ -0,0 +1,95 @@
...
+import org.web3j.protocol.core.methods.response.EthSendTransaction;
+import org.web3j.protocol.exceptions.TransactionTimeoutException;
+import org.web3j.utils.Numeric;
+
+import java.math.BigInteger;
+import java.util.HashMap;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * Created by fernando on 2/1/17.
```
conor10: Please can you use the Javadoc header to include a simple class description? Your name will be visible from the commit log, so it's not required here.

src/main/java/org/web3j/tx/HighThroughputTransactionManager.java
```diff
@@ -0,0 +1,95 @@
...
+public class HighThroughputTransactionManager extends RawTransactionManager {
+
+	private static HashMap<String,HighThroughputTransactionManager> map = new HashMap<>();
+	private String accountAddress;
+	private BigInteger nonce = BigInteger.valueOf(-1);
+	private HashMap<BigInteger, Object> transactions;
+	private Web3j web3j;
+	private byte chainId;
+
+	public static HighThroughputTransactionManager getInstance(Web3j web3j, Credentials credentials)
```
conor10: I don't see why we need this map for storing instances of the transaction manager. Can't individual instances for credentials be defined & then shared in the client code? I'm not clear on what this advantage of having this is in web3j.

conor10:
Thanks for taking the time to submit this - please see the inline comments.

src/main/java/org/web3j/tx/ManagedTransaction.java
```diff
@@ -75,8 +81,13 @@ protected TransactionReceipt processResponse(EthSendTransaction transactionRespo
         }
 
         String transactionHash = transactionResponse.getTransactionHash();
-
-        return waitForTransactionReceipt(transactionHash);
+        if (attempts > 0) {
```
conor10: What sort of transaction volumes per minute would you like this to support? Given the current limitations in Ethereum on transaction throughput, I think the polling is a relatively unobtrusive operation.

src/main/java/org/web3j/tx/HighThroughputTransactionManager.java
ferOnti: Ok, this and more change will be submit in next PR

src/main/java/org/web3j/tx/HighThroughputTransactionManager.java
ferOnti: Let me show you in next Pull Request, how this evolved.

src/main/java/org/web3j/tx/ManagedTransaction.java
ferOnti: In Ethereum I reached 60+ transactions per minute for some simple smart contract, with Quorum I executed 100-150 transactions per second.  Before I tried with web3j, and I disabled the polling for each transaction because I never reached more than 25 transactions per minute, my laptop was frozen because the polling made all network connections busy and the system refused to accept new connections until I closed nodejs.

tramonex-nate: Rather than polling, what about subscribing to the block observable and checking each block to see if it contains the transaction receipt.

ferOnti: exactly I'm doing now, please wait until my next PR, probably in a couple of hours

tramonex-nate: 👍 Looking forward to it

src/main/java/org/web3j/tx/HighThroughputTransactionManager.java
ferOnti: I was thinking on this topic, if we need to put all this code inside the web3j library, so I will decouple a little bit more.  An external service in the client side will manage connections and queues of transactions and logs processed.

PR #56, opened 2017-02-09
High throughput 

ferOnti:
this is the second commit with changes for the HQTransactionManager.

Example:

		//create an instance of your smart contract wrapper
		MySmartContract contract;

            //use the loadHT function instead the load 
		contract = MyContract.loadHT(contractAddress, web3, credentials, gasPrice, gasLimit);
	       
             //call the transaction  
         	Future<TransactionReceipt> futureReceipt = contract.myfunction(param1, param2);

            //get the TransactionHash from the TransactionReceipt
		TransactionReceipt tr = futureReceipt.get();
		transactionHash = tr.getTransactionHash();
    
             //the function futureReceipt.get is returning an incomplete receipt with only the transaction hash
            // Clients should take care of this transaction hash
            //  because the polling is disabled for  HQTransactionManager


src/main/java/org/web3j/tx/HighThroughputTransactionManager.java
```diff
@@ -0,0 +1,127 @@
...
+import java.io.IOException;
+import java.math.BigInteger;
+import java.util.HashMap;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * High Throughput transaction manager.
+ */
+
+public class HighThroughputTransactionManager extends RawTransactionManager {
```
conor10: Please could you include an integration tests that demonstrates how to use this class - I'm still not completely clear on the advantages of this approach. An integration test will help clarify.

See https://github.com/web3j/web3j/blob/master/src/integration-test/java/org/web3j/protocol/scenarios/FastRawTransactionManagerIT.java for an example.

src/main/java/org/web3j/tx/HighThroughputTransactionManager.java
```diff
@@ -0,0 +1,127 @@
...
+			} else {
+				instance = new HighThroughputTransactionManager(web3j, credentials, chainId);
+			}
+			entryWeb3j.put(credentials.getAddress(), instance);
+
+			return instance;
+		}
+	}
+
+	public HighThroughputTransactionManager(Web3j web3j, Credentials credentials, byte chainId) {
```
conor10: Is it necessary to duplicate a lot of the functionality that is in the FasRawTransactionManager into this class? I can't help but think this class should be split in two to make the purpose clearer.

src/main/java/org/web3j/tx/ManagedTransaction.java
```diff
@@ -75,8 +81,15 @@ protected TransactionReceipt processResponse(EthSendTransaction transactionRespo
         }
 
         String transactionHash = transactionResponse.getTransactionHash();
-
-        return waitForTransactionReceipt(transactionHash);
+        //if attempts is zero, means the transaction will not wait for a receipt,
+        if (attempts > 0) {
+            return waitForTransactionReceipt(transactionHash);
+        } else {
+            //but returns an incomplete receipt with only the transactionHash in it
```
conor10: I think put this in a separate method "emptyTransactionReceipt" it makes the intention clearer.

src/main/java/org/web3j/tx/HighThroughputTransactionManager.java
```diff
@@ -0,0 +1,127 @@
...
+	private BigInteger nonce = BigInteger.valueOf(-1);
+	private Web3j web3j;
+	private byte chainId;
+
+	//this private hashmap stores receipts for each block
+	// ---> block number ---> txindex ---> TransactionReceipt
+	private HashMap<BigInteger, HashMap<Long, TransactionReceipt>> transactions;
+
+	public static HighThroughputTransactionManager getInstance(Web3j web3j, Credentials credentials)
+	{
```
conor10: Please can you ensure code style is consistent with the rest of web3j - { should be on the same line as the method declaration. I will be getting checkstyle rules set up shortly so this will be less of an issue.

src/main/java/org/web3j/tx/HighThroughputTransactionManager.java
```diff
@@ -0,0 +1,127 @@
+package org.web3j.tx;
+
+import org.web3j.crypto.Credentials;
```
conor10: I'd be grateful if you can use the following import ordering for consistency.

screen shot 2017-02-10 at 12 31 50

src/main/java/org/web3j/tx/HighThroughputTransactionManager.java
conor10: Hey @ferOnti - do you need any further input from me on creating an integration test for the HighThroughputTransactionManager. I see you responded with a confused emoticon.

src/main/java/org/web3j/tx/HighThroughputTransactionManager.java
```diff
@@ -0,0 +1,120 @@
...
+	public synchronized void resetNonce(String address) throws InterruptedException, ExecutionException {
+		nonce = super.getNonce();
+	}
+
+	public synchronized void setNonce(BigInteger value) {
+		nonce = value;
+	}
+
+	@Override
+	public EthSendTransaction executeTransaction(
```
crazyquark: This might need to be synchronized

conor10:
Closing as no further updates have been made.

PR #71, opened 2017-03-16
Fix problems around size validation of large uint values.

eztierney:
I found two problems related to large uint values:
 - `ArrayIndexOutOfBoundException` when encoding a large uint value
 - No exception when creating an `IntType` with a >256 bit value

The code was using `BigInteger.bitCount()` but I believe it should be using `BigInteger.bitLength()`.  `bitCount` calculates the number of bits that differ from the sign bit, `bitLength` calculates the minimal number of bits needed to represent the number without the sign bit, which I think is what the intention was here.

I added some tests to TypeEncoderTests that demonstrate the 2 bugs.


conor10:
Awesome - thanks for picking this up.

PR #83, opened 2017-05-02
Add Missing hex prefix

h2mch:
According to the the documentation (https://docs.web3j.io/getting_started.html#transactions), a raw transaction should work like:
`String hexValue = Hex.toHexString(signedMessage);
EthSendTransaction ethSendTransaction = web3j.ethSendRawTransaction(hexValue).sendAsync().get();`

Unfortunately the `Hex.toHexString` produces a hex value without the the 0x prefix, which is needed.

To be a bit more robust, I added a check if the prefix exists. if not, the prefix is added


conor10:
Hi @h2mch, thanks for taking the time to submit this.

I'd rather keep the Requests implementations as simple as possible, and not be checking for edge cases. If people don't want to use web3j's smart contract wrappers, I'd expect them to follow the docs for creating transactions (appreciate there was an error which you spotted here).

Hence I'm closing this PR, please refer to my comments in https://github.com/web3j/web3j/issues/84 for the resoultion.

Thanks.

PR #97, opened 2017-05-23
Highlight more code blocks in README

ligi:


conor10:
Thanks @ligi

ligi:
You are welcome! Was just browsing the samples and it reads nicer this way - also better for OCD as some was highlighted and some was not ;-)

conor10:
OCD is definitely a healthy trait in coders :)

PR #106, opened 2017-06-19
FunctionReturnDecoder.decodeIndexedValue always returns `Bytes32`

jaycarey:
We have some solidity contracts with events containing `Bytes16` values. These generate the correct types in the contract wrappers, but when executed they always arrive as `Bytes32` types instead and result in a cast error. I've adjusted the method responsible slightly to generate the specific `BytesN` class required and added some unit tests. 

I had considered an integration test for this but, it seems more like something that should be covered by a unit test.


conor10:
Thanks!

Issue #100, opened 2017-06-10
Support event filters from block 0/earliest


conor10:
When a new event filter is created in web3j, block parameters can be provided which specify which block the filter should run from.

According to the [JSON-RPC new filter specification](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_newfilter), stating the earliest block does not start at block 0. Testing has confirmed this.

Apparently, web3.js does support the replay from block 0 on events, so it would be worthwhile providing this support in web3j too.

conor10:
This can be addressed via the following:

From @earizon on Gitter:

*To obtain the expected behaviour I applied next patch that just process all logs before starting the scheduler:*
```
diff --git a/src/main/java/org/web3j/protocol/core/filters/Filter.java b/src/main/java/org/web3j/protocol/core/filters/Filter.java
index cad9fba..961e7de 100644
--- a/src/main/java/org/web3j/protocol/core/filters/Filter.java
+++ b/src/main/java/org/web3j/protocol/core/filters/Filter.java
@@ -39,6 +39,8 @@ public abstract class Filter<T> {
             }

             filterId = ethFilter.getFilterId();
+            EthLog ethLogInit = web3j.ethGetFilterLogs(filterId).send();
+            process(ethLogInit.getLogs());

             schedule = scheduledExecutorService.scheduleAtFixedRate(() -> {
                 EthLog ethLog = null;
```

However, some of the methods in [JsonRpc2_0Rx](https://github.com/web3j/web3j/blob/9e25b7624cb5bd3d821e93acde60c9ef21f72faa/src/main/java/org/web3j/protocol/rx/JsonRpc2_0Rx.java#L96) will need to be updated accordingly.

PR #115, opened 2017-06-27
Fix https://github.com/web3j/web3j/issues/100. Ancient event logs not…

earizon:
Fix https://github.com/web3j/web3j/issues/100


conor10:
Thanks for submitting - it's a great inclusion.

PR #118, opened 2017-06-30
Adding value as an option for a Smart Contract transactional function call

ajberchek:
I needed to include ether as value when calling transactional functions from my generated smart contract. The only way to do that was to modify the auto-generated smart contract java file, then manually encode and send the transaction with a value attached so the smart contract on the blockchain would contain value. 

I decided modifying some transaction functions in Contract.java, and generating overloaded transactional functions in the smart contract wrapper would be a cleaner solution. It works, and I think it will be a good addition to the current features since there are many use cases that involve a smart contract handling ether. 

Please let me know if there are any problems and I will gladly fix them.

### DIFF
--- a/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
+++ b/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
@@ -134,7 +134,8 @@ public class SolidityFunctionWrapper {
 
         for (AbiDefinition functionDefinition : functionDefinitions) {
             if (functionDefinition.getType().equals("function")) {
-                methodSpecs.add(buildFunction(functionDefinition));
+                methodSpecs.addAll(buildFunction(functionDefinition));
+
 
             } else if (functionDefinition.getType().equals("event")) {
                 buildEventFunctions(functionDefinition, classBuilder);
@@ -291,8 +292,10 @@ public class SolidityFunctionWrapper {
         return result;
     }
 
-    static MethodSpec buildFunction(
+    static List<MethodSpec> buildFunction(
             AbiDefinition functionDefinition) throws ClassNotFoundException {
+        List<MethodSpec> overloadedMethods = new ArrayList<>();
+
         String functionName = functionDefinition.getName();
 
         MethodSpec.Builder methodBuilder =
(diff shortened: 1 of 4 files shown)


conor10:
Thanks for taking the time to do this - you're right, this functionality is definitely required in web3j.

May I suggest that instead of overloading the methods, you simply refer to the "payable" field in the abi file? If this is set, then we know that the method should take an Ether value.

e.g.
```json
   {  
      "constant":false,
      "inputs":[  
         {  
            "name":"_from",
            "type":"address"
         },
         {  
            "name":"_to",
            "type":"address"
         },
         {  
            "name":"_value",
            "type":"uint256"
         }
      ],
      "name":"transferFrom",
      "outputs":[  
         {  
            "name":"success",
            "type":"bool"
         }
      ],
      "payable":false,
      "type":"function"
   },
```

If you were able to update the constructor function generation to take this into account also, it would be much appreciated :)

ajberchek:
My initial thinking with the function overloading was to provide backwards compatibility with those users who have already implemented workarounds for the value sending. A default value of 0 in the function signature would also solve this issue without overloading, however Java doesn't allow default valued parameters; is there a Java equivalent to default parameters that could be used for this?

Should the backwards compatibility be preserved or should the overloaded function, without value as a parameter, be removed?

Either way, adding the value as a parameter based on the payable field would be a great improvement; I will make that code change.

As for the constructor function generation, I am a little lost, do you mean that if the constructor is payable, then the generated constructor should also take in value as a parameter?

conor10:
I wouldn't worry about people who've implemented workarounds. Other than overloaded methods I don't believe there's a more suitable approach for default values in this instance.

I think remove the function without value as a parameter - then the wrapper code stays true to the Solidity code. As long as the abstract Contract class doesn't change, backward compatibility will not be broken.

_As for the constructor function generation, I am a little lost, do you mean that if the constructor is payable, then the generated constructor should also take in value as a parameter?_ - yes

ajberchek:
The function without the value as a parameter for the overloaded functions has been removed. 

As for the constructor value, the deploy function already generates a BigInteger initialWeiValue parameter. Does this satisfy the payable constructor requirement?

conor10:
Yes it should do.

ajberchek:
Great, thank you! Can this be merged?

src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -291,8 +292,11 @@ static String createValidParamName(String name, int idx) {
         return result;
     }
 
-    static MethodSpec buildFunction(
+    static List<MethodSpec> buildFunction(
```
conor10: I don't see why returning a list is necessary. All that should be required is that the method is slightly different if payable is set, otherwise the logic should be identical to as per previous.

conor10:
@ajberchek, I'm merging the other PR - please can you refer to it, as it accomplishes the same functionality in fewer lines of code, and I'm keen to get a new release out.

I'm very grateful for you taking the time to work on this, and hope it won't put you off contributing more PRs - there's still plenty of items marked "help wanted" I'd be grateful for assistance on.

PR #128, opened 2017-07-08
JsonMappingException fix when compiling a solidity contract with eth_compileSolidity

vgorin:
fixes an exception
com.fasterxml.jackson.databind.JsonMappingException: Can not construct instance of org.web3j.protocol.core.methods.response.EthCompileSolidity$Code: no String-argument constructor/factory method to deserialize from String value ('0x6060604052346000575b60458060156000396000f3606060405260e060020a6000350463c6888fa18114601c575b6000565b346000576029600435603b565b60408051918252519081900360200190f35b600781025b91905056')

when compiling a solidity code
web3j.ethCompileSolidity("pragma solidity ^0.4.2; contract test { function multiply(uint a) returns(uint d) {   return a * 7;   } }").send()


conor10:
Thanks!

PR #134, opened 2017-07-14
Added isAddress() validation method

assafY:
While using the Web3j Numeric and Hash libraries to implement a isAddress() method in a local project, I have realised the method hexStringToByteArray() is not only unnecessary (you can just use the native .getBytes() on a string), but also returns an incorrect byte array.

To verify this, the isAddress() method in this PR fails if using the aforementioned method, but works if using the native .toBytes(). 

I have therefore removed the method content and replaced it simply with .getBytes(), and implemented the isAddress() method in a new class . This method is based on the web3.js implementation and is working perfectly in my own project.


assafY:
so I realise now many tests depend on the methods in Numeric.java and Hash.java to remain as they are. Therefore I have kept them as they are, and use the method in Hash.java that accepts a byte array directly.

conor10:
Hi @assafY, please can you provide a unit test that demonstrates this error you believe exists in hexStringToByteArray?

assafY:
Sure, but it would be useful if you told me first whether you intended for hexStringToByteArray to transform a String into a byte array in a similar way that String.getBytes() does? Because the two return different results for the same string, so if that was the intention I can quickly write a unit test that demonstrates this error.

conor10:
If you call String.getBytes() on a hex string, you get the ASCII values for each of the characters in the String. This is not the same as a byte array which is a binary encoded array of data. 

E.g. `Using getBytes`:
```java
"123".getBytes() == new byte[] { '1', '2', '3'} 
```
Or equivalently:
```java
new byte[] { (byte) 49, (byte) 50, (byte) 51 };
```

Whereas with a byte array (remember the 123 is a hex not decimal value here):
```java
hexStringToByteArray("123") == new byte[] { (byte) 1, (byte) 35 }
```

assafY:
I understand - so I am assuming that the checksum validation for newer Ethereum account addresses, as described [here](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md), uses the ASCII value byte representation rather than the one you implemented in hexStringToByteArray.

In this case I think the comparison between the two is irrelevant, and the checksum validation method I added which uses getBytes works independently. Checksum validation is now used by most major ether wallets and I assumed it would be a useful addition to the Web3j library.

clowdrgn:
Hi, I try to validate a address but cant pass.Maybe this line `String hash = Numeric.toHexString(Hash.sha3(address.replace("0x", "").getBytes()));`should toLowerCase when get address hash value ?and I can verify success when I  change to `String hash = Numeric.toHexStringNoPrefix(Hash.sha3(address.replace("0x", "").toLowerCase().getBytes()));` Or do I  miss some method ?

assafY:
That is correct, thanks! You do need `toLowerCase()` before validation. I added that now.

src/main/java/org/web3j/utils/Account.java
```diff
@@ -38,7 +38,7 @@ public static Boolean isAddress(String address) {
 
     private static Boolean validateChecksumAddress(String address) {
         String formattedAddress = address.replace("0x", "").toLowerCase();
-        String hash = Numeric.toHexString(Hash.sha3(formattedAddress.getBytes()));
+        String hash = Numeric.toHexStringNoPrefix(Hash.sha3(formattedAddress.getBytes()));
         for (int i = 0; i < 40; i++) {
             if (Character.isLetter(address.charAt(i))) {
```
Mawster: Why do you check from the address? This seems to be wrong - address still contains the leading '0x'

src/main/java/org/web3j/utils/Account.java
```diff
@@ -38,7 +38,7 @@ public static Boolean isAddress(String address) {
 
     private static Boolean validateChecksumAddress(String address) {
         String formattedAddress = address.replace("0x", "").toLowerCase();
```
Mawster: I have to validate eth addresses in a current project - I modified your version:

`String formattedAddress = address.replace("0x", "");
 String lowercaseAddress = formattedAddress.toLowerCase();
 String hash = Numeric.toHexStringNoPrefix(Hash.sha3(lowercaseAddress.getBytes()));
     for (int i = 0; i < 40; i++) {
         if (Character.isLetter(formattedAddress.charAt(i))) {
`

src/main/java/org/web3j/utils/Account.java
assafY: Sure, I missed that method. In this case you can also remove the `.replace("0x", "")` above, which I assumes is what happens in `toHexStringNoPrefix`?

assafY: Ah I misunderstood your question. This is actually the correct way to do it. The hash needs to be computed from the lowercase address without a prefix, but the validation of the hash should happen against the original, untouched address.

https://github.com/ethereum/EIPs/issues/55

Mawster: No I don't think that you can remove it - if you remove it it would be used in the sha3 hashing.

Mawster: compare with
https://ethereum.stackexchange.com/questions/1374/how-can-i-check-if-an-ethereum-address-is-valid

especially this line:
`address = address.replace('0x','');`
>

assafY: Oh you are correct - thank you for pointing that out. The checksum validation should be carried out on the address without the 0x prefix. Fixed that now, thanks!

Mawster: Now you are ussing the lower cased version also in the comparision - thats not right.

assafY: Slow day for me today isn't it? Fixed that now, thanks.

Mawster:
Some tests wouldn't be bad

assafY:
anything stopping this from being merged?

core/src/main/java/org/web3j/utils/Account.java
```diff
@@ -0,0 +1,55 @@
...
+    private static final Pattern lowerCaseAddrPattern = Pattern.compile("^(0x)?[0-9a-f]{40}$");
+    private static final Pattern upperCaseAddrPattern = Pattern.compile("^(0x)?[0-9A-F]{40}$");
+
+    /**
+     * Verify that a hex account string is a valid Ethereum address.
+     *
+     * @param address given address in HEX
+     * @return is this a valid address
+     */
+    public static Boolean isAddress(String address) {
```
conor10: WalletUtils.isValidAddress achieves similar functionality.

core/src/main/java/org/web3j/utils/Account.java
```diff
@@ -0,0 +1,55 @@
...
+                || upperCaseAddrPattern.matcher(address).find()) {
+            // if it's all small caps or caps return true
+            return true;
+        } else {
+            // if it is mixed caps it is a checksum address and needs to be validated
+            return validateChecksumAddress(address);
+        }
+    }
+
+    private static Boolean validateChecksumAddress(String address) {
```
conor10: Validating checksum addresses is useful functionality.

It might be better to have isChecksumAddress as a method on the Address type, and a toChecksumAddress that operates on an Address or String instance

conor10:
See comments. It's also useful to reference the original EIP for checksum addresses https://github.com/ethereum/EIPs/issues/55, and include the corresponding issue in your commits.

I'd recommend you create a brand new PR, as there are a lot of commits in this PR.

Thanks for taking the time to get this far.

Issue #111, opened 2017-06-22
Optionally provide raw JSON-RPC response in Response objects


conor10:
It would be useful to be able to store the JSON-RPC response used to create a Response object in a local variable such as *rawResponse*.

This should be a property that is configured in org.web3j.protocol.Service.

PeterJames12:
Hello, use this code Web3j web3 = Web3j.build(new HttpService());  // defaults to http://localhost:8545/
Credentials credentials = WalletUtils.loadCredentials("password", "/path/to/walletfile");
TransactionReceipt transactionReceipt = Transfer.sendFunds(
        web3, credentials, "0x...", BigDecimal.valueOf(1.0), Convert.Unit.ETHER);

why i have this exception com.fasterxml.jackson.core.JsonParseException: Unexpected character ('<' (code 60)): expected a valid value (number, String, array, object, 'true', 'false' or 'null')

Thanks

mawenpeng:
@PeterJames12 
You may try the latest released Ethereum. 
And you'd better open a new issue for this. It's irrelevant to current topic.

PeterJames12:
@mawenpeng does it have previous version?

PeterJames12:
@mawenpeng  actually i use this Web3j web3 = Web3j.build(new HttpService("https://etherscan.io/address/my address")) instead of this Web3j web3 = Web3j.build(new HttpService());  maybe it is a problem

PeterJames12:
i don't have a clue why i have this exception java.lang.RuntimeException: Error processing transaction request: insufficient funds for gas * price + value

it is my code 


        TransactionReceipt transactionReceipt = Transfer.sendFundsAsync(
                web3, credentials, "0x to address", wei, Convert.Unit.GWEI).get();

conor10:
PR merged from @craigwilliams84.

tmsrikanth:
Can you please help with an example of storing an json rpc request and response in an variable.
Regards,
Srikanth

PR #135, opened 2017-07-14
#111 Optionally provide raw JSON-RPC response in Response objects

craigwilliams84:
The populating of the rawResponse variable within Responses is configured in the constructor of Service implementations.


conor10:
Awesome - thanks @craigwilliams84.

PR #138, opened 2017-07-19
Support payable functions in wrapper generator

jorpic:
This adds ability to send some ether along with a function call.

The generated code for payable function accepts one extra argument `weiValue`. E.g. such a wrapper 
```
    public Future<TransactionReceipt> pay(Utf8String name, BigInteger weiValue) {
        Function function = new Function("pay", Arrays.<Type>asList(name), Collections.<TypeReference<?>>emptyList());
        return executeTransactionAsync(function, weiValue);
    }
```
is generated for this Sloidity function:

```
function pay(string name) payable public {}
```

Ref. https://ethereum.stackexchange.com/questions/21481


ajberchek:
This functionality is already being addressed in a PR submitted a few weeks ago - #118

PR #147, opened 2017-08-05
Updated JSON-RPC for Parity and added Geth calls

iikirilov:
Response to changes of Parity namespace as detailed in [this](https://github.com/paritytech/parity/pull/3143) PR. 

Implemented [parity_accounts Module](https://github.com/paritytech/parity/wiki/JSONRPC-parity_accounts-module).

Implemented Geth [personal Module](https://github.com/ethereum/go-ethereum/wiki/Management-APIs#personal).

Implemented tests for all Response/Request.


src/main/java/org/web3j/crypto/WalletFile.java
```diff
@@ -62,6 +63,43 @@ public void setVersion(int version) {
         this.version = version;
     }
 
+    @Override
+    public int hashCode() {
```
conor10: With the hashCode and equals methods, please can you use the same implementations as per https://github.com/web3j/web3j/blob/master/src/main/java/org/web3j/protocol/core/methods/response/TransactionReceipt.java#L159

src/main/java/org/web3j/protocol/admin/JsonRpc2_0Personal.java
```diff
@@ -0,0 +1,90 @@
...
+        return new Request<>(
+                "personal_unlockAccount",
+                attributes,
+                ID,
+                web3jService,
+                PersonalUnlockAccount.class);
+    }
+    
+    @Override
+    public Request<?, PersonalUnlockAccount> personalUnlockAccount(
```
conor10: You shouldn't need to duplicate the logic in this second method - you should be able to use a single implementation, and override the implementation. Example here https://github.com/web3j/web3j/blob/master/src/main/java/org/web3j/protocol/parity/JsonRpc2_0Parity.java#L134

src/main/java/org/web3j/protocol/parity/JsonRpc2_0Parity.java
```diff
@@ -9,207 +8,324 @@
...
+                "parity_allAccountsInfo",
+                Collections.<String>emptyList(),
+                ID,
+                web3jService,
+                ParityAllAccountsInfo.class);
+    }
+
+    @Override
+    public Request<?, BooleanResponse> parityChangePassword(
+            String accountId, String oldPass, String newPass) {
```
conor10: I would use oldPassword & newPassword here as the parameter names - no need to shorten them

src/main/java/org/web3j/protocol/parity/JsonRpc2_0Parity.java
```diff
@@ -9,207 +8,324 @@
...
+                "parity_changePassword",
+                Arrays.asList(accountId, oldPass, newPass),
+                ID,
+                web3jService,
+                BooleanResponse.class);
+    }
+
+    @Override
+    public Request<?, ParityDeriveAddress> parityDeriveAddressHash(
+            String accountId, String password, Map<String, Object> hashType, boolean toSave) {
```
conor10: You should create a class for hashType rather then using a map here. For example see https://github.com/web3j/web3j/blob/master/src/main/java/org/web3j/protocol/core/methods/request/Transaction.java

src/main/java/org/web3j/protocol/parity/JsonRpc2_0Parity.java
```diff
@@ -9,207 +8,324 @@
...
+                Arrays.asList(accountId, password, hashType, toSave),
+                ID,
+                web3jService,
+                ParityDeriveAddress.class);
+    }
+
+    @Override
+    public Request<?, ParityDeriveAddress> parityDeriveAddressIndex(
+            String accountId, String password,
+            List<Map<String, Object>> indexType, boolean toSave) {
```
conor10: HashType can be used here too

src/main/java/org/web3j/protocol/parity/JsonRpc2_0Parity.java
```diff
@@ -9,207 +8,324 @@
...
                 web3jService,
-                PersonalUnlockAccount.class);
+                BooleanResponse.class);
     }
 
     @Override
-    public Request<?, PersonalUnlockAccount> personalUnlockAccount(
-            String accountId, String password) {
+    public Request<?, BooleanResponse> paritySetAccountName(
+            String accountId, String newAccountName) {
```
conor10: These parameters should be "address" and "name".

src/main/java/org/web3j/protocol/parity/JsonRpc2_0Parity.java
```diff
@@ -9,207 +8,324 @@
...
                 Collections.<String>emptyList(),
                 ID,
                 web3jService,
-                PersonalRequestsToConfirm.class);
+                SignerRequestsToConfirm.class);
     }
 
     @Override
-    public Request<?, EthSendTransaction> personalConfirmRequest(
+    public Request<?, EthSendTransaction> signerConfirmRequest(
```
conor10: It would be better to use a custom type here instead of the existing Transaction type which has a lot of unused fields for this method.

conor10: On second thoughts - you can get rid of the signer module methods - given its only a subset of them anyway. I'd question how useful they are for web3j users.

src/test/java/org/web3j/protocol/admin/ResponseTest.java
```diff
@@ -0,0 +1,98 @@
...
+        buildResponse("{\n"
+                + "    \"jsonrpc\": \"2.0\",\n"
+                + "    \"id\": 1,\n"
+                + "    \"result\": \"0xf1aabd691c887ee5c98af871239534f194a51fdeb801b1601d77c45afa74dae67ddd81aa5bb8a54b7974ef5be10b55a8535b040883501f76d14cb74e05e5635d1c\"\n"
+                + "}");
+        //CHECKSTYLE:ON
+
+        PersonalSign personalSign = deserialiseResponse(PersonalSign.class);
+        //CHECKSTYLE:OFF
+        assertThat(personalSign.getSignedMessage(),is("0xf1aabd691c887ee5c98af871239534f194a51fdeb801b1601d77c45afa74dae67ddd81aa5bb8a54b7974ef5be10b55a8535b040883501f76d14cb74e05e5635d1c"));
```
conor10: the "is(...)" should be on a new line

iikirilov:
I've used netbeans to autogenerate equals and hashcode, hence the discrepancy.
Have tested with Parity but not Geth. Will get this all sorted and tested today!

conor10:
Great work!

PR #154, opened 2017-08-25
Fix web3j#152. Filters create requests based on type

the-wastl:
Not all filters can provide historic data. In fact only the LogFilter can. Thus all other filters (i.e. BlockFilter and PendingTransactionFilter) must use the ethGetFilterChanges method instead of ethGetFilterLogs

### DIFF
--- a/src/main/java/org/web3j/protocol/core/filters/BlockFilter.java
+++ b/src/main/java/org/web3j/protocol/core/filters/BlockFilter.java
@@ -3,6 +3,7 @@ package org.web3j.protocol.core.filters;
 import java.io.IOException;
 import java.math.BigInteger;
 import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 
 import org.web3j.protocol.Web3j;
@@ -37,9 +38,17 @@ public class BlockFilter extends Filter<String> {
         }
     }
 
+    /**
+     * Since the block filter does not support historic filters, the filterId is ignored
+     * and an empty optional is returned
+     * @param filterId
+     * Id of the filter for which the historic log should be retrieved
+     * @return
+     * Optional.empty()
+     */
     @Override
-    protected Request<?, EthLog> createFilterRequest(BigInteger filterId) {
-        return this.web3j.ethGetFilterChanges(filterId);
+    protected Optional<Request<?, EthLog>> getFilterLogs(BigInteger filterId) {
+        return Optional.empty();
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/web3j/protocol/core/filters/Filter.java
```diff
@@ -85,6 +86,8 @@ public void cancel() {
         }
     }
 
+    protected abstract Request<?, EthLog> createFilterRequest(BigInteger filterId);
```
conor10: Please can we rename this to `getExistingFilterLogs` or similar?

src/main/java/org/web3j/protocol/core/filters/Filter.java
```diff
@@ -39,7 +40,7 @@ public void run(ScheduledExecutorService scheduledExecutorService, long blockTim
             }
 
             filterId = ethFilter.getFilterId();
-            EthLog ethLogInit = web3j.ethGetFilterLogs(filterId).send();
+            EthLog ethLogInit = this.createFilterRequest(this.filterId).send();
```
conor10: I think it makes sense for this initial request to take place asynchronously too once we have the filter id. Appreciate the current implementation is down to me, but it would be great if you could update this too. E.g.:

```java
scheduledExecutorService.submit(() -> {
                EthLog ethLog = null;
                try {
                    ethLog = this.createFilterRequest(this.filterId).send();
                } catch (IOException e) {
                    throwException(e);
                }
                
                process(ethLog.getLogs());
            });
```

src/main/java/org/web3j/protocol/core/filters/BlockFilter.java
```diff
@@ -34,5 +36,10 @@ void process(List<EthLog.LogResult> logResults) {
             }
         }
     }
+
+    @Override
+    protected Request<?, EthLog> createFilterRequest(BigInteger filterId) {
+        return this.web3j.ethGetFilterChanges(filterId);
```
conor10: It would be nice if this could be a no-op, with a comment that getFilterChanges is not supported for BlockFilters.

src/main/java/org/web3j/protocol/core/filters/PendingTransactionFilter.java
```diff
@@ -34,5 +36,10 @@ void process(List<EthLog.LogResult> logResults) {
             }
         }
     }
+
+    @Override
+    protected Request<?, EthLog> createFilterRequest(BigInteger filterId) {
+        return this.web3j.ethGetFilterChanges(filterId);
```
conor10: It would be nice if this could be a no-op, with a comment that getFilterChanges is not supported for PendingTransactionFilters.

the-wastl:
All filters that do not support fetching a full log, now return an empty optional and the initial fetching works async now

conor10:
Awesome - thanks @the-wastl.

PR #180, opened 2017-09-29
Integrate Parity traces API

saulbein:
Integration of Parity traces API.

The API is very useful to track transaction calls to contracts or check for value received in internal contract transactions (value transfers).

To use: run Parity with `--tracing on` and add `traces` to JSON-RPC API list (will require resync of chain to build full DB).

**Any feedback / critique / complaints are welcome.**

If preferred, I can squash it all into a single commit as well.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/parity/JsonRpc2_0Parity.java
+++ b/core/src/main/java/org/web3j/protocol/parity/JsonRpc2_0Parity.java
@@ -1,25 +1,34 @@
 package org.web3j.protocol.parity;
 
+import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.web3j.crypto.WalletFile;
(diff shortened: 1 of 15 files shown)


saulbein:
Also, these changes may not be final, because I am not sure that I got all variants of the JSON types, due to the documentation being exceedingly sparse.

conor10:
Thanks for submitting. A squash would be preferred.

Please, could you create a response test using EqualsVerifier - see [here](https://github.com/web3j/web3j/blob/958ad9819e29a08cf0ef2b7263b64618df740dc2/core/src/test/java/org/web3j/protocol/core/EqualsVerifierResponseTest.java).

This will ensure that the project coverage remains around 80%. This PR brings it down to 75%. See the [coverage report](https://codecov.io/gh/web3j/web3j/pull/180/diff).

saulbein:
Well, improved the coverage, however I'm not sure how exactly to deal with the recursive structures (it works now, but should be checked), and if you want to keep the `toString` methods, since they are not covered.

conor10:
This is great - thanks for submitting.

Issue #167, opened 2017-09-05
Contract Functions returning a static array throwing class cast exception when using the wrapper generated code


zaf187:
I have a contract function;

    function getFields() constant returns (bytes32[11], bytes32[11]) {
        var fields =[   bytes32('ASDFASDF'),
                        bytes32('ZXCVZXCV'),
                        bytes32('DFGHDFGH'),
                        bytes32('CVBNCVBN'),
                        bytes32('RTJFTGHJ'),
                        bytes32('SDFGSERT'),
                        bytes32('XCBVNCBVN'),
                        bytes32('DFGHDRTH'),
                        bytes32('ERTYERTY'),
                        bytes32('CVBNCXVB'),
                        bytes32('KTYIKTYUJ')];
                        
        var data = [    someStruct.id,
                        someStruct.name,
                        someStruct.number,
                        someStruct.cost,
                        someStruct.wombat,
                        someStruct.zombie,
                        someStruct.cow,
                        someStruct.chicken,
                        someStruct.anotherStruct.goose,
                        someStruct.anotherStruct.monkey,
                        someStruct.anotherStruct.pig];
        
        return (fields, data);
    }

the java generated function for this is;

    public Future<List<Type>> getFields() {
        Function function = new Function("getFields", 
                Arrays.<Type>asList(), 
                Arrays.<TypeReference<?>>asList(new TypeReference<StaticArray<Bytes32>>() {}, new TypeReference<StaticArray<Bytes32>>() {}));
        return executeCallMultipleValueReturnAsync(function);
    }

There is a problem in the FunctionReturnDecoder;

    private static List<Type> build(
            String input, List<TypeReference<Type>> outputParameters) {
        List<Type> results = new ArrayList<>(outputParameters.size());

        int offset = 0;
        for (TypeReference<?> typeReference:outputParameters) {
            try {
                Class<Type> type = (Class<Type>) typeReference.getClassType();

                int hexStringDataOffset = getDataOffset(input, offset, type);

                Type result;
                if (DynamicArray.class.isAssignableFrom(type)) {
                    result = TypeDecoder.decodeDynamicArray(
                            input, hexStringDataOffset, typeReference);
                    offset += MAX_BYTE_LENGTH_FOR_HEX_STRING;
                } else if (StaticArray.class.isAssignableFrom(type)) {
                    int length = ((TypeReference.StaticArrayTypeReference) typeReference).getSize();
                    result = TypeDecoder.decodeStaticArray(
[...]

PR #183, opened 2017-10-09
Static Arrays

jaycarey:
Added new set of generated types (StaticArray#) to represent the size of static arrays up to length 32. Beyond that, the code will fall back to using the usual 'StaticArray' without the length suffix.

This should fix problems where the wrappers that call EVM methods that return static arrays throw invalid cast exceptions such as #167.


conor10:
Great stuff.

My only comment is that I think there should be an invariant that ensures the size of the list or array wrapped by the StaticArray*N* matches what is expected. There is something similar in the [IntType](https://github.com/web3j/web3j/blob/master/core/src/main/java/org/web3j/abi/datatypes/IntType.java)

conor10:
Awesome thanks.

PR #200, opened 2017-10-23
Refactor addHeaders() in HttpService allowing customize headers.

mawenpeng:
Users may need customized headers if they access Ethereum via a reverse proxy or a gateway.
The customized headers may be used for authentication or give some other info to the reverse proxy or gateway.

I refactored the code to allow users to customize headers.


conor10:
Thanks!

Issue #195, opened 2017-10-17
Would like to override the Sign.sign(byte[] transactionHash, BigInteger privateKey) method


eztierney:
I would like to override the behavior of the `sign` method in `org.web3k.crypto.Sign`(I'd like to have the signing done out of process), but there is no easy way to override the static method without copying most of the `Sign` and `TransactionEncoder` classes.

One possible solution would be to move the `sign` method to the `ECKeyPair` class, which could then be extended and overridden to change the behavior. Happy to put up a PR if this seems like an ok idea.

conor10:
Hey @eztierney, out of process signing would definitely be useful. Go for it!

ZhiyuHelloWorld:
👍

PR #203, opened 2017-10-23
Refactoring - move Sign.sign method into ECKeyPair so it can be overridden.

eztierney:
This will make it easier to override the default signing behavior, such as implementing an ECKeyPair that does out of process signing.

This is for issue #195


conor10:
Thanks @eztierney.

Issue #211, opened 2017-10-30
[Android] Unsubscribing from Observable<Log> returned by Web3j.ethLogObservable() causes an Exception


mochalovv:
When I unsubscribe from the `Observable<Log>` returned by `Web3j.ethLogObservable(new Filter(...))`, my app crashes with the following exception:
```
E/ActivityThread: Performing stop of activity that is not resumed: {ru.vmochalov.web3jtest/ru.vmochalov.web3jtest.MainActivity}
                  java.lang.RuntimeException: Performing stop of activity that is not resumed: {ru.vmochalov.web3jtest/ru.vmochalov.web3jtest.MainActivity}
                      at android.app.ActivityThread.performStopActivityInner(ActivityThread.java:3829)
                      at android.app.ActivityThread.handleStopActivity(ActivityThread.java:3917)
                      at android.app.ActivityThread.access$1200(ActivityThread.java:177)
                      at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1478)
                      at android.os.Handler.dispatchMessage(Handler.java:102)
                      at android.os.Looper.loop(Looper.java:145)
                      at android.app.ActivityThread.main(ActivityThread.java:5951)
                      at java.lang.reflect.Method.invoke(Native Method)
                      at java.lang.reflect.Method.invoke(Method.java:372)
                      at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1400)
                      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1195)
E/AndroidRuntime: FATAL EXCEPTION: RxIoScheduler-3
                  Process: ru.vmochalov.web3jtest, PID: 3217
                  java.lang.IllegalStateException: Fatal Exception thrown on Scheduler.Worker thread.
                      at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:59)
                      at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:422)
                      at java.util.concurrent.FutureTask.run(FutureTask.java:237)
                      at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:152)
                      at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:265)
[...]

PR #212, opened 2017-10-30
Removing the duplicated calls for installing and uninstalling the filter

mochalovv:
Investigating the #211 issue, I found out that `eth_uninstallFilter` method is called twice. The first call is successful, but the second call causes an exception as its result is false:
`{"jsonrpc":"2.0","id":1,"result":false}`.

This pull request fixes the issue.


conor10:
Thanks for spotting.

PR #220, opened 2017-11-06
Issue 217 - Configure the way hex numbers are encoded

4ntoine:
Added starting zero for values 0x1-0xf.

Actually no configuration was added since there are no any settings or configurations methods
Tested with TestRPC (not tested with Geth though).


conor10:
This is likely to be problematic with Geth.

See the following comment https://github.com/web3j/web3j/blob/master/utils/src/test/java/org/web3j/utils/NumericTest.java#L53.

PR #245, opened 2017-11-16
Improvement for UNIX domain socket

iongchun:
For long JSON/RPC response, current code has the following issues:

1. Multiple asynchronous requests may read response interleavingly, leads to incorrect JSON document;
2. If reading is faster than writing, read buffer might not be full but subsequent data will still "come" later.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/ipc/IpcService.java
+++ b/core/src/main/java/org/web3j/protocol/ipc/IpcService.java
@@ -4,6 +4,8 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 
+import java.util.concurrent.locks.ReentrantLock;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


fcorneli:
This issue can be easily demonstrated by reading "all" blocks in a loop.

fcorneli:
Problem can also be replicated when you have two observers:
```
web3.pendingTransactionObservable().subscribe(tx -> {
   ....
});
this.web3.blockObservable(true).subscribe(ethBlock -> {
   ....
});
```
On HttpService this works as expected.
On UnixIpcService, nothing gets through.

fcorneli:
I think the problem goes deeper than just locking. Looking at `HttpService` we see that each invocation has its own input/output stream, while `UnixIpcService` shared the input/output stream for different invocations. I think this simply cannot work this way.

4ntoine:
Any plans to accept this pull request or resolve it in different way?

conor10:
Apologies for the lack of input on this @iongchun - I presume this has been replaced by the updated PR? Please reopen if this is not the case.

iongchun:
No worry @conor10 
Yes this one is replaced by #494 sorry for not clearly stated

PR #264, opened 2017-12-01
Fix example. Check contract has been really deployed

earizon:
Since deploy(..).send() silently fails when running out of gas I guess next code is the proper one to deploy new contracts in a fail-fast (https://en.wikipedia.org/wiki/Fail-fast) way

### DIFF
--- a/docs/source/smart_contracts.rst
+++ b/docs/source/smart_contracts.rst
@@ -209,7 +209,10 @@ Construction and deployment of smart contracts happens with the *deploy* method:
            <web3j>, <credentials>, GAS_PRICE, GAS_LIMIT,
            [<initialValue>,]
            <param1>, ..., <paramN>).send();
-
+   if (contract.isValid()) { 
+     throw new Exception("contract deploy failed. Hint: try increasing gas limit"); 
+   }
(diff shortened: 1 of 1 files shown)


conor10:
Thanks for submitting @earizon. I'm in two minds about this change. My original intention was to leave it to the user to verify the deployment themselves. Which is why I didn't add this check to the contract deployment code.

docs/source/smart_contracts.rst
```diff
@@ -209,7 +209,10 @@ Construction and deployment of smart contracts happens with the *deploy* method:
            <web3j>, <credentials>, GAS_PRICE, GAS_LIMIT,
            [<initialValue>,]
            <param1>, ..., <paramN>).send();
-
+   if (contract.isValid()) { 
+     throw new Exception("contract deploy failed. Hint: try increasing gas limit");
```
iikirilov: I believe an extra space is required here

iikirilov:
@earizon Please could you follow through with small requested change so we can merge this. Thanks!

conor10:
Thanks for submitting however.

PR #267, opened 2017-12-04
Allow usage of adding directory instead of one file to truffle generator

harveymatt:
I'm currently adding CI to my solidity development and now have a process in place but would like to make the process a little simpler:

Currently the flow looks as follows:
1. Create .json files using truffle ( managed by npm )
2. Use gradle to pull in the Web3J dependencies
3. Execute defined gradle ExecJava tasks with TruffleJsonFunctionWrapperGenerator.java to generate the java contract files one by one
4. Using the artifactory plugin to upload the packaged jar.

Step 3 would be a little bit simpler if I could generate all the contracts for a given solidity source directory 

(I'm aware this could also be done via a bash script or a more complicated gradle task but thought here seemed reasonable)

Keen to get thoughts..

### DIFF
--- a/codegen/src/main/java/org/web3j/codegen/TruffleJsonFunctionWrapperGenerator.java
+++ b/codegen/src/main/java/org/web3j/codegen/TruffleJsonFunctionWrapperGenerator.java
@@ -2,12 +2,16 @@ package org.web3j.codegen;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


conor10:
It's certainly a valid idea. Would there be any issues running on directories such as [this](https://github.com/web3j/web3j/tree/master/codegen/src/test/resources/solidity/contracts), where files have a dependency on one another, or due to the nature of the Truffle metadata files, would this not actually be a concern?

eepstein:
This is trivial to do with a Bash script. 
I can add mine to the docs. 
My vote is: no need.

eepstein:
This is trivial to do with a Bash script. 
I can add mine to the docs. 
My vote is: no need.

conor10:
I'm tempted to agree with @eepstein here. I think it could be tricky managing dependencies on other smart contract metadata located in the same directory too, as mentioned above.

conor10:
Thanks for taking the time to submit @harveymatt.

PR #275, opened 2017-12-11
Include Log in EventResponse Object

jaspervdbijl:
SolidityFunctionWrapper.buildTypedResponse to include org.web3j.protocol.core.methods.response.Log in the generated EventResponse object


conor10:
Cool - please could you update the generated classes in the integration-tests module? You should just need to run the following [script](https://github.com/web3j/web3j/blob/master/codegen/src/test/resources/solidity/build.sh).

jaspervdbijl:
implemented as requested

conor10:
Thanks!

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -741,6 +767,13 @@ CodeBlock buildTypedResponse(String objectName,
         }
 
         CodeBlock.Builder builder = CodeBlock.builder();
+        if (observable) {
+            builder.addStatement("$L.log = log", objectName);
+        } else {
+            builder.addStatement(
+                    "$L.log = transactionReceipt.getLogs().get(valueList.indexOf(eventValues))",
```
danieldietrich: @jaspervdbijl I'm afraid, this will not work. In general the following holds:

* valueList.size() = k
* transactionReceipt.getLogs().size() = n

where k <= n.

Internally, the reason for this is because the hashes of Log topics are filtered by the encoded signature of the given contract event.

In my opinion, the only valid possibility to refer from events to logs is to transport the log reference in `org.web3j.abi.EventValues` during the filtering process in `org.web3j.tx.Contract.extractEventParameters(Event, Log)` resp. `staticExtractEventParameters(Event, Log)`.

However, this is not possible, because `EventValues` resides in the package `org.web3j.abi`, which is not aware of the package `org.web3j.protocol.core.methods.response` of the `Log` class. This means that we need to wrap `EventValues` in a class `EventValuesWithLog` (modulo name).

I would be happy to provide a PR that contains a fix.

/cc @conor10

danieldietrich: In order to fully understand it, an example might help:

Please consider two contracts:

```solidity
contract ContractA {
    event ThisDone(string msg);

    ContractB public contractB;

    function ContractA() {
        contractB = new ContractB();
    }
    
    function doThis() {
        contractB.doThat();
        ThisDone("this done");
    }
}

contract ContractB {
    event ThatDone(string msg);
    
    function doThat() {
        ThatDone("that done");
    }
}
```

On the call-site we do the following (informally):

```javascript
// truffle code
const ContractA = artifacts.require("ContractA");
...
await deployer.deploy(ContractA);
const contractA = await ContractA.deployed();
contractA.doThis({ from: senderAddress });
...
```

This leads to:

* transactionReceipt.getLogs().size() == 2
* valueList.size() == 1 // in both cases, filtering for ThisDone and ThatDone events
* the code `transactionReceipt.getLogs().get(valueList.indexOf(eventValues))` will always refer to the Log entry at index 0 (which is wrong)

jaspervdbijl:
Agreed,

https://github.com/web3j/web3j/pull/374

PR #280, opened 2017-12-13
fix for BlockObservable failing when node connection is lost

### LINKED ISSUE #279: new block Observable stops and does not recover when node connection fails
Creating a Observable via `web3j.blockObservable(true)` creates a 'ScheduledFuture' object to poll for filter changes. If an IOException is thrown due to the connection to the node going down, the exception is suppressed by the `ScheduledFuture`. Polling for future blocks also stops.

Expected behavior is either of:
1. Have a thread to check if an exception is thrown and propagate it accordingly. [This](https://www.cosmocode.de/en/blog/schoenborn/2009-12/17-uncaught-exceptions-in-scheduled-tasks?) resource provides a starting point to implement this.
2. Continue polling even if the connection to the node goes down and log exceptions rather than throwing them.

N.B. this behavior also occurs with all other observables that poll for filter changes

iikirilov:
fixes #279 
This changes allow for polling to continue even if node connection is lost.
What log message would be appropriate here? 
Should I just create a log object in the Filter class?

### DIFF
--- a/core/src/main/java/org/web3j/protocol/core/filters/Filter.java
+++ b/core/src/main/java/org/web3j/protocol/core/filters/Filter.java
@@ -75,13 +75,15 @@ public abstract class Filter<T> {
         EthLog ethLog = null;
         try {
             ethLog = web3j.ethGetFilterChanges(filterId).send();
+
+            if (ethLog.hasError()) {
+                throwException(ethFilter.getError());
+            }
+            process(ethLog.getLogs());
         } catch (IOException e) {
-            throwException(e);
+            System.out.println("LOG ME");
+            //throwException(e);
         }
-        if (ethLog.hasError()) {
-            throwException(ethFilter.getError());
-        }
-        process(ethLog.getLogs());
     }
 
     abstract EthFilter sendRequest() throws IOException;


conor10:
I've been having a play with this, the below approach should work using logs.

That being said, I'm still in two minds as to if this is the best approach. Downsides with the below implementation:

- If the initial getInitialFilterLogs() fails, an error is logged and the filter continues which isn't great
- Ultimately we're hiding an issue from the client from users, and putting in a workaround

If we were to propagate the exception back up to users via the Future.get approach you link to, the downsides are:

- Users will have to add the Observable.retry() method to observable requests (or we add them ourselves)
- It puts managing the re-connection choices in the hands over the users which might be a bad thing
- The Future.get() approach is a blocking task, that will mean we have to have enough threads to support the operation in the thread pool, which may not go down so well on Android - having a dedicated thread just for this

@jeremie-H you might be interested in this - we've got to the bottom of the issue, just thinking about the option for users

```java
package org.web3j.protocol.core.filters;

import java.io.IOException;
import java.math.BigInteger;
import java.util.Collections;
import java.util.List;
import java.util.Optional;
import java.util.concurrent.ScheduledExecutorService;
import java.util.concurrent.ScheduledFuture;
import java.util.concurrent.TimeUnit;

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

import org.web3j.protocol.Web3j;

import org.web3j.protocol.core.Request;
import org.web3j.protocol.core.Response;
import org.web3j.protocol.core.methods.response.EthFilter;
import org.web3j.protocol.core.methods.response.EthLog;
import org.web3j.protocol.core.methods.response.EthUninstallFilter;


/**
 * Class for creating managed filter requests with callbacks.
 */
public abstract class Filter<T> {

    private static final Logger log = LoggerFactory.getLogger(Filter.class);

    final Web3j web3j;
    final Callback<T> callback;

    private volatile BigInteger filterId;

    private ScheduledFuture<?> schedule;

    public Filter(Web3j web3j, Callback<T> callback) {
        this.web3j = web3j;
        this.callback = callback;
    }

    public void run(ScheduledExecutorService scheduledExecutorService, long blockTime) {
        try {
            EthFilter ethFilter = sendRequest();
            if (ethFilter.hasError()) {
                throwException(ethFilter.getError());
            }

            filterId = ethFilter.getFilterId();

            /*
            We want the filter to be resilient against client issues. On numerous occasions
            users have reported socket timeout exceptions when connected over HTTP to Geth and
            Parity clients. For examples, refer to
            https://github.com/web3j/web3j/issues/144 and
            https://github.com/ethereum/go-ethereum/issues/15243.

            Hence we consume errors and log them as errors, allowing our polling for changes to
            resume. The downside of this approach is that users will not be notified of
            downstream connection issues. But given the intermittent nature of the connection
            issues, this seems like a reasonable compromise.

            The alternative approach would be to have another thread that blocks waiting on
            schedule.get(), catching any Exceptions thrown, and passing them back up to the
            caller. However, the user would then be required to recreate subscriptions manually
            which isn't ideal given the aforementioned issues.
            */
            scheduledExecutorService.submit(this::getInitialFilterLogs);

            schedule = scheduledExecutorService.scheduleAtFixedRate(
                    () -> this.pollFilter(ethFilter),
                    0, blockTime, TimeUnit.MILLISECONDS);
            
        } catch (IOException e) {
            throwException(e);
        }
    }

    private void getInitialFilterLogs() {
        try {
            Optional<Request<?, EthLog>> maybeRequest = this.getFilterLogs(this.filterId);
            EthLog ethLog = null;
            if (maybeRequest.isPresent()) {
                ethLog = maybeRequest.get().send();
            } else {
                ethLog = new EthLog();
                ethLog.setResult(Collections.emptyList());
            }
            process(ethLog.getLogs());
        } catch (Throwable e) {
            // All exceptions must be caught, otherwise our job terminates without any notification
            log.error("Error sending request", e);
        }
    }

    private void pollFilter(EthFilter ethFilter) {
        try {
            EthLog ethLog = web3j.ethGetFilterChanges(filterId).send();

            if (ethLog.hasError()) {
                log.error("Invalid request: " + ethFilter.getError().getMessage());
            } else {
                process(ethLog.getLogs());
            }
        } catch (Throwable e) {
            // All exceptions must be caught, otherwise our job terminates without any notification
            log.error("Error sending request", e);
        }
    }

    abstract EthFilter sendRequest() throws IOException;

    abstract void process(List<EthLog.LogResult> logResults);

    public void cancel() {
        schedule.cancel(false);

        EthUninstallFilter ethUninstallFilter = null;
        try {
            ethUninstallFilter = web3j.ethUninstallFilter(filterId).send();
        } catch (IOException e) {
            throwException(e);
        }

        if (ethUninstallFilter.hasError()) {
            throwException(ethUninstallFilter.getError());
        }

        if (!ethUninstallFilter.isUninstalled()) {
            throwException(ethUninstallFilter.getError());
        }
    }

    /**
     * Retrieves historic filters for the filter with the given id.
     * Getting historic logs is not supported by all filters.
     * If not the method should return an empty EthLog object
     *
     * @param filterId Id of the filter for which the historic log should be retrieved
     * @return Historic logs, or an empty optional if the filter cannot retrieve historic logs
     */
    protected abstract Optional<Request<?, EthLog>> getFilterLogs(BigInteger filterId);

    void throwException(Response.Error error) {
        throw new FilterException("Invalid request: " + error.getMessage());
    }

    void throwException(Throwable cause) {
        throw new FilterException("Error sending request", cause);
    }
}

```

jeremie-H:
Hi @conor10 ! Thanks for this workaround :)
It will help for sure, but i'm still wondering what's the real problem. Because i lost connection to geth/parity. But on the same server, in localhost, it shouldn't. So I cant work on my java program because, plenty of blocks are skipped in errors due this timeout. (and the socket ipc dont work for big project (the unix socket seems to trim some messages).
I'm still searching for another solution (ethereumj but there are performance problems).
Thanks for your work ! :)

conor10:
It's a tricky one, as there's nothing fancy happening in web3j wrt to its HTTP connections. It's simply using OkHttp in a pretty vanilla way.

PR #284, opened 2017-12-19
Testrpc filter fix

NickFitton:
I found I was having issues with using filters with `testrpc/ganache-cli`, because it expects the first 9 filter id's to start with `0x0`, this pull hopes to fix this by preceding the filter id with a `0` when it is one of the first 9 filters.

### DIFF
--- a/utils/src/main/java/org/web3j/utils/Numeric.java
+++ b/utils/src/main/java/org/web3j/utils/Numeric.java
@@ -113,7 +113,7 @@ public final class Numeric {
         return toHexStringZeroPadded(value, size, true);
     }
     
-    public static String toHexStringWithPrefixSave(BigInteger value) {
+    public static String toHexStringWithPrefixSafe(BigInteger value) {
         String result = toHexStringNoPrefix(value);
         if (result.length() < 2) {
             result = Strings.zeros(1) + result;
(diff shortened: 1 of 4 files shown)


conor10:
Can you confirm that this works against Geth & Parity?

NickFitton:
Tested for ganache-cli, currently testing geth & parity, will let you know asap.

NickFitton:
I missed out some methods that use `filterId`. Amending this I will test on a Parity node.

NickFitton:
Can confirm the changes work with `geth` and `parity` nodes.

utils/src/main/java/org/web3j/utils/Numeric.java
```diff
@@ -112,6 +112,14 @@ public static String toHexStringNoPrefix(byte[] input) {
     public static String toHexStringWithPrefixZeroPadded(BigInteger value, int size) {
         return toHexStringZeroPadded(value, size, true);
     }
+    
+    public static String toHexStringWithPrefixSafe(BigInteger value) {
```
conor10: Please can you add a few test cases for this specific method? Appreciate that the RequestTest provides coverage, but I'd like something explicit.

NickFitton: Yeah sure, I'll write it out by the end of the day

NickFitton:
I ran tests with filters on local parity and geth chains and on external chains, both worked fine.

nf-PostQuantum:
Tests added for the leading `0`'s

conor10:
Thanks!

Issue #281, opened 2017-12-15
Got Error:trouble processing "java/lang/SafeVarargs.class": under Android Studio after upgrade web3j lib from 2.2.1 to 3.1.1


vikulin:
Error:trouble processing "java/lang/SafeVarargs.class":
Error:Ill-advised or mistaken usage of a core class (java.* or javax.*)
Error:when not building a core library.
Error:This is often due to inadvertently including a core library file
Error:in your application's project, when using an IDE (such as
Error:Eclipse). If you are sure you're not intentionally defining a
Error:core class, then this is the most likely explanation of what's
Error:going on.
Error:However, you might actually be trying to define a class in a core
Error:namespace, the source of which you may have taken, for example,
Error:from a non-Android virtual machine project. This will most
Error:assuredly not work. At a minimum, it jeopardizes the
Error:compatibility of your app with future versions of the platform.
Error:It is also often of questionable legality.
Error:If you really intend to build a core library -- which is only
Error:appropriate as part of creating a full virtual machine
Error:distribution, as opposed to compiling an application -- then use
Error:the "--core-library" option to suppress this error message.
Error:If you go ahead and use "--core-library" but are in fact
Error:building an application, then be forewarned that your application
Error:will still fail to build or run, at some point. Please be
Error:prepared for angry customers who find, for example, that your
Error:application ceases to function once they upgrade their operating
Error:system. You will be to blame for this problem.
Error:If you are legitimately using some code that happens to be in a
Error:core package, then the easiest safe alternative you have is to
Error:repackage that code. That is, move the classes in question into
Error:your own package namespace. This means that they will never be in
Error:conflict with core system classes. JarJar is a tool that may help
Error:you in this endeavor. If you find that you cannot do this, then
Error:that is an indication that the path you are on will ultimately
Error:lead to pain, suffering, grief, and lamentation.
Error:1 error; aborting
Error:Execution failed for task ':app:transformClassesWithDexForRelease'.
> com.android.build.api.transform.TransformException: com.android.ide.common.process.ProcessException: java.util.concurrent.ExecutionException: com.android.ide.common.process.ProcessException: Return code 1 for dex process

maratsubkhankulov:
Also experiencing this issue, worked around it by downgrading to 3.0.1-android

conor10:
What dependencies does the project use?

@joanesespanol - do you have any thoughts?

joanesespanol:
Hey guys,

[...]

PR #287, opened 2017-12-20
made sure to not include the cloned SafeVarargs class in the final build

joanesespanol:
@conor10 This should address the issue reported here https://github.com/web3j/web3j/issues/281.


conor10:
@vikulin @maratsubkhankulov would be great to get your feedback as per @joanesespanol's comment.

PR #290, opened 2017-12-26
Added annotation for address validation.

ganchix:
Very useful for validating inputs, you only have to annotate the input, for example:
```
@Data
@NoArgsConstructor
@AllArgsConstructor
@Builder
public class TransferEvent {

    /**
     * From address
     */
    @NotNull
    @Address
    String address;

    /**
     * To address
     */
    @NotNull
    @Address
    String addressTarget;

    /**
     * Value in ether
     */
    @NotNull
    BigDecimal value;
}
```


iikirilov:
Thank you for taking the time to submit this PR. It has been decided that we will stick to the philosophy that "less is more" and I am against adding extra dependencies for just one annotation.

Issue #289, opened 2017-12-21
any reason not to make Contract's gasPrice field settable?


eepstein:
I'm sometimes running into errors in a long-running service due to the increase in gasPrice on the Ethereum network.  

I've got a way to dynamically load current gas prices into the service, the problem is passing those along to the contract for use.  Currently we use a singleton-bean for the contract and just create it during app wire-up.  But that means we're stuck with the initial gasPrice.

I see two options: change from a singleton pattern to instead creating a new instance of the contract whenever we invoke it - or at least as often as we notice and issue with gasPrice or... more simply, add a setter method to the parent org.web3j.tx.Contract class so that we can dynamically update the gasPrice.

Any reason not to do the latter?

eepstein:
Doing a bit more digging - the existing "getGasPrice()" method is declared on `Contract`'s parent class - `ManagedTransaction` and is implemented as a synchronous send to `ethGasPrice()` making in wholly disconnected from `Contract`'s `gasPrice` field.  

This might be something that happened through accretion and never got revisited ?

Anyway, it looks like that `getGasPrice()` method is only used in one place in the code: the `Transfer` class.  But since the method is public it could be people are using it elsewhere.

So... probably the best way to align things is to give the `gasPrice` field on `Contract` a new name.  This is the route I suggest since a) the field is declared `protected` and until recently the generated classes were all final and offered no way to set it outside the constructor.  So, it's more likely that no one's code references that field and hence a name change will likely cause little if any disruption.

Anyway, the point is: these things aren't aligned and they should be.  If we have a gasPrice field, let's make it bean-like.  And if we need a dynamic method to get the value, let's use a different name. OR just don't add that method since it's a one-liner anyway just have `Transaction` (and anywhere else) call: 

    web3j.ethGasPrice().send().getGasPrice();

That doesn't seem like so much code that it needs a convenience method - especially one that's called in only one place in the framework.

What's the best way to fix this?

Thoughts?

eepstein:
How about this:

1. declare the `getGasPrice()` method on `ManagedTransaction` abstract;
1. push down the current concrete implementation to the `Transfer` class;
1. implement this method on `Contract` to just return the `gasPrice` field.

If you're ok with this I'll do a pull request.

NOTE: this is a potentially breaking change for any users who rely on the current implementation on `ManagedTransaction`, as if they're using it from a Contract they'll now get the non-dynamic value instead.

[...]

PR #292, opened 2017-12-28
Allow gasPrice to be set post construction

eepstein:
gasPrice is a dynamic value - and can be retrieved from the eth gasstation API - contract instances may be long-lived, a setter method allows a contract's gasPrice to be adjusted for long-lived contract instances.

per #289


conor10:
Makes sense - thanks.

PR #293, opened 2017-12-29
Mnemonics/BIP39 support

### LINKED ISSUE #188: Mnemonics/BIP39 support
Hi,
  Do you plan, or maybe there is a way (which I did not find) to export the stored private keys as a BIP-39 compliant set of mnemonic words? Thanks a lot!

Marcin

alimate:
`MnemonicUtils` class contains two methods:
 - `generateMnemonic` which generates a random mnemonic given a 128-256 bit of initial entropy. 
 - `generateSeed` which generates a BIP-39 seed given a mnemonic and a passphrase.

Both methods are totally compatible with test vectors provided by the [BIP-39 reference implementation.](https://github.com/trezor/python-mnemonic/blob/master/vectors.json)

In order to create a wallet supporting BIP-39 mnemonics, we've added a new method in `WalletUtils` named `generateBip39Wallet` which basically generates a wallet with a random mnemonic. The generated wallet is a [*Type 1 deterministic wallet*](https://en.bitcoin.it/wiki/Deterministic_wallet) with:
```
SHA256(seed(mnemonic, walletPass))
```
As the private key.

Resolves #188


crypto/src/main/java/org/web3j/crypto/WalletUtils.java
```diff
@@ -134,4 +168,9 @@ public static boolean isValidAddress(String input) {
 
         return cleanInput.length() == ADDRESS_LENGTH_IN_HEX;
     }
+
+    private static byte[] sha256(byte[] input) throws NoSuchAlgorithmException {
```
conor10: This should be using the version defined in `MnemonicUtils`

crypto/src/main/java/org/web3j/crypto/WalletUtils.java
```diff
@@ -24,6 +26,7 @@
 public class WalletUtils {
 
     private static final ObjectMapper objectMapper = new ObjectMapper();
+    private static final SecureRandom secureRandom = new SecureRandom();
```
conor10: Please get the SecureRandom instance from here https://github.com/web3j/web3j/blob/master/crypto/src/main/java/org/web3j/crypto/SecureRandomUtils.java

crypto/src/test/java/org/web3j/crypto/WalletUtilsTest.java
```diff
@@ -154,4 +166,12 @@ public void testIsValidAddress() {
         assertFalse(isValidAddress(SampleKeys.ADDRESS + 'a'));
         assertFalse(isValidAddress(SampleKeys.ADDRESS.substring(1)));
     }
+
+    private byte[] sha256(byte[] input) {
```
conor10: How come this is defined again here?

crypto/src/main/java/org/web3j/crypto/MnemonicUtils.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * To create a binary seed from the mnemonic, we use the PBKDF2 function with a
+     * mnemonic sentence (in UTF-8 NFKD) used as the password and the string "mnemonic"
+     * + passphrase (again in UTF-8 NFKD) used as the salt. The iteration count is set
+     * to 2048 and HMAC-SHA512 is used as the pseudo-random function. The length of the
+     * derived key is 512 bits (= 64 bytes).
+     *
+     * @param mnemonic The input mnemonic
```
conor10: I think it's worth expanding slightly on the mnemonic format in the parameter description.

crypto/src/test/java/org/web3j/crypto/WalletUtilsTest.java
alimate: Moved to `Hash` utility class in `utils` module

crypto/src/main/java/org/web3j/crypto/WalletUtils.java
alimate: Moved to Hash utility class in utils module

crypto/src/main/java/org/web3j/crypto/WalletUtils.java
alimate: Fixed.

crypto/src/main/java/org/web3j/crypto/MnemonicUtils.java
alimate: Done!

crypto/src/main/java/org/web3j/crypto/Bip39Wallet.java
```diff
@@ -0,0 +1,40 @@
+package org.web3j.crypto;
+
+/**
+ * Data class encapsulating a BIP-39 compatible Ethereum wallet.
+ *
+ * @author Ali Dehghani
```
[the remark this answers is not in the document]
alimate: Removed!

crypto/src/test/java/org/web3j/crypto/WalletUtilsTest.java
conor10: Looks like this code hasn't been removed yet - please remove, and I'll merge.

alimate:
@conor10 All requested changes have been applied 😊

conor10:
Great work 💯

vberezkin:
Is it possible to add BIP32 derivation path to make generated address be compatible with other wallet apps ?

chenhao597:
How I can load Bip39 wallet by mnemonic without password?
`public static Credentials loadBip39Credentials(String password, String mnemonic)` need password,but if I just have mnemonic,how I can do it???

PR #294, opened 2018-01-01
Closed FileInputStream which is not closed after reading from file.

kamaci:
FileInputStream is closed.


conor10:
Good spot - but you need to remove the `close()`, then I'll merge.

kamaci:
@conor10 could you check it again?

conor10:
Thanks!

PR #301, opened 2018-01-05
make Contract's gasPrice a bean property

### LINKED ISSUE #289: any reason not to make Contract's gasPrice field settable?

eepstein:
getGasPrice renamed to ManagedTransaction.requestCurrentGasPrice 

Contract now has regular bean property accessor and mutator methods getGasPrice() and setGasPrice()

Fixes #289


eepstein:
Seems the codecov ding, above, was from a different commit ?

eepstein:
@conor10 any love ?

conor10:
Looking good 👍

Issue #269, opened 2017-12-05
Error in documentation page about Transactions


javahippie:
In the documentation page about transactions, there seems to be an issue with the part ["Transacting with a smart contract"](https://docs.web3j.io/transactions.html#transacting-with-a-smart-contract). It shows a call of the function Transaction#createEthCallTransaction only with the contract address and data like this:

    Transaction.createEthCallTransaction(contractAddress, encodedFunction)

The [method itself](https://github.com/web3j/web3j/blob/958ad9819e29a08cf0ef2b7263b64618df740dc2/core/src/main/java/org/web3j/protocol/core/methods/request/Transaction.java) on the other hand also requires the `from` address to be passed into:

    public static Transaction createEthCallTransaction(String from, String to, String data) {

        return new Transaction(from, null, null, null, to, null, data);
    }

conor10:
Are you able to provide a PR with this update? Makes my life easier :)

javahippie:
No problem, I hope it is okay this way?

PR #307, opened 2018-01-10
Added 'from' property to the documentation about 'Querying the state …

javahippie:
…of a smart contract'

Referencing issue #269 

I enclosed the <from> parameter in brackets, as it seemed to be the existing style for parameters in the code examples which were not introduced before.


javahippie:
Maybe the 'docs' folder should be exluded from the code coverage analysis

conor10:
Thanks!

PR #314, opened 2018-01-12
Fixed #257 Add warning to Solidity when non-constant function returns…

### LINKED ISSUE #257: Add warning to Solidity when non-constant function returns a value
Smart contract methods should either be constant (i.e. return a value) or transactions which mutate a value internally (never returning a value).

The SolidityFunctionWrapper looks for the `constant` modifier on function definitions to ascertain if they are [view functions](http://solidity.readthedocs.io/en/develop/contracts.html#view-functions) or not. If the `constant` modifier is present, the smart contract wrapper will return a value.

If it isn't, then it should return a transaction receipt, as it should be invoked as part of a transaction.

The problem is that not all smart contract developers include these modifiers. This is a point of confusion, especially when they are getting started with Ethereum.

SolidityFunctionWrapper should be updated to provide a warning in the following scenario:

if abiDefinition.functionType != constant && abiDefinition.returnValues != empty
log warning "function definition <function> returns a value but is not defined as a constant function. Please ensure it contains the constant modifier if you want to read the return value".

moureauf:
Added a warning when a function is not constant but returns a value.

### DIFF
--- a/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
+++ b/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
@@ -21,6 +21,8 @@ import com.squareup.javapoet.ParameterizedTypeName;
 import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeSpec;
 import com.squareup.javapoet.TypeVariableName;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import rx.functions.Func1;
 
 import org.web3j.abi.EventEncoder;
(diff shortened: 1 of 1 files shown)


conor10:
The issue with this approach is that the SolidityFunctionWrapperGenerator operates on the console via System.out.println. So this message would never appear to users running `web3j solidity generate` via the command line.

So there are a couple of options here:

1. Migrate to using an appropriate logger configuration that looks like just like a command line output, but is in fact a logger (you'll need to remove the slf4j-nop dependency), and update SolidityFunctionWrapperGenerator
2. Return a list of warning messages from the `new SolidityFunctionWrapper(useJavaNativeTypes).generateJavaFiles(...)` call in SolidityFunctionWrapperGenerator, then print them to the console

I'll let you decide on which you believe to be the most appropriate

PR #326, opened 2018-01-20
changed the ordering of nonIndexedParameter and indexParameters to ensure proper sha3 hash can be created

justinchan1991:


abi/src/main/java/org/web3j/abi/EventEncoder.java
```diff
@@ -20,21 +20,21 @@
     private EventEncoder() { }
 
     public static String encode(Event function) {
-        List<TypeReference<Type>> indexedParameters = function.getIndexedParameters();
         List<TypeReference<Type>> nonIndexedParameters = function.getNonIndexedParameters();
+        List<TypeReference<Type>> indexedParameters = function.getIndexedParameters();
```
conor10: Why is this change required?

abi/src/main/java/org/web3j/abi/EventEncoder.java
```diff
@@ -20,21 +20,21 @@
...
 
         String methodSignature = buildMethodSignature(function.getName(),
-                indexedParameters, nonIndexedParameters);
+                nonIndexedParameters, indexedParameters);
 
         return buildEventSignature(methodSignature);
     }
 
     static <T extends Type> String buildMethodSignature(
```
conor10: A unit test should accompany any changes.

abi/src/main/java/org/web3j/abi/EventEncoder.java
justinchan1991: It's just to make the re-ordering look consistent throughout the EventEncoder class. Since I changed from placing indexParameters then unindexedParameter in the following code, 

`-        List<TypeReference<T>> parameters = new ArrayList<>(indexParameters);
 -        parameters.addAll(nonIndexedParameters);
 +        List<TypeReference<T>> parameters = new ArrayList<>(nonIndexedParameters);
 +        parameters.addAll(indexParameters);`

I thought it would be good to also change rest of the class to follow the same suit (nonindexedParameters then indexedParameters)

abi/src/main/java/org/web3j/abi/EventEncoder.java
justinchan1991: will do this asap

iikirilov:
@justinchan1991 please fix the styling

justinchan1991:
added a commit https://github.com/web3j/web3j/pull/326/commits/0ad8df9e7595ce7be864872eef1ec1ed6fa32656

PR #330, opened 2018-01-23
build: quash warning in core/build.gradle

benjumanji:
Remove usage of deprecated method.


benjumanji:
That code coverage failure seems a bit off.

conor10:
Thanks.

Issue #328, opened 2018-01-22
Support for async transaction signing.


benjumanji:
Hi would you be open to a pull request that allowed keypair to complete a signature with a callback?

Use case: Using hashicorp vault's transit backend to do signing.

Obviously I can write this code myself, but if I did it might be nice to have in  web3j directly.

conor10:
Hey Ben, absolutely.

benjumanji:
Thanks! Expect a pr in the next week or so.

PR #331, opened 2018-01-23
[WIP] Add new Async sign method to TransactionEncoder

benjumanji:
Add async methods to allow for out of process signing to use efficient
non-blocking constructions.

We use a double-barrelled continuation passing style to allow for both
sync and async code to share the majority of code with minimal overhead
whhile still leaving the door open to the user overriding the
async methods with their framework of choice (Futures etc).

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/ECKeyPair.java
+++ b/crypto/src/main/java/org/web3j/crypto/ECKeyPair.java
@@ -3,6 +3,7 @@ package org.web3j.crypto;
 import java.math.BigInteger;
 import java.security.KeyPair;
 import java.util.Arrays;
+import java.util.function.Consumer;
 
 import org.bouncycastle.crypto.digests.SHA256Digest;
 import org.bouncycastle.crypto.params.ECPrivateKeyParameters;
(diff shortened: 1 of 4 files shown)


benjumanji:
First pass at #328

benjumanji:
Things to like:
1. close to zero cost for the existing methods
2. execution agnostic for people extending
Things to dislike:
1. If you switch out the eckeypair at the bottom of the stack for one that really does it's work asynchronously it becomes profoundly unsafe to call the other methods above it synchronously. I'd argue that if you are overriding the signing to do it out of process you _should_ know this. We could consider adding an assertion to the `Sync.call` method that at least _one_ of the continuations has been called the Async method returns which should always be true when the callbacks were executed directly on the same thread, and not true when execution was migrated to a different threads in the case of a genuine async boundary.

conor10:
Hi @benjumanji, I got the impression that this change would be fairly light touch in terms of providing a callback. In its current form, it's a bit too intrusive - I'm not so keen on embedding these async methods within the class. It's quite an opinionated way of doing things, which differs somewhat from the style of web3j overall.

I appreciate that you've taken the time to do this, and there's clearly nothing wrong with this approach per say, but I'd rather not complicate the library with this. Additionally it would be another piece of code to maintain that differs from the Android branch.

benjumanji:
@conor10 That's totally fine! I'm not sure it can be made _less_ intrusive. I mean, I can move the two consumers into a single callback, but that doesn't address your concern. If I don't have callbacks through the complete stack, I can't do too much. I have to have this, whether it gets into upstream or not, so the work isn't wasted.

I can maintain this patch. I'll close this if you confirm that you can't see a way to make this work in a way that would make you happy.

benjumanji:
Also I'd be happy to make this code android compatible, if that would change your mind.

benjumanji:
Ok I just opened another pull that is significantly lighter over at #334, it still adds methods, but is half the size, doesn't aim for such a wacky method of performing async calculations. Hopefully it feels a bit more native? No pressure, this took me 10 minutes to whip up.

PR #334, opened 2018-01-25
[WIP] Add Async Transaction signing

benjumanji:
Adapt TransactionEncoder and friends to allow for non-blocking
constructions when performing out of band transaction signing.

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/Async.java
+++ b/crypto/src/main/java/org/web3j/crypto/Async.java
@@ -1,0 +1,16 @@
+package org.web3j.crypto;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+
+public class Async {
+
+    public static <R> R sync(CompletableFuture<R> future) {
+        try {
+            return future.get();
+        } catch (InterruptedException | ExecutionException e) { 
+            throw new RuntimeException(e);
+        }
+    }
+}
+
--- a/crypto/src/main/java/org/web3j/crypto/ECKeyPair.java
+++ b/crypto/src/main/java/org/web3j/crypto/ECKeyPair.java
@@ -3,6 +3,8 @@ package org.web3j.crypto;
 import java.math.BigInteger;
 import java.security.KeyPair;
 import java.util.Arrays;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 
 import org.bouncycastle.crypto.digests.SHA256Digest;
 import org.bouncycastle.crypto.params.ECPrivateKeyParameters;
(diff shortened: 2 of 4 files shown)


benjumanji:
Attempt number 2 at solving #328.

crypto/src/main/java/org/web3j/crypto/Async.java
```diff
@@ -0,0 +1,16 @@
+package org.web3j.crypto;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+
+public class Async {
```
conor10: I'd rather not have a sync helper method in all of this. My preference is to have all of the core code paths taking place in a synchronous manner by default, without additional futures in the mix.

Besides, there is already an Async helper class in the project - see https://github.com/web3j/web3j/blob/master/core/src/main/java/org/web3j/utils/Async.java. Appreciate it doesn't cater for this specific functionality, but these utils should only be defined in one place.

crypto/src/main/java/org/web3j/crypto/ECKeyPair.java
```diff
@@ -39,13 +41,18 @@ public BigInteger getPublicKey() {
      * @return  An {@link ECDSASignature} of the hash
      */
     public ECDSASignature sign(byte[] transactionHash) {
+        return Async.sync(signAsync(transactionHash));
+    }
+
+    public CompletableFuture<ECDSASignature> signAsync(byte[] transactionHash) {
```
conor10: To my point above, can't we leave the default sync as is, then just wrap the async versions in a completable future. I'm not a fan of doing this all async first, as it results in lots of futures embedded in the methods which is unnecessary IMHO.

crypto/src/main/java/org/web3j/crypto/Async.java
benjumanji: On the existing class, I did notice it, but I realised it would add an extra dependency to `crypto` and I wasn't sure if it was desired or not. I can certainly change that.

crypto/src/main/java/org/web3j/crypto/ECKeyPair.java
benjumanji: That's easy enough to do (I will do that) in this case at the bottom of the stack. Where it gets more interesting is in `Sign.java`. You have a bunch of computation depending on a result that _could_ be delivered async. None of the functionality in that class is packaged up in such a way that someone can reuse it if they were to want async execution.

I just want to check something: you don't like this for reasons other than perf right? Because perf wise I am sure this introduces negligible overhead. It's just going to be a couple of CAS ops / extra objects. I'd be happy to provide jmh benches.

Alternative way forward: I do the wrapping as suggested (sync code stays the same, extra method, wraps sync call in completed future), _but_ inside of `Sign` I make a couple of protected methods that make it easy for someone extending to repackage that code to use an inner async signature call.

Thanks for sticking with me!

conor10:
Hi @benjumanji, looking at this again. Fundamentally I'd like the existing transaction signing code code to be sync first, then if you want certain parts to be async, I'm certainly happy for refactorings to support this. I'm not keen on existing users of the code having a move to async behind the scenes.

PR #338, opened 2018-01-25
Add TransactionReceipt.toString() and Log.toString()

oscarguindzberg:


core/src/main/java/org/web3j/protocol/core/methods/response/Log.java
```diff
@@ -207,4 +207,20 @@ public int hashCode() {
         result = 31 * result + (getTopics() != null ? getTopics().hashCode() : 0);
         return result;
     }
+
+    @Override
+    public String toString() {
+        return "Log{" +
+                "removed=" + removed +
```
conor10: The `+` need to be at the start of the line to keep Checkstyle happy.
Will merge once updated.

oscarguindzberg:
@conor10 done.

conor10:
Thanks!

PR #346, opened 2018-01-31
Contract wrapper generator should convert array items to native types

dukei:
This pull request fixes the following bug:

Suppose contract function returns `bytes32[]`. The generator generates method returning `List<byte[]>` (for `--javaTypes` flag). But in fact the method returns `List<Bytes32>` - only `Array` gets converted.  So web3j codegen does not convert items in arrays to native types.

Suggested fix: generated functions perform deep conversion for arrays.

**Warning:** for converting arrays the generator creates method `convertToNative` in each contract class. This method currently uses Java8 streams. Consider rewriting this function (and moving it to base `Contract` class) to support Android.

I did not move it to `Contract` base class myself because in this case I will not be able to load web3j from maven until you release the version with this pull request merged. Now only code generator is affected.

All tests passed.


conor10:
Thanks for spotting this and submitting a fix.

There is one issue with generating the streams - all smart contract wrapper code that is generated should be Java 1.6 compatible, otherwise Android users would have to use a different version of the command line tools which is not appropriate.

Is it possible to submit the change against the contract class and manually insert into the smart contract wrappers you're working with?

dukei:
Sorry, I didn't quite catch what you mean. But if you are talking about wrapper java 1.6 compatibility it can be easily achieved by moving `convertToNative` method from wrapper into the `Contract` base class and removing `buildListConverter` method (and the call to it) from  `SolidityFunctionWrapper`.

If you need web3j source code to be compatible with java 1.6 too, you can rewrite `convertToNative` into the following:

```Java
protected static <OriginalType, NewType> List<NewType> convertToNative(List<OriginalType> arr) {
    List<NewType> out = new ArrayList<NewType>();
    for(Iterator<OriginalType> it = arr.iterator(); it.hasNext(); ){
        out.add(it.next());
    }
    return out;
}
```

Will it solve the problem? If yes I can update the pull request with these changes.

iikirilov:
Please go ahead and update this PR as you have suggested so it can be merged.

dukei:
I have made the changes to support Java 6. I am a bit confused why coverage tests fail this time. I have mostly removed some code since the last commit.

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -9,6 +9,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
+import java.util.stream.Collectors;
```
conor10: This still needs removing.

core/src/main/java/org/web3j/tx/Contract.java
```diff
@@ -430,4 +431,13 @@ public final String getDeployedAddress(String networkId) {
         return addr == null ? getStaticDeployedAddress(networkId) : addr;
     }
 
+    @SuppressWarnings("unchecked")
+    protected static <OriginalTypeT extends Type, NewTypeT>
```
conor10: Could we just use S & T here? I'm not so keen on lengthy parameterised type names.

dukei:
@conor10 I made the changes you requested

conor10:
Thanks!

PR #363, opened 2018-02-14
implementation of ENS Reverse name resolution

yuriymyronovych:
Added support for ENS Reverse name resolution according to https://docs.ens.domains/en/latest/userguide.html#reverse-name-resolution. 
Tested and working.


iikirilov:
Thank you for taking the time to submit this PR! :100:

PR #367, opened 2018-02-16
Improve quick start documentation. remove confusing InfuraHttpClient …

alexprut:
Remove unnecessary complexity for the quick introduction of the usage.
Alternatively if the usage of the class `InfuraHttpClient` is required with the service `Infura` then explicitly specify to the end used to insert in it's gradle (or whatever) dependency manager also the dependency: `compile 'org.web3j:infura:3.2.0'`


iikirilov:
Thank you for submitting this PR! :100:

Issue #364, opened 2018-02-15
Contribution vs platform dependent names


danieldietrich:
Currently there seem to be same file names with different capitalization of letters(see image below). That may be a problem for windows users, they are not able to locally clone the repository.

<img width="983" alt="screen shot 2018-02-15 at 10 40 35" src="https://user-images.githubusercontent.com/743833/36250073-06dac9da-123d-11e8-8daf-d23dd13b1e6d.png">

Interestingly, I also get an error on my Mac when I download a fresh copy of the web3j repo.

When I do:

```bash
$ git clone https://github.com/web3j/web3j.git
```

I get the following:

```bash
$ git status
On branch master
Your branch is ahead of 'origin/master' by 390 commits.
  (use "git push" to publish your local commits)
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   codegen/src/test/resources/solidity/shipit/build/ShipIt.bin

no changes added to commit (use "git add" and/or "git commit -a")
```

Does anyone else experience these problems?

saulbein:
Experiencing the same issue, windows can't handle names with different capitalizations, and git is bugging out leaving me unable to pull because of shadow changes to the ShipIt.bin file. When reverting it changes to another diff, thus making it impossible for me to not have working tree changes.

danieldietrich:
Even more interesting - git states that one file was modified but it isn't present in the directory of a clean master copy:

```bash
#
# The revision
#
$ git show
commit fdf83488a4b5bdde5968dabd2ce0930afee17619 (HEAD -> master, upstream/master, origin/master, origin/HEAD)
Merge: 231b15ae fa97c996
Author: Conor Svensson <conor10@gmail.com>
Date:   Fri Feb 9 09:31:39 2018 +0000

    Merge pull request #338 from oscarguindzberg/toString
    
    Add TransactionReceipt.toString() and Log.toString()

#
# The status directly after a clean clone
#
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   ShipIt.bin

no changes added to commit (use "git add" and/or "git commit -a")

#
# The content of the ominous directory
#
$ pwd
/Users/daniel/git/danieldietrich/web3j/codegen/src/test/resources/solidity/shipit/build

[...]

PR #375, opened 2018-02-22
Fixes the Log addition to contract events

### LINKED ISSUE #270: Relate events to event sender
I have several instances of the same contract living on the blockchain. I want to pick up events by the event method provided by the wrapper classes. Those methods deliver event objects which have no relation to the underlying Log instance and thereby don't allow me to determine which instance sent the event.

Log contains the needed information (address) but not the event.

```
public class Log {
    private boolean removed;
    private String logIndex;
    private String transactionIndex;
    private String transactionHash;
    private String blockHash;
    private String blockNumber;
    private String address;
    private String data;
    private String type;
    private List<String> topics;
...
}
```

And since the event does not include the log index I cannot safely navigate from the event to the log instance.

Please provide the requested information or provide a way to derive the event directly from the log instance.

danieldietrich:
Fixes #270 

See PR #275, especially [this review comment](https://github.com/web3j/web3j/pull/275#pullrequestreview-98460589).

_Note: The issue #364 hinders me to squash the commits._


iikirilov:
Thanks!

PR #378, opened 2018-02-22
fix #368

### LINKED ISSUE #368: Add support for generating functions with structs as parameters.
Generate inner classes from struct parameters in the contract wrapper.

https://github.com/web3j/web3j/blob/master/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java

```
"inputs":[  
     {  
        "components":[ {  
              "name":"var1",
              "type":"string"
           },
           {  
              "name":"var2",
              "type":"uint32"
           }, 
           {...},
           ... ],
        "name":"_myStruct",
        "type":"tuple"
     }, 
     {...}, 
     ...
  ]
```

iikirilov:
A `struct` used in param/return will be encoded with `type: tuple` and will contain a `components: [...]` element.

Things to consider:
1. Struct used in two functions as param/return with different variable name
2. Different Structs used in different functions as param/return with the same variable name

Regards issue #368


iikirilov:
This PR is no longer relevant - please see #775

PR #387, opened 2018-02-27
Expose additional behaviors from auto-generated contract classes for use with ethGetLogs

xaviarias:
Adds an event type definition as a public constant  with a suffix `_EVENT` for each declared contract event. 

To provide more fine-grained control over event filters, adds a method  that accepts an `EthFilter` filter in addition to the default event observable.


conor10:
Thanks @xaviarias.

Would you be able to update the Smart Contract wrappers in https://github.com/web3j/web3j/tree/master/codegen/src/test/resources/solidity too?

crypto/src/main/java/org/web3j/crypto/Sign.java
```diff
@@ -108,7 +108,7 @@ private static BigInteger recoverFromSignature(int recId, ECDSASignature sig, by
         BigInteger i = BigInteger.valueOf((long) recId / 2);
         BigInteger x = sig.r.add(i.multiply(n));
         //   1.2. Convert the integer x to an octet string X of length mlen using the conversion
-        //        routine specified in Section 2.3.7, where mlen = ⌈(log2 p)/8⌉ or mlen = ⌈m/8⌉.
```
xaviarias: I removed the `⌈` character because was giving me compilation errors on OSX 10.13.3 with Oracle JDK 1.8.0_162, and I end up committing this by mistake.

`:rlp:javadoc
/Users/xavi/Development/Projects/web3j/crypto/src/main/java/org/web3j/crypto/Sign.java:111: error: unmappable character for encoding ASCII
        //        routine specified in Section 2.3.7, where mlen = ???(log2 p)/8??? or mlen = ???m/8???.`

A workaround would be to add this to the  global `build.gradle`:
```
javadoc {
    options.encoding = 'UTF-8'
}
```

I'll push an additional commit fixing this.

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -715,22 +731,44 @@ MethodSpec buildEventObservableFunction(
                         .build())
                 .build();
 
+        observableMethodBuilder
+                .addStatement("return web3j.ethLogObservable(filter).map($L)", converter);
```
conor10: The generated code needs to be Java 1.6 compatible - the reason being that smart contract wrappers for Java and Android are generated using exactly the same process.

xaviarias: I don't see where the generated code is not Java 6 compatible 😕 

The generated code for this specific line is perfectly fine:
```
return web3j.ethLogObservable(filter).map(new Func1<Log, TransferEventResponse>() {
   ...
});
```
Am I missing something?

xaviarias: There's however an incompatibility in the `_addresses` field initialiser:
```
static {
    _addresses = new HashMap<>();
    _addresses.put("4", "0xaea9d31a4aeda9e510f7d85559261c16ea0b6b8b");
    _addresses.put("4447", "0x254dffcd3277c0b1660f6d42efbb754edababc2b");
}
```
This should be `_addresses = new HashMap<String, String>();`

conor10: Apologies my bad!

conor10:
👍

PR #388, opened 2018-02-28
query for historical contract state

fcorneli:
One of the interesting features of blockchain technology is that it preserves historical state.
Hence this patch makes it possible to query for such historical state on contracts.


conor10:
Thanks!

PR #395, opened 2018-03-01
Prepare web3j-android v3.3.0

iikirilov:


codegen/src/main/java/org/web3j/codegen/TruffleJsonFunctionWrapperGenerator.java
```diff
@@ -123,9 +123,10 @@ private void generate() throws IOException, ClassNotFoundException {
             System.out.printf("Generating " + basePackageName + "." + className + " ... ");
             Map<String, String> addresses;
             if (c.networks != null && !c.networks.isEmpty()) {
-                addresses = c.networks.entrySet().stream().collect(Collectors.toMap(
-                        Map.Entry::getKey, e -> e.getValue().getAddress()
-                ));
+                addresses = c.networks.entrySet().stream()
```
conor10: This should cause the build to fail as its not 1.6 compatible.

core/src/main/java/org/web3j/utils/Files.java
```diff
@@ -13,9 +13,9 @@ private Files() { }
 
     public static byte[] readBytes(File file) throws IOException {
         byte[] bytes = new byte[(int) file.length()];
-        FileInputStream fileInputStream = new FileInputStream(file);
-        fileInputStream.read(bytes);
-        fileInputStream.close();
+        try (FileInputStream fileInputStream = new FileInputStream(file)) {
```
conor10: Again this is Java 1.7 functionality so should be removed.

PR #396, opened 2018-03-01
fix incompatible code

iikirilov:


crypto/src/main/java/org/web3j/crypto/MnemonicUtils.java
```diff
@@ -163,7 +164,12 @@ private static byte calculateChecksum(byte[] initialEntropy) {
         URL url = Thread.currentThread().getContextClassLoader()
                 .getResource("en-mnemonic-word-list.txt");
         try {
-            return Files.readAllLines(Paths.get(url.toURI()));
+            BufferedReader br = new BufferedReader(new FileReader(url.toURI().getSchemeSpecificPart()));
```
conor10: Please can this be a utility function (Android only) in the Strings class?

crypto/src/test/java/org/web3j/crypto/MnemonicUtilsTest.java
```diff
@@ -44,7 +41,12 @@
      */
     @Parameters
     public static Collection<Object[]> data() throws IOException {
-        String data = Files.lines(Paths.get(SAMPLE_FILE)).collect(Collectors.joining("\n"));
+        BufferedReader br = new BufferedReader(new FileReader(SAMPLE_FILE));
```
conor10: This should also be a utility function, but in the Files utility class.

Also, please can you ensure you have a space after the `for (` for consistency.

crypto/src/test/java/org/web3j/crypto/MnemonicUtilsTest.java
```diff
@@ -55,6 +54,12 @@
         return parameters;
     }
 
+    private static String readAllLinesWithDeliminator(String path, String delimiter)
```
conor10: `readAllLinesWithDelimiter`

```
private static String readAllLinesWithDeliminator(
        String path, String delimiter) throws IOException {
```

utils/src/main/java/org/web3j/utils/Strings.java
```diff
@@ -10,12 +10,21 @@
...
     public static String toCsv(List<String> src) {
-        // return src == null ? null : String.join(", ", src.toArray(new String[0]));
         return join(src, ", ");
     }
 
     public static String join(List<String> src, String delimiter) {
-        return src == null ? null : String.join(delimiter, src.toArray(new String[0]));
+        if (src != null) {
+            StringBuilder builder = new StringBuilder();
+            if(!src.isEmpty()) {
```
conor10: please can you include a space after the `if`

PR #398, opened 2018-03-01
Allow for dynamic gas prices and limits rather than using static values for all remote contract calls [READY]

yuriymyronovych:
Currently there is no good way to set individual gas price and gas limit for contract calls, those values are set once at contract creation. Ideally we would want to be able to specify it dynamically. Proposed solution is to use some sort of 'Provider' which can dynamically provide gas price and gas limit depending on the function called. Therefore I have added an interface ContractGasProvider.

In order to simplify function names matching from the generated contracts I have extended generators to create constants as well. All contracts in integration tests have been updated appropriately.
Generated constants look like this:

    public class Fibonacci extends Contract {

    private static final String BINARY = "..."; 

    public static final String FUNC_FIBONACCINOTIFY = "fibonacciNotify";

    public static final String FUNC_FIBONACCI = "fibonacci";

Let me know what do you think guys.

Ready for merge.


codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -245,6 +248,29 @@ private FieldSpec createBinaryDefinition(String binary) {
         return methodSpecs;
     }
 
+    Iterable<FieldSpec> buildFuncNameConstants(List<AbiDefinition> functionDefinitions) {
```
conor10: What was the reason for adding all function names as constants?

integration-tests/src/test/java/org/web3j/generated/Arrays.java
```diff
@@ -20,14 +20,20 @@
...
 
+    public static final String FUNC_RETURNARRAY = "returnArray";
```
conor10: If these constants are necessary, please can you get rid of the blank line between their definitions?

yuriymyronovych:
Hey @conor10. I would love to remove empty lines, but I don't think it is possible. It seems those empty lines are hardcoded in javapoet library... did I miss anything?
The reason we need those constants is that there is no way currently to refer to contract functions... Here is an example of method in GasPriceProvider:
```
public BigInteger getGasPrice(String contractFunc) {
switch(contractFunc) 
 case org.web3j.generated.Fibonacci.FUNC_FIBONACCINOTIFY: return BigInteger.TEN;
}
```
The benefit of this approach is that if contract changes it will fail at compilation time rather than at runtime.

conor10:
Thanks for your hard work here. Please can you write some documentation on how to use this functionality in the [transactions](https://github.com/web3j/web3j/blob/master/docs/source/transactions.rst#gas) and  [smart contracts](https://github.com/web3j/web3j/blob/master/docs/source/smart_contracts.rst) section of the docs?

Issue #406, opened 2018-03-04
support more parity rpc calls


p-s-dev:
I see some parity rpc calls are implemented, like `parity_allAccountsInfo`.  

There seem to be a few that aren't implemented, when looking at this list:
https://github.com/paritytech/parity/wiki/JSONRPC-parity-module

Is there a policy about which to implement?  Or are all valid to implement, but just haven't been coded yet?

conor10:
Feel free to add any you need. The reason they haven't all been added is due to the overhead of implementing, supporting and testing them.

PR #407, opened 2018-03-04
add rpc:parity_listAccounts which "Returns all addresses if Fat DB is…

p-s-dev:
… enabled (--fat-db), null otherwise."

reference issue #406

### DIFF
--- a/parity/src/main/java/org/web3j/protocol/parity/Parity.java
+++ b/parity/src/main/java/org/web3j/protocol/parity/Parity.java
@@ -1,5 +1,6 @@
 package org.web3j.protocol.parity;
 
+import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -26,7 +27,9 @@ public interface Parity extends Admin, Trace {
     static Parity build(Web3jService web3jService) {
         return new JsonRpc2_0Parity(web3jService);
     }
-    
+
+    Request<?, ParityAddressesResponse> parityListAccounts(BigInteger quantity);
+
     Request<?, ParityAllAccountsInfo> parityAllAccountsInfo();
     
     Request<?, BooleanResponse> parityChangePassword(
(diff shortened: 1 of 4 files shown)


parity/src/main/java/org/web3j/protocol/parity/Parity.java
```diff
@@ -26,7 +27,9 @@
     static Parity build(Web3jService web3jService) {
         return new JsonRpc2_0Parity(web3jService);
     }
-    
+
+    Request<?, ParityAddressesResponse> parityListAccounts(BigInteger quantity);
```
conor10: Shouldn't the block parameter and address tags be here too?

https://github.com/paritytech/parity/wiki/JSONRPC-parity-module#parity_listaccounts

p-s-dev:
I believe the changes requested have been made.  I think this PR is ready for another review.  Thank you.

parity/src/test/java/org/web3j/protocol/parity/RequestTest.java
```diff
@@ -133,7 +133,16 @@ public void testParityKillAccount() throws Exception {
...
     }
-    
+
+    @Test
+    public void testParityListAccounts() throws Exception {
+        BigInteger maxQuantityReturned = BigInteger.valueOf(100);
+        web3j.parityListAccounts(maxQuantityReturned, null, null).send();
+
+        verifyResult("{\"jsonrpc\":\"2.0\",\"method\":\"parity_listAccounts\","
+                + "\"params\":[100,null,null],\"id\":1}");
```
conor10: Please can the test case use non-null values. I'm happy to merge once updated with this.

p-s-dev:
first and second params are required.  api spec says second param should be null if no value:
"Address - 20 Bytes - Offset address from which the batch should start in order, or null."
third param is optional and should not exist if not used.
"Tag - (optional) integer block number, or the string 'latest', 'earliest' or 'pending'."
parity wiki moved: https://wiki.parity.io/JSONRPC-parity-module.html#parity_listaccounts

PR #433, opened 2018-03-19
Adds web3j shutdown method

### LINKED ISSUE #273: Add web3j shutdown method
As per the following [issue](https://github.com/web3j/web3j/issues/206#issuecomment-344610752), there are a number of shutdown hooks in web3j for closing the ScheduledExecutor instances. 

web3j should provide a `web3j.shutdown()` method, that calls `scheduledExecutorService.shutdown()`, so that users can control when the internal ScheduledExectorService is shutdown.

mushketyk:
Adds the `shutdown` method to the `Web3j` interface to close internal executor.

Closes #273.


conor10:
Thanks!

mushketyk:
Thank you @fcorneli !

PR #436, opened 2018-03-20
Outputs a warning when non-constant function returns a value

### LINKED ISSUE #257: Add warning to Solidity when non-constant function returns a value

mushketyk:
If a transaction function is generated and this function returns a value, generator outputs a warning message. The warning message is sent to the output using the new interface `GenerationReporter`.

Fixes #257

### DIFF
--- a/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
+++ b/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
@@ -676,8 +676,8 @@ public class SolidityFunctionWrapper extends Generator {
         if (functionDefinition.hasOutputs()) {
             //CHECKSTYLE:OFF
             reporter.report(String.format(
-                    "Definition of the function %s returns a value but is not defined as a constant function. "
-                            + "Please ensure it contains the constant modifier if you want to read the return value",
+                    "Definition of the function %s returns a value but is not defined as a view function. "
+                            + "Please ensure it contains the view modifier if you want to read the return value",
                     functionDefinition.getName()));
             //CHECKSTYLE:ON
         }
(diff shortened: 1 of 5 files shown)


codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -659,11 +668,20 @@ private static ParameterizedTypeName buildRemoteCall(TypeName typeName) {
...
 
-    private static void buildTransactionFunction(
+    private void buildTransactionFunction(
             AbiDefinition functionDefinition,
             MethodSpec.Builder methodBuilder,
             String inputParams) throws ClassNotFoundException {
 
+        if (functionDefinition.hasOutputs()) {
+            //CHECKSTYLE:OFF
+            reporter.report(String.format(
```
conor10: Now the correct term to use for constant functions is "view" as per http://solidity.readthedocs.io/en/v0.4.21/contracts.html#view-functions. Please can you update to use view instead of the constant modifier?

mushketyk:
Hi @conor10 

Thank you for your review. I've updated the PR.

conor10:
Good work. 👍

mushketyk:
Thank you @conor10 !

PR #440, opened 2018-03-21
Fixes events hashing computation

### LINKED ISSUE #209: sha3 Event hash consistency issue
Say you have an event:

```
event ReceivedETH(address indexed backer, uint amount, uint8 indexed step);
```

`EventEncoder.encode` won't create the proper sha3 hash for this Event due to the fact that there is an indexed property after an unindexed property. `EventEncoder.encode`orders the parameters by `indexed` first, which results in the hash being:

```
ReceivedETH(address,uint8,uint)
```

but the proper hash used in the Event's tokens will be based on:

```
ReceivedETH(address,uint,uint8)
```

I am admittedly new to Ethereum, but based on how I've seen everything working on the blockchain, it seems like this is a bug in Web3j?

To workaround the issue, we just rearranged the parameters on the event definition so that it is compatible with the forced ordering of Web3j:

```
event ReceivedETH(address indexed backer, uint8 indexed step, uint amount);
```

mushketyk:
Problem: the `Event` class stores parameters in two lists instead of storing them in a single list. This causes issues when an event should be hashed since parameters should be hashed in corect order (see #209)

This PR changes the `Event` and now it contains a single list of parameters. To preserve information about what parameters are indexed this PR adds the `indexed` field to the `TypeReference` class.

Fixes #209.

In this PR generated contracts like `ENS` are edited manually since I didn't find a way to regenerate them. This should be fixed before this PR is merged, but it should not impede this PR's review.


conor10:
Thanks for fixing. You should be able to regenerate the code wrappers by running the below.

`integration-tests/scripts/generateWrappers.sh`
`core/src/main/resources/solidity/build.sh`

conor10:
Once the code wrappers are regenerate I'll be happy to merge.

PR #443, opened 2018-03-22
deserialize json object in error response data field 

p-s-dev:
testrpc populates the "data" field of an error response with a json object. web3j wants the "data" field to be a string and is unable to parse it as an object.

use custom deserializer to allow error response data field to be json object.  jackson throws error when trying to deserialize json object to a string with default deserializers.


conor10:
Thanks!

PR #453, opened 2018-03-27
Fix JsonRpc2_0Geth protections

alexprut:
Fix `JsonRpc2_0Geth` protections, it should be open for further extensions.


conor10:
Thanks!

PR #455, opened 2018-03-27
import test dependencies from testArtifacts rather than archives.

p-s-dev:
Building always worked from the command line.  However, IntelliJ refused to accept the "archives" configuration as a valid source for test dependencies.  By defining a new configuration "testArtifacts" and exporting it from the subprojects that contain shared test classes, I'm able to open the project in IntelliJ and compile it with no "can not find symbol" errors in the cross-project imports.

Also Removed a duplicate TempFileProvider class and change to dependency reference.


conor10:
Thanks! 👍

PR #464, opened 2018-04-01
Admin build with ScheduledExecutorService

fcorneli:
This patch makes it possible to construct an Admin object with ScheduledExecutorService parameter.


conor10:
Thanks!

PR #465, opened 2018-04-01
Add AUR as an install option for cli

benjumanji:


conor10:
Great stuff! How will future releases be added?

benjumanji:
Currently they just need a push from me and anyone else that wants to maintain the package. It could be automated. It's a very trivial pkg build. If you were interested I could get a travis build to push it or something and eventually if that worked well it could move under the web3j umbrella?

conor10:
Certainly happy for it to be an official web3j project as long as you or someone else is prepared to maintain it on an ongoing basis.

PR #475, opened 2018-04-04
deployRemoteCall public

fcorneli:
This patch makes it easier to deploy a contract if you only have the compiled contract class available at runtime:
```
Class<? extends Contract> contractClass
```
and thus you cannot call `YourCompiledContract.deploy(...)` directly.


conor10:
Makes sense - thanks!

PR #478, opened 2018-04-05
getChainId on Transaction

fcorneli:


fcorneli:
Added a unit test.

fcorneli:
mmm... codecov... loving it. :)

core/src/test/java/org/web3j/protocol/core/ResponseTest.java
```diff
@@ -899,6 +899,30 @@ public void testEthTransaction() {
         assertThat(ethTransaction.getTransaction().get(), equalTo(transaction));
     }
 
+    @Test
+    public void testEthTransactionChainId() {
+        Transaction transaction = new Transaction(
```
conor10: Why not just create a Transaction() then use the setter? The nulls are not very nice.

Issue #477, opened 2018-04-05
why not-enough gas?


basquiat78:
<code>
		TransactionReceipt transactionReceipt = null;

		try {
		
			Credentials credentials = WalletUtils.loadCredentials(ETH_COINBASE_PASSWORD, PATH_TO_WALLET);
			TESTToken tokenContract = TESTToken.load(TESTTOKEN_CONTRACT_ADDRESS, web3j, credentials, Contract.GAS_PRICE, Contract.GAS_LIMIT);
			transactionReceipt = tokenContract.transfer(toAddress, amount).sendAsync().get();
			
		} catch (Exception e) {
			e.printStackTrace();
		}
		
		return transactionReceipt;
</code>

ok! i trasfered TESTToken to other address. but when call sendAsync().get()  below error

java.util.concurrent.ExecutionException: org.web3j.protocol.exceptions.TransactionException: Transaction has failed with status: null. Gas used: 37005. (not-enough gas?)

why not-enough gas?

BUT TESTToken was sent to other address. 

how can i set gas for contract? T^T

mushketyk:
Hi @basquiat78 
`web3j` expects a transaction result to have status equal to `0x1`, but for some reason, the status is equal to `null`. What client and what network are you using?

mushketyk:
@basquiat78 @conor10 I've created a PR that should fix the issue: https://github.com/web3j/web3j/pull/479

basquiat78:
and one more question.

Is this issue related to ens?

bellaj:
in my case there was an error in 
private static final String BINARY = "0x0....."
strangely there was an "\n" before "0x0.." after using web3j to generate the contract warpper
so removing this parasite bring thing back to normal

reuvenf:
I receive the same error:
`org.web3j.protocol.exceptions.TransactionException: Transaction has failed with status: 0x0. Gas used: 35513. (not-enough gas?)`

Please advise how to fix it.

willijk:
Maybe this is because of the web3j wrapper. I have encountered the same problem and tried to redeploy many times. Sometimes it can deploy successfully but most of the time I got the above exception. Thanks to [this](https://github.com/web3j/web3j/issues/515#issuecomment-456648080), I start to compare the bytecode generated by web3j to the bytecode generated by **remix**. It was different! After replacing the bytecode with **remix** bytecode, problem solved! At least my problem solved.

lilyhu2014:
I receive the same error: 

[...]

PR #479, opened 2018-04-05
Allow transaction receipts to have "null" status

mushketyk:
Fixes https://github.com/web3j/web3j/issues/477


conor10:
Great stuff.

PR #482, opened 2018-04-06
migrate to RxJava2

vpriscan:
Migration to RxJava2 (version 2.1.10), with minimal code modifications, without any method or class renaming, 
which means methods are still called *observable() even though they return RxJava2 Flowable.
Clients will have to update their code that expects old RxJava1 Observable to expect new RxJava2 Flowable.

**Alternative suggestion on migration requiring renaming and creating new methods:**
Dependency to 3rd party RxJava1 to RxJava2 and vice-versa interoperability library could be added so that clients who still use RxJava1 can continue using it.
Then, *observable() methods part of public API could be made to return RxJava1 Observable, using one-line RxJava2 to RxJava1 Interop library call, and methods called *flowable() that return new RxJava2 Flowable could be created.
That way client code won't break at all, until they decide to migrate to RxJava2 themselves.


core/src/main/java/org/web3j/protocol/core/RemoteCall.java
```diff
@@ -44,16 +44,7 @@ public T send() throws Exception {
...
-                subscriber -> {
-                    try {
-                        subscriber.onNext(send());
-                        subscriber.onCompleted();
-                    } catch (Exception e) {
-                        subscriber.onError(e);
-                    }
-                }
-        );
+    public Flowable<T> observable() {
```
conor10: 👍

conor10:
This is great work! Let me have a think about the API. It will require a major release of web3j for it.
That being said, it could be good to bundle up with @mushketyk's websocket work.

conor10:
I'm thinking it will make sense to ready a web3j 4.0 release with this once the websockets work is done.

vpriscan:
Websocket work uses RxJava1 in some places, depending on which pull request you merge first, someone (@mushketyk or me) will need to update those usages to RxJava2 too.

vpriscan:
@conor10 I have merged master back into rxjava2-migration branch and fixed conflicts

pgreze:
Most (all?) of JSON-RPC methods are just returning 1 value, so [Single](http://reactivex.io/RxJava/javadoc/io/reactivex/Single.html) should be favored IMO.

vpriscan:
@conor10 have you decided on the method names (*observable or *flowable) and whether [Single](http://reactivex.io/RxJava/javadoc/io/reactivex/Single.html) should be used for JSON-RPC methods?

conor10:
@snazha-blkio I'm keen to see this change in 4.0.

snazha-blkio:
No write permissions to @vpriscan's repo, so I have created a branch that pulls in changes from that branch.
The new PR is https://github.com/web3j/web3j/pull/753/files,

snazha-blkio:
Thanks @vpriscan for the hard work, this has been replaced with: https://github.com/web3j/web3j/pull/753

so we can make some changes.

vpriscan:
@snazha-blkio It has been a pleasure, keep up the good work!

PR #483, opened 2018-04-07
Compare hex transaction status as real ints not as pure strings.

### LINKED ISSUE #481: executeTransaction is using explicit string match on status check
Hi there,

On or around line 259 the following code is executed:
```
        if (!SUCCESSFUL_TRANSACTION_STATUS.equals(receipt.getStatus())) {
```
This constant is set to a final String of "0x1".

Ganache +/- CLI returns a Transaction status of 0x01. While this is equivalent in hex since this evaluation is a String it results in the following:
```
org.web3j.protocol.exceptions.TransactionException: Transaction has failed with status: 0x01. Gas used: 310277. (not-enough gas?)
	at org.web3j.tx.Contract.executeTransaction(Contract.java:259)
	at org.web3j.tx.Contract.create(Contract.java:295)
	at org.web3j.tx.Contract.deploy(Contract.java:324)
	at org.web3j.tx.Contract.lambda$deployRemoteCall$5(Contract.java:362)
	at org.web3j.protocol.core.RemoteCall.send(RemoteCall.java:30)
```
I suggest TRANSACTION_STATUS comparison is conducted as a parsed hex match rather than explicit string.

perlboy:
This should fix #481

I switch to using Integer.decode to process string values. While the constant could simply be a 1 I use the same decode method for clarity more than anything.

### DIFF
--- a/core/src/main/java/org/web3j/tx/Contract.java
+++ b/core/src/main/java/org/web3j/tx/Contract.java
@@ -43,7 +43,8 @@ public abstract class Contract extends ManagedTransaction {
 
     // https://www.reddit.com/r/ethereum/comments/5g8ia6/attention_miners_we_recommend_raising_gas_limit/
     public static final BigInteger GAS_LIMIT = BigInteger.valueOf(4_300_000);
-    public static final String SUCCESSFUL_TRANSACTION_STATUS = "0x1";
+    // keep the string definition but store it as an int, perhaps pointless?
+    public static final int SUCCESSFUL_TRANSACTION_STATUS = Integer.parseInt("0x1", 16);
 
     protected final String contractBinary;
     protected String contractAddress;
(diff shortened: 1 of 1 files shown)


fcorneli:
Maybe move this logic to `TransactionReceipt`?
```
public boolean isStatusOK() {
    if (status == null) {
        return true;
    }
    ... the rest of the logic
}
```

perlboy:
Hmm, perhaps... That code block in general seems a bit pointless and has been hacked on relatively recently (to allow null states). Declaring a constant that is used only once in the entire codebase seems a bit pointless at this stage. Right now this is a show stopped for anyone using Ganache and deploying contracts but a general refactor may be worth considering.

fcorneli:
In our projects, we use the same getStatus logic in two different places already. So having `TransactionReceipt.isStatusOK()` would serve more than just web3j internal usage.

perlboy:
I reviewed TransactionReceipt and it is ostensibly a pojo (org.web3j.protocol.core.methods.response package appears to have this purpose). Introducing logic into this directly doesn't seem compliant with style chosen.

fcorneli:
View as it as just another type of getter... some response classes already do this... getXXXRaw and getXXX... which is like a "formatted" version of the raw.

fcorneli:
According to:
https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt
the status field is of type `QUANTITY`. Maybe use `Numeric.decodeQuantity` instead?

conor10:
Please can we use Numeric.decodeQuantity as per @fcorneli's comment?

conor10:
I've merged @fcorneli updated version.

@perlboy I appreciate you taking the time to submit a PR.

PR #491, opened 2018-04-10
Transaction decoder

fcorneli:
This PR allows you to decode raw transactions, both unsigned and signed.
Furthermore, you can verify the signature on decoded signed raw transactions and retrieve the chain id used during signing of the raw transaction.


conor10:
@fcorneli Please can you provide a description for this PR?

fcorneli:
Done.

fcorneli:
Added padding fix for r and s, which can both be less than 32 bytes in size on some signatures.

conor10:
👍

PR #493, opened 2018-04-11
isStatusOK on TransactionReceipt

fcorneli:


fcorneli:
We already use this logic in 5 different places in our code.

core/src/test/java/org/web3j/protocol/core/ResponseTest.java
```diff
@@ -1061,6 +1061,63 @@ public void testeEthGetTransactionReceiptAfterByzantium() {
                 equalTo(transactionReceipt));
     }
 
+    @Test
+    public void testTransactionReceiptIsStatusOK() {
+        TransactionReceipt transactionReceipt =
+                new TransactionReceipt(
```
conor10: Again, why not construct a default instance and use a setter?

fcorneli: Using default constructor now.

Issue #471, opened 2018-04-03
IPC Request/Response threading issue


4ntoine:
I can see response JSON is forwarded to wrong thread (see request with id=623 is sent from main thread, but response to main thread came for request with id=625, obviously it can't be parsed because another response structure is expected):
```
00:45:18.603 [main] DEBUG IpcService - >> {"jsonrpc":"2.0","method":"eth_uninstallFilter","params":["0x30e2c248a5a44ace8519e1b4a0dc36c2"],"id":609}
00:45:18.614 [main] DEBUG IpcService - << {"jsonrpc":"2.0","id":609,"result":true}

00:45:18.615 [main] DEBUG IpcService - >> {"jsonrpc":"2.0","method":"eth_uninstallFilter","params":["0xfc8e6a163d3a360bc4507ba3d2303ab8"],"id":610}
00:45:18.626 [main] DEBUG IpcService - << {"jsonrpc":"2.0","id":610,"result":true}

00:45:18.626 [main] DEBUG IpcService - >> {"jsonrpc":"2.0","method":"eth_uninstallFilter","params":["0x3808d1194aed334244c1b2aed48b758e"],"id":611}
00:45:18.638 [main] DEBUG IpcService - << {"jsonrpc":"2.0","id":611,"result":true}

00:45:18.639 [main] DEBUG IpcService - >> {"jsonrpc":"2.0","method":"eth_uninstallFilter","params":["0x9db806819162cb62f6c737efd53bb2a9"],"id":612}
00:45:18.673 [main] DEBUG IpcService - << {"jsonrpc":"2.0","id":612,"result":true}

00:45:18.674 [main] DEBUG IpcService - >> {"jsonrpc":"2.0","method":"eth_uninstallFilter","params":["0x6bc4c429ba320a2ed9eb45ec5de549a7"],"id":613}
00:45:18.696 [main] DEBUG IpcService - << {"jsonrpc":"2.0","id":613,"result":true}

00:45:18.696 [pool-2-thread-1] DEBUG IpcService - >> {"jsonrpc":"2.0","method":"eth_getFilterChanges","params":["0xa93bcf648afbea2332b3f2895992b960"],"id":614}
00:45:18.711 [pool-2-thread-1] DEBUG IpcService - << {"jsonrpc":"2.0","id":614,"result":[]}

[...]

PR #494, opened 2018-04-12
Fix simultaneous requests for UNIX IPC

iongchun:
This PR attempt to solve the same problems described in #245, and #471, 
with opening one IPC connection for each request.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/ipc/IpcService.java
+++ b/core/src/main/java/org/web3j/protocol/ipc/IpcService.java
@@ -60,7 +60,8 @@ public class IpcService extends Service {
 
         // It's not ideal converting back into an inputStream, but we want
         // to be consistent with the HTTPService API.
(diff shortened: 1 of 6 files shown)


fcorneli:
Why be so nice and keep the old constructors alive while we all know that they are wrong?
Sometimes hard breaking an API helps integrating developers more...

iongchun:
Because they are public, and might still used by some simple/single-threaded applications unaffected by these issues? ;)

They are marked as deprecated anyway.
If this PR is merged, I could create another PR to remove them, which could be merged later (e.g., in next major version).

iongchun:
@conor10 Yes. I finally find some time to get synchronized on rinkeby, and runs my test application with subscriptions of mined blocks and pending transactions, and simultaneous send some transactions. It runs without exceptions for many hours/blocks ;)

conor10:
👍

Issue #403, opened 2018-03-03
 java.lang.IndexOutOfBoundsException


badboy-tian:
```
 val wallet = WalletUtils.generateBip39Wallet(password, fileDir)
                val seed = MnemonicUtils.generateSeed(wallet.mnemonic, password)
                val credentials = Credentials.create(ECKeyPair.create(sha256(seed)))
                Log.e("", Numeric.encodeQuantity(credentials.ecKeyPair.privateKey))
```

```
 W/System.err: java.lang.IndexOutOfBoundsException: Index: 1595
03-03 14:15:36.498 28860-28860/com.i7play.qunwallet W/System.err:     at java.util.Collections$EmptyList.get(Collections.java:4500)
03-03 14:15:36.498 28860-28860/com.i7play.qunwallet W/System.err:     at org.web3j.crypto.MnemonicUtils.generateMnemonic(MnemonicUtils.java:61)
03-03 14:15:36.498 28860-28860/com.i7play.qunwallet W/System.err:     at org.web3j.crypto.WalletUtils.generateBip39Wallet(WalletUtils.java:95)
03-03 14:15:36.498 28860-28860/com.i7play.qunwallet W/System.err:     at com.i7play.qunwallet.MainActivity$onCreate$1.onClick(MainActivity.kt:43)
03-03 14:15:36.498 28860-28860/com.i7play.qunwallet W/System.err:     at android.view.View.performClick(View.java:6256)
03-03 14:15:36.498 28860-28860/com.i7play.qunwallet W/System.err:     at android.view.View$PerformClick.run(View.java:24701)
03-03 14:15:36.498 28860-28860/com.i7play.qunwallet W/System.err:     at android.os.Handler.handleCallback(Handler.java:789)
03-03 14:15:36.498 28860-28860/com.i7play.qunwallet W/System.err:     at android.os.Handler.dispatchMessage(Handler.java:98)
03-03 14:15:36.498 28860-28860/com.i7play.qunwallet W/System.err:     at android.os.Looper.loop(Looper.java:164)
03-03 14:15:36.498 28860-28860/com.i7play.qunwallet W/System.err:     at android.app.ActivityThread.main(ActivityThread.java:6541)
03-03 14:15:36.498 28860-28860/com.i7play.qunwallet W/System.err:     at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #498, opened 2018-04-14
Read word list as stream and throw possible error earlier

comodoro:
Fix for #403, which was happening on at least some Android devices. Error handling is up to discussion.

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/MnemonicUtils.java
+++ b/crypto/src/main/java/org/web3j/crypto/MnemonicUtils.java
@@ -1,15 +1,16 @@
 package org.web3j.crypto;
 
-import java.net.URL;
-import java.nio.file.Files;
-import java.nio.file.Paths;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-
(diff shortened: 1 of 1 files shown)


crypto/src/main/java/org/web3j/crypto/MnemonicUtils.java
```diff
@@ -1,16 +1,17 @@
...
-import java.nio.file.Paths;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-
 import org.bouncycastle.crypto.digests.SHA512Digest;
 import org.bouncycastle.crypto.generators.PKCS5S2ParametersGenerator;
 import org.bouncycastle.crypto.params.KeyParameter;
 
+import java.io.BufferedReader;
```
conor10: Please can you keep this imports at the top?

comodoro: I'm sorry, the IDE changed the order. Is it now OK?

conor10:
Thanks!

PR #534, opened 2018-05-09
fix List<Boolean> type transfer bug

mmmmmagina:
In abi generated code,  below code will raise a error : "org.web3j.abi.TypeMappingException: java.lang.NoSuchMethodException: org.web3j.abi.datatypes.Bool.<init>(java.lang.Boolean)"

new org.web3j.abi.datatypes.DynamicArray<org.web3j.abi.datatypes.Bool>(
                        org.web3j.abi.Utils.typeMap(boolListParams, org.web3j.abi.datatypes.Bool.class)),

because of the Bool.class has contructor method with param boolean  but not Boolean


conor10:
@mmmmmagina - please can you provide a test that demonstrates this?

PR #541, opened 2018-05-12
Make http error handling more informative

kondratovich:
Add http code and response message to exception


kondratovich:
Can you check this pretty simple PR? I want to fix the situation, when underlying http error is suppressed.

conor10:
Thanks!

PR #546, opened 2018-05-16
[#545] add missing constructor of WalletUtils by using lightWalletFile default

### LINKED ISSUE #545: Missing constructor of WalletUtils with 2-param password and location
As the usage from https://docs.web3j.io/transactions.html#creating-and-working-with-wallet-files

> String fileName = WalletUtils.generateNewWalletFile(
        "your password",
        new File("/path/to/destination"));

hackty:
fix #545 add missing constructor of WalletUtils by using lightWalletFile default


hackty:
add test case later

PR #547, opened 2018-05-16
[#545] add missing constructor of WalletUtils by using fullWalletFile default

### LINKED ISSUE #545: Missing constructor of WalletUtils with 2-param password and location

hackty:
fix #545
add missing constructor of WalletUtils by using lightWalletFile default
add and optimize the testcase for it

### DIFF
--- a/crypto/src/test/java/org/web3j/crypto/WalletUtilsTest.java
+++ b/crypto/src/test/java/org/web3j/crypto/WalletUtilsTest.java
@@ -51,16 +51,23 @@ public class WalletUtilsTest {
     @Test
     public void testGenerateFullNewWalletFile() throws Exception {
         String fileName = WalletUtils.generateFullNewWalletFile(PASSWORD, tempDir);
-        testGenerateNewWalletFile(fileName);
+        testGeneratedNewWalletFile(fileName);
     }
+
+    @Test
+    public void testGenerateNewWalletFile() throws Exception {
+        String fileName = WalletUtils.generateLightNewWalletFile(PASSWORD, tempDir);
+        testGeneratedNewWalletFile(fileName);
+    }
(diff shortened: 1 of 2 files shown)


conor10:
I don't understand why we need this method? Why do we need a default?

crypto/src/test/java/org/web3j/crypto/WalletUtilsTest.java
```diff
@@ -51,16 +51,23 @@ public void testGenerateBip39Wallets() throws Exception {
     @Test
     public void testGenerateFullNewWalletFile() throws Exception {
         String fileName = WalletUtils.generateFullNewWalletFile(PASSWORD, tempDir);
-        testGenerateNewWalletFile(fileName);
+        testGeneratedNewWalletFile(fileName);
     }
 
+    @Test
+    public void testGenerateNewWalletFile() throws Exception {
+        String fileName = WalletUtils.generateLightNewWalletFile(PASSWORD, tempDir);
```
conor10: We should default to the full wallet file - see https://github.com/ethereum/go-ethereum/blob/8f8774cf6dd5bbcba9fa3773fa34d13d6523a147/node/config.go#L376

hackty: Well, it has been changed default to the full wallet file.

hackty: Well, it has been changed as the full wallet file by default.

Issue #565, opened 2018-05-30
UnixDomainSocket.read is O(N^2)


santtu:
The UnixDomainSocket used for local IPC implements an O(N^2) algorithm for reading data from the IPC socket: https://github.com/web3j/web3j/blob/master/core/src/main/java/org/web3j/protocol/ipc/UnixDomainSocket.java#L60

I created a test program (see https://gist.github.com/santtu/c47f667812051b20721e3293babd061e) and used it to scan a series of blocks in the Rinkeby testnet:

```
	$ java -jar web3jspeedtest-string.jar /Users/santtu/Library/Ethereum/geth.ipc 405000 405149
	405000 405001 405002 405003 405004 405005 405006 405007 405008 405009 405010 405011 405012 405013 405014 405015 405016 405017 405018 405019 405020 405021 405022 405023 405024 405025 405026 405027 405028 405029 405030 405031 405032 405033 405034 405035 405036 405037 405038 405039 405040 405041 405042 405043 405044 405045 405046 405047 405048 405049 405050 405051 405052 405053 405054 405055 405056 405057 405058 405059 405060 405061 405062 405063 405064 405065 405066 405067 405068 405069 405070 405071 405072 405073 405074 405075 405076 405077 405078 405079 405080 405081 405082 405083 405084 405085 405086 405087 405088 405089 405090 405091 405092 405093 405094 405095 405096 405097 405098 405099 405100 405101 405102 405103 405104 405105 405106 405107 405108 405109 405110 405111 405112 405113 405114 405115 405116 405117 405118 405119 405120 405121 405122 405123 405124 405125 405126 405127 405128 405129 405130 405131 405132 405133 405134 405135 405136 405137 405138 405139 405140 405141 405142 405143 405144 405145 405146 405147 405148  done
	149 blocks in 285.852 seconds
```

The fix is simple (I'll get PR on that soon), just replace String appends with a StringBuffer. With the fix the time taken drops to 1/50 of the String append version:

```
	$ java -jar web3jspeedtest-stringbuffer.jar /Users/santtu/Library/Ethereum/geth.ipc 405000 405149
[...]

PR #566, opened 2018-05-30
Fix O(N^2) UNIX domain socket read behavior

santtu:
See issue https://github.com/web3j/web3j/issues/565.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/ipc/UnixDomainSocket.java
+++ b/core/src/main/java/org/web3j/protocol/ipc/UnixDomainSocket.java
@@ -59,7 +59,7 @@ public class UnixDomainSocket implements IOFacade {
     @Override
     public String read() throws IOException {
         CharBuffer response = CharBuffer.allocate(bufferSize);
-        StringBuffer result = new StringBuffer();
+        StringBuilder result = new StringBuilder();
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/web3j/protocol/ipc/UnixDomainSocket.java
```diff
@@ -59,15 +59,15 @@ public void write(String payload) throws IOException {
     @Override
     public String read() throws IOException {
         CharBuffer response = CharBuffer.allocate(bufferSize);
-        String result = "";
+        StringBuffer result = new StringBuffer();
```
conor10: Please can you change this to use StringBuilder?

santtu:
Used StringBuilder instead now. I toyed around increasing the buffer size but did not see any such significant differences in performance between stringbuilder, stringbuffer, and different read buffer sizes to justify any additional changes.

core/src/main/java/org/web3j/protocol/ipc/UnixDomainSocket.java
santtu: Done, didn't that change show in the tracker?

conor10:
Thanks!

PR #571, opened 2018-06-01
Remove unused walletFile.setCrypto(crypto)

thundertrick:
On line 122, `walletFile.setCrypto(crypto)` is called again, which will override the first call. I suggest to remove the first call.


conor10:
Makes sense - thanks!

PR #584, opened 2018-06-08
Trustless tx hash computation

yuriymyronovych:
Addressing a philosophical issue rather than technical here.

Problem: when we use infura or similar service to execute our transactions we rely on txHash returned and trust third party to give us correct value. Ideally we would want to follow blockchain concepts of 'client in control' and 'trustless execution'. Therefore, proposed change is to calculate txHash locally before submitting it to node and verify that returned hash matches the one we expect.

Note: at Argent we did have rare issues with infura not returning txHash or returning empty results. I think this change would be highly desirable  by anyone using web3j in production especially for fintech products.


conor10:
Definitely a useful addition to have - apologies for the delay in merging.

amitOodles:
@conor10  @yuriymyronovych 
"Note: at Argent we did have rare issues with infura not returning txHash or returning empty results. I think this change would be highly desirable by anyone using web3j in production especially for fintech products."

When there was no txHash , was a transaction being generated by Infura ?

I am facing an issue using web3j , in which a transaction is being generated on geth ,  but EthSendTransaction.getTransactionHash() returns null.

Which results in a transaction being sent on the network but not saved in the database. 

Here is a check i make on EthSendTransaction respnse - 
if(response != null && response.getTransactionHash != null){
     "Hash generated"
}else{
     "Error generating tx hash "
}

the code ends up in the else part even when when geth has generated a transaction .

yuriymyronovych:
@amitOodles can you provide additional details? What is the json response from the node when that happens? At which point is response object lost?

amitOodles:
@yuriymyronovych @conor10 
I do not have any additional information , but 

Geth server log 
INFO [09-08|05:57:46] Submitted transaction fullhash=0x61baad408420cd15c480854ec3519e8f94a26a7df5b39fee33644298a25e839b 

and on the app server i have
08-09-2018 05:57:55 EthereumTransactionService.sendTransactionAsync - Unable to generate transaction hash for ASYNC ETH txn with id 6714784

I only have the above logs , and i have faced similar condition more than once , i have just come to know about this issue.

Is the if expression faulty ? Also i have noticed some time no transaction hash was generated by geth server also .

amitOodles:
@yuriymyronovych 
Geth version - 1.8.11 and web3j version 3.4.0

amitOodles:
@yuriymyronovych 

Ok , i will try to find further details. Regarding this PR only ,

Suppose i use the RawTransactionManager to send transactions in web3j , in what scenario will the transaction hash be different ?

Using the Transcoder.signMessage i can get the signedmessage as

signedMessage = TransactionEncoder.signMessage(rawTransaction, chainId, credentials);

then i can get the raw transaction value as 

String hexValue = Numeric.toHexString(signedMessage);

and then using the following function i can get the transaction hash as follows

String txHashLocal = Hash.sha3(hexValue);

So i wanted to know , in what case the transaction hash returned from geth will be different to the one calculated above ?

yuriymyronovych:
@amitOodles It should always be the same.

amitOodles:
@yuriymyronovych 
@conor10 
So i can wait for the 
web3j.ethSendRawTransaction response ,
EthSendTransaction txResponse = web3j.ethSendRawTransaction(hexValue).send();
and if the response object satisfies the following criteria

if(txResponse != null && !txResponse.hasErrors)
Then i can save the transaction hash in the database and do further work. Ok ?

Or do i have to make further checks on the txResponse object ?

tjanson:
I’m using ganache-cli (`Ganache CLI v6.2.3 (ganache-core: 2.3.1)`) for testing and am getting `TxHashMismatchException` on seemingly every transaction (not 100% sure).

This did not happen previously, but I changed a lot of things (migrated to truffle 5, web3.js 1.0, Solidity 0.5).

Anything I can do to debug this?

yuriymyronovych:
@tjanson try to look inside the exception data, it should have information regarding why hashes are different...

tjanson:
Well, I can see that the hashes are different of course, but that doesn’t tell me much. I’m not specifying a `chainId`, so the `signMessage(RawTransaction, Credentials)` method is used in `signAndSend` [here](https://github.com/web3j/web3j/blob/7641f5d529ba400c25dcfe8288e04d0fc1b8747a/core/src/main/java/org/web3j/tx/RawTransactionManager.java#L115).

I can post some of the `RawTransaction`s, it’s not secret stuff.

I’m also okay with just ignoring the error / disabling the `txHashVerifier` (as far as I can tell everying runs perfectly fine), I just wanted to let you know about the issue and am curious if others encounter the same thing.

yuriymyronovych:
thanks for letting us know. It is hard to say why you are getting it. I can say it works for us on ropsten and mainet with infura. We never tried ganache though, maybe they do something funky with transaction or hash of it...

Issue #402, opened 2018-03-02
NPE at org.web3j.utils.Numeric.containsHexPrefix(Numeric.java:80)


asoldano:
Hi,
I'm getting a NPE when calling a view function in a contract that returns an out-of-gas error. Using v. 3.3.1 and ganache-cli locally.
It looks like FunctionReturnDecoder/Numeric need to return null in such case, so that at least the client gets a decent error message.

java.lang.NullPointerException
	at org.web3j.utils.Numeric.containsHexPrefix(Numeric.java:80)
	at org.web3j.utils.Numeric.cleanHexPrefix(Numeric.java:64)
	at org.web3j.abi.FunctionReturnDecoder.decode(FunctionReturnDecoder.java:37)
	at org.web3j.tx.Contract.executeCall(Contract.java:178)
	at org.web3j.tx.Contract.executeCallSingleValueReturn(Contract.java:184)
	at org.web3j.tx.Contract.executeCallSingleValueReturn(Contract.java:195)
	at org.web3j.tx.Contract.lambda$executeRemoteCallSingleValueReturn$1(Contract.java:251)
	at org.web3j.protocol.core.RemoteCall.send(RemoteCall.java:30)
        [...]

VanLuong1996:
do you fix it

AlexandrouR:
Looks like this issue was resolved and merged.

PR #585, opened 2018-06-08
fixed NPE when calling a view function in a contract that returns an out-of-gas error, issue 402

yuriymyronovych:
https://github.com/web3j/web3j/issues/402

@asoldano @iikirilov how does this look? I believe it should fix the issue.


conor10:
Great - thanks 👍

Issue #595, opened 2018-06-17
IndexOutOufBoundsException in MnemonicUtils.java in android branch version 3.3.1


mnl-assasin:
Hi everyone, 

I applied the fixed from master branch in the issue of returning an empty set of WORD_LIST
https://github.com/web3j/web3j/pull/594

Got the changes from here:
https://github.com/web3j/web3j/blob/master/crypto/src/main/java/org/web3j/crypto/MnemonicUtils.java

I only actually copied and pasted that in android branch and send a pull request.
I also think that it was only forgotten to applied it in the android branch

Thanks, hope you can review and approved my PR

mnl-assasin:
Hey there,

My PR only contains MnemonicUtils.java from master branch copied to android branch. 
If my PR fails for unknown reason. Can someone fixed the issue why it fails and send new PR?

We can't use generateMnemonic(byte[] initialEntropy) method because it throws a IndexOutOfBoundsException probably because WORD_LIST returns an empty list

Solution comes from here:
https://github.com/web3j/web3j/issues/403

ehanoc:
@mnl-assasin  i had the same problem. I've merged MnemonicUtils from the master branch into android and fixed the import issues; spongycastle (android) vs bouncycastle(master)

AlexandrouR:
seems this was addressed in the 4.1.0

PR #598, opened 2018-06-20
MnemonicUtils fails to build WORD_LIST in android

ehanoc:
A similar fix for this was added in #498 but was merged into master and not into the android branch. 

This was highlighted in #595 

MnemonicUtilsTest, updated too


M0rrisChang:
any reviewers?

ehanoc:
> any reviewers?

Good point @M0rrisChang . I think a lot of people want a stable Android branch to develop on android and not wait for a release for fixes like these.

snazha-blkio:
4.1.0-android has just been released and addresses this issue.

magicgoose:
There's still a similar issue on 4.2.0-android - `inputStream` ends up = null, `populateWordList` throws an exception, this makes `MnemonicUtils` unusable.

There should be some more flexible and simple way to provide the word list, IMO.

Issue #359, opened 2018-02-08
eth_getFilterChanges should rebuild the filter if the filter is invalid


iikirilov:
This affects observables that poll for new data from the node using eth_getFilterChanges. Currently, if an exception is thrown the observable tries again, which is fine if there was some `SocketException`. However, if the node was to be restarted, the filter becomes invalid but web3j continues to poll with the invalid filter indefinitely.

The [Filter](https://github.com/web3j/web3j/blob/master/core/src/main/java/org/web3j/protocol/core/filters/Filter.java) class is throwing the exception.

Regarding the following exception:

```
15:26:26.205 [pool-2-thread-2] DEBUG org.web3j.protocol.http.HttpService - --> POST <MYSECRETNODE> http/1.1
15:26:26.205 [pool-2-thread-2] DEBUG org.web3j.protocol.http.HttpService - Content-Type: application/json; charset=utf-8
15:26:26.205 [pool-2-thread-2] DEBUG org.web3j.protocol.http.HttpService - Content-Length: 109
15:26:26.205 [pool-2-thread-2] DEBUG org.web3j.protocol.http.HttpService - 
15:26:26.205 [pool-2-thread-2] DEBUG org.web3j.protocol.http.HttpService - {"jsonrpc":"2.0","method":"eth_getFilterChanges","params":["0x3ca672eecb7329eaac5d2ee0c55a6f46"],"id":828471}
15:26:26.205 [pool-2-thread-2] DEBUG org.web3j.protocol.http.HttpService - --> END POST (109-byte body)
15:26:26.348 [pool-2-thread-2] DEBUG org.web3j.protocol.http.HttpService - <-- 200 OK <MYSECRETNODE> (143ms)
15:26:26.348 [pool-2-thread-2] DEBUG org.web3j.protocol.http.HttpService - Content-Type: application/json
15:26:26.348 [pool-2-thread-2] DEBUG org.web3j.protocol.http.HttpService - Date: Tue, 16 Jan 2018 15:26:26 GMT
15:26:26.348 [pool-2-thread-2] DEBUG org.web3j.protocol.http.HttpService - Content-Length: 83
15:26:26.348 [pool-2-thread-2] DEBUG org.web3j.protocol.http.HttpService - 
15:26:26.348 [pool-2-thread-2] DEBUG org.web3j.protocol.http.HttpService - {"jsonrpc":"2.0","id":828471,"error":{"code":-32000,"message":"filter not found"}}
[...]

PR #604, opened 2018-06-23
eth_getFilterChanges should rebuild the filter if the filter is invalid #359

upers:
Rebuild filter when got RPC error code: -32000(Filter not found).#core


upers:
What's wrong with my commit?

core/src/main/java/org/web3j/protocol/core/RpcErrors.java
```diff
@@ -0,0 +1,13 @@
+package org.web3j.protocol.core;
+
+public interface RpcErrors {
```
conor10: Please can you use a class for constants not an interface.

PR #607, opened 2018-06-25
Rebuild the filter if the filter is invalid #359

upers:
Rebuilding the filter if the node returned -32000 code(Filter not found)

### DIFF
--- a/core/src/main/java/org/web3j/protocol/core/filters/Filter.java
+++ b/core/src/main/java/org/web3j/protocol/core/filters/Filter.java
@@ -16,6 +16,8 @@ import org.web3j.protocol.Web3j;
 
 import org.web3j.protocol.core.Request;
 import org.web3j.protocol.core.Response;
+import org.web3j.protocol.core.Response.Error;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/web3j/protocol/core/filters/Filter.java
```diff
@@ -110,7 +118,14 @@ private void pollFilter(EthFilter ethFilter) {
...
         if (ethLog.hasError()) {
-            throwException(ethLog.getError());
+            Error error = ethLog.getError();
+            switch (error.getCode()) {
+                case RpcErrors.FILTER_NOT_FOUND: reinstallFilter();
+                    break;
+                default:
+                    break;
+            }
+            throwException(error);
```
conor10: We shouldn't be throwing the exception if we're reinstalling the filter.

It should log a warning that the filter has not been found indicating a possible client disconnect.

upers: https://github.com/web3j/web3j/pull/609

PR #609, opened 2018-06-26
Rebuild the filter if the filter is invalid

upers:
Rebuilding the filter if the node returned -32000 code(Filter not found) #359


conor10:
One final request - please can you add a unit test for this? See core/src/test/java/org/web3j/protocol/core/filters/FilterTester.java

upers:
@conor10 
https://github.com/web3j/web3j/pull/614

PR #612, opened 2018-06-28
modify rawTransaction decode.

TOTOROYang:
get r s. use Numeric.toBytesPadded to add zero padding.
remove zeroPadded method in this TransactionDecoder class


conor10:
Thanks!

Issue #324, opened 2018-01-19
Only transactional methods should have void return types


berlinguyinca:
Dear all,

using the latest version of the generator (3.2.0, I think), I keep encountering this error:

Only transactional methods should have void return types
        at org.web3j.codegen.SolidityFunctionWrapper.buildConstantFunction(SolidityFunctionWrapper.java:524)
        at org.web3j.codegen.SolidityFunctionWrapper.buildFunction(SolidityFunctionWrapper.java:505)
        at org.web3j.codegen.SolidityFunctionWrapper.buildFunctionDefinitions(SolidityFunctionWrapper.java:209)
        at org.web3j.codegen.SolidityFunctionWrapper.generateJavaFiles(SolidityFunctionWrapper.java:108)
        at org.web3j.codegen.TruffleJsonFunctionWrapperGenerator.generate(TruffleJsonFunctionWrapperGenerator.java:134)
        at org.web3j.codegen.TruffleJsonFunctionWrapperGenerator.main(TruffleJsonFunctionWrapperGenerator.java:98)

I provide a really simple json file, generate by truffle, which is based on open zeppelins mintable token. 

Anybody has any idea for this error?

kind regards

berlinguyinca:
**example code:**

pragma solidity ^0.4.18;

import 'zeppelin-solidity/contracts/token/MintableToken.sol';

import 'zeppelin-solidity/contracts/ownership/rbac/RBAC.sol';

contract ABCToken is MintableToken, RBAC{

  string public name = "ABC";
  string public symbol = "ABC";
  uint8 public decimals = 18;

  function ABCToken() public {

  }
  /**
   * @dev Throws if called by any user, who is not a ROLE_ADMIN
   */
  modifier onlyOwner() {
    checkRole(msg.sender, ROLE_ADMIN);
    _;
  }

}

perlboy:
+1 on this, having the same issue, investigating now. @berlinguyinca this appears to be to do with the zeppelin-solidity RBAC import in particular. Unfortunately web3j doesn't provide much info on specific logging. Will update once I have more info.

perlboy:
@berlinguyinca Proposed change submitted to zeppelin-solidity.

If you wish to avoid waiting for the change to be released you can use our fork here: https://github.com/ecaustralia/zeppelin-solidity

frangio:
@perlboy Please do not encourage using a fork with commits that have not been released yet!

perlboy:
[...]

PR #642, opened 2018-07-21
Do not generate methods for constant functions that return void

macalinao:
Fixes https://github.com/web3j/web3j/issues/324 -- this is needed to use OpenZeppelin's RBAC.


conor10:
👍

Issue #562, opened 2018-05-29
A big bug in TransactionEncode.encode()!!!


Jesse0722:
Hello:
I use  TransactionEncoder.encode to encode RawTransaction, but when i decode the hexString,the value was incorrect.
Here is my test:
```
public static void main(String[] args) {
      
        RawTransaction tx = RawTransaction.createEtherTransaction(
                new BigInteger("228"),//nonce
                new BigInteger("19000000000"),//gasprice
                new BigInteger("21000"),//limit
                "0xdcb988e15807c8232920309fb93e56dc63b785a0",//to
                new BigInteger("16406500000000000000"));//value
        String hex = Numeric.toHexString(TransactionEncoder.encode(tx));
        RawTransaction tx1 = TransactionDecoder.decode(hex);
        System.out.println("Nonce:"+tx1.getNonce()+"\n"+
                "gasprice:"+tx1.getGasPrice()+"\n"+
                "limit:"+tx1.getGasLimit()+"\n"+
                "value:"+tx1.getValue()+"\n");
    }
```
The console:
```
Nonce:-28
gasprice:19000000000
limit:21000
value:-2040244073709551616
```
The result of nonce and value  are incorret!!!

And I see the source code : 
   
```
public static RlpString create(BigInteger value) {
        // RLP encoding only supports positive integer values
        if (value.signum() < 1) {
            return new RlpString(EMPTY);
        } else {
            byte[] bytes = value.toByteArray();
            if (bytes[0] == 0) {  // remove leading zero
                return new RlpString(Arrays.copyOfRange(bytes, 1, bytes.length));
            } else {
                return new RlpString(bytes);
            }
        }
    }
```
Here the code  remove leading zero, so when I use TransactionDecode to decode the hexTransaction ,the result is incorrect.

Please help me to resolve this problem.
Hope to reply me quickly.
Thank you very much.

caibinbinno1:
hello:
i don't know if you solve this problem or bug.  it's easy to solve this. 
we just padding the zero when the first index of bytes is small than 0.
just about nonce,gasPrice,gasLimit and value。 
maybe they just think them absolute are positive number.

folex:
Okay, you can reproduce it like that:
`RlpString.create(BigInt(3000000000L).bigInteger).getBytes` (Scala)
result will be
`Array(-78, -48, 94, 0)`
which is **CLEARLY** invalid:
[...]

PR #648, opened 2018-07-26
Fix bug in RlpString.asBigInteger

folex:
Address https://github.com/web3j/web3j/issues/562


conor10:
Thanks for this - an important issue.

donald-jackson:
thanks for this patch, I was about to also add a pull request :) given the severity of this issue, can this warrant an immediate minor release?

Issue #575, opened 2018-06-04
The signMessage operation does not prefix messages (security vulnerability)


ricmoo:
The [signMessage](https://github.com/web3j/web3j/blob/master/crypto/src/main/java/org/web3j/crypto/Sign.java#L47) operation does not correctly prefix the message with `"\19Ethereum Signed Message:\n" + len(message)`, which means it does not inter-operate with any other library or Ethereum node.

It is also a serious security vulnerability, since I could request the signing of a message such as `0xbf0067909b4332f01c3ab9e228857a4cd61b39358605b2cebdfb1ea1d35e0d44`, which may seem odd, but as the UI states I am simply signing a message, it is not obvious I am actually signing a transaction to send myself 1 ether from that account.

The reason for prefixing is that it enforces that the signed data is invalid as a transaction.

See [ethers.js](https://github.com/ethers-io/ethers.js/blob/master/wallet/wallet.js#L315) for an example of how to hash messages for signing.

Here is the specification of [eth_sign](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign).

The following issues are related:
- https://github.com/ethers-io/ethers.js/issues/190 (how I originally discovered this issue)
- https://github.com/web3j/web3j/issues/486
- https://github.com/web3j/web3j/issues/476
- https://github.com/web3j/web3j/issues/347

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 300.0 DAI (300.0 USD @ $1.0/DAI) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/web3j/web3j/575/676).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $26,254.78 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 4 weeks, 1 day from now. Please review their questions below:


[...]

PR #650, opened 2018-07-28
[WIP] Prefix message signing according to eth_sign. (#575)

ricmoo:
As we discussed before, I have been unable to run the ./gradlew integration tests. But opening a PR so it is easier to get this discussion going. :)

I've updated the test cases by using the key pair to sign a message using ethers. I will be adding more test cases to ethers soon, generated by parity and geth.

I am also concerned that any dapp that has previously signed a message with web3j will no longer get the correct key back when calling `signedMessageToKey`, so this is at best, something that should probably be a major version change. The other option is to add a `signedPrefixedMessageToKey` and a `signPrefixedMessage` and deprecate the non-prefixed version.

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/Sign.java
+++ b/crypto/src/main/java/org/web3j/crypto/Sign.java
@@ -56,17 +56,15 @@ public class Sign {
         return signMessage(message, keyPair, true);
     }
 
-    // WIP - RicMoo: When is needToHash used? Should it not always be true?
-    //
-    // I'm also concerned anyone who has historically used this function will
-    // now get different results, if they try to recover the address. Should
-    // there be a new recover function?
+    public static SignatureData signPrefixedMessage(byte[] message, ECKeyPair keyPair) {
+        return signMessage(getMessageHash(message), keyPair, false);
+    }
 
     public static SignatureData signMessage(byte[] message, ECKeyPair keyPair, boolean needToHash) {
         BigInteger publicKey = keyPair.getPublicKey();
         byte[] messageHash;
         if (needToHash) {
-            messageHash = getMessageHash(message);
+            messageHash = Hash.sha3(message);
         } else {
             messageHash = message;
         }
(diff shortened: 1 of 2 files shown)


ricmoo:
Ah. After Travis CI, I now see how the needToHash works... I think I have an idea, one sec. :)

ricmoo:
There are now separate messages for `signMessage` and `signPrefixedMessages` (and `signedMessageToKey` and `signedPrefixedMessageToKey`). This should preserve backward compatibility, since I see the `signMessage` is used for signing transaction RLP.

The documentation should point out that `signPrefixedMessage` should be used to emulate signing messages (e.g. `eth_sign`).

conor10:
Thanks @ricmoo. So should `signPrefixedMessages` only be used in `eth_sign`, not in transactions signed by web3j?

ricmoo:
Currently the signPrefixedMessage does the same thing as [eth_sign](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign) for messages, so it can be used in  lieu of calling the eth_sign method of a Geth node, if you have the private key.

The current signMessage, is what signing transactions requires.

Ideally we would do the following:

1. Rename signMessage to signDigest (and pass in the Hash.sha3 instead of compute it inside)
2. Rename signPrefixedMessage to signMessage

But that would cause some non-backwards compatible changes. Maybe something that can be rolled into your next major release?

snazha-blkio:
I've changed the target branch to 4.0, as recommended by @iikirilov

ricmoo:
I agree with that plan. :)

crypto/src/main/java/org/web3j/crypto/Sign.java
```diff
@@ -36,6 +36,26 @@
             CURVE_PARAMS.getCurve(), CURVE_PARAMS.getG(), CURVE_PARAMS.getN(), CURVE_PARAMS.getH());
     static final BigInteger HALF_CURVE_ORDER = CURVE_PARAMS.getN().shiftRight(1);
 
+    static final String MessagePrefix = "\u0019Ethereum Signed Message:\n";
```
conor10: This should be MESSAGE_PREFIX

crypto/src/main/java/org/web3j/crypto/Sign.java
```diff
@@ -36,6 +36,26 @@
...
+
+    static byte[] getMessagePrefix(int messageLength) {
+        return MessagePrefix.concat(String.valueOf(messageLength)).getBytes();
+    }
+
+    static byte[] getMessageHash(byte[] message) {
```
conor10: Should we call this `etEthereumMessageHash` instead? As it is an Ethereum specific method for hashing messages.

snazha-blkio: Address in: https://github.com/web3j/web3j/pull/761/

crypto/src/main/java/org/web3j/crypto/Sign.java
snazha-blkio: Addressed in https://github.com/web3j/web3j/pull/761/

snazha-blkio:
Hi @ricmoo 
We have created a separate branch from this fork to address some PR feedback, we have now merged the changes into the release/4.0 branch
c.f. https://github.com/web3j/web3j/pull/761

Thanks for the contribution, great work!

PR #659, opened 2018-08-03
Added ERC20, ERC721 Contract 

yuriymyronovych:
Currently there is no way to interact with ERC20 tokens using web3j unless you do a bit of extra work importing contract. I believe this is something we can include into web3j provided ERC20, ERC721 is a settled standard now. 
I simply finished the work started in [issue_213](https://github.com/web3j/web3j/issues/213) by generating ERC20Contract following the same rules as any other contract would.

Example how to use it:
```java
public RemoteCall<BigInteger> getERC20Balance(String ercContract, String holder) {
    ERC20 erc = ERC20.load(ercContract, web3j, txManager, BigInteger.ZERO, BigInteger.ZERO);
    return erc.balanceOf(holder);
}
```


conor10:
These implementations should be based on the documented [ERC20 standard](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-20.md). The link to the official ERC should be included in the class Javadoc at the start. It should only contain those methods defined in the ERC. 

These should reside in a new module named `eips`

config/checkstyle/suppressions.xml
```diff
@@ -6,4 +6,5 @@
 
 <suppressions>
     <suppress files="[\\/]generated[\\/]" checks="."/>
+    <suppress files="ERC20Contract.java" checks="."/>
```
conor10: As its handcrafted, it shouldn't need to be added to the suppressions in Checkstyle.

yuriymyronovych:
@conor10 will get it done today/tomorrow!

yuriymyronovych:
@conor10 @iikirilov major rework here:
 - create eips module as requested and organised all the work there. EIPS module is built as a separate jar.
 - added solidity interfaces for ERC20, ERC165, ERC721 that are used to generate contract wrappers using scripts/generateWrappers.sh. Java wrappers are generated and included in PR, there is not custom code in there, only auto-generated code.
 - added references to original EIPs as requested 
 - updated docs 

Questions:
1. Do you want to keep it as a separate jar and distribute it as optional extension library? Do I need to do anything else for that? I am not sure I did it correctly so far(added dependancy in console/build.gradle)
2. Could not think of a good way to test, the only way I can imagine would be to add integration-tests that would query hardcoded tokens/addresses to check name/symbol/supply/balance. However that would be tied up to a specific chain and would fail if executed on any other chain.

iikirilov:
> Questions:
> 
> 1. Do you want to keep it as a separate jar and distribute it as optional extension library? Do I need to do anything else for that? I am not sure I did it correctly so far(added dependancy in console/build.gradle)
> 2. Could not think of a good way to test, the only way I can imagine would be to add integration-tests that would query hardcoded tokens/addresses to check name/symbol/supply/balance. However that would be tied up to a specific chain and would fail if executed on any other chain.

1. No, as long as they are in a separate module.
2. Yes, an IT test would be awesome. We are currently working on how you can do this without hardcoding values. See #661 #779 

We will wait for IT before merge.

yuriymyronovych:
Thank you for the tip, will do.

conor10:
I'm interested in getting this merged.

@yuriymyronovych would you be willing to change the module name to contracts  (and update the docs) then we can pulll it in? I don't think EIP is the most appropriate name for the module, given it's supporting contract standards.

yuriymyronovych:
@conor10 sure, will get it done by the end of today.

yuriymyronovych:
@conor10 all the renaming and build is taken care of.

SteadyAction:
@yuriymyronovych  why is there "weiValue" parameter for ERC721.transferFrom method?
There is no such parameter in the spec. http://erc721.org/

What should I put for the parameter? can you explain please?

yuriymyronovych:
@SteadyAction It is not part of the ERC721 interface, it is simply ether value that will be transferred with this transaction. Not sure why web3j generates those now, but simply put BigInteger.ZERO in there.

conor10:
That's not good - @iikirilov or @snazha-blkio is this a bug in the smart contract codegen?

yuriymyronovych:
@conor10 , I have looked into it in more details, weiValue is added as parameter when solidity function is marked as 'payable' which is correct logic in web3j. 
Here is good explanation why it was done that way:
https://ethereum.stackexchange.com/questions/42466/why-are-the-transfer-and-approve-functions-payable-in-the-erc-721-spec

conor10:
Thanks @yuriymyronovych I'd missed that. Open Zepplin don't make it payable in their implementation, but we want to be in line with the spec so this is all good. Thanks!

PR #660, opened 2018-08-04
Fixes #656 usage strings to be in sync with command parsing

### LINKED ISSUE #656: usage say "solidity generate .... --output ..." while code uses "--outputDir"
In line 23 of https://github.com/web3j/web3j/blob/master/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapperGenerator.java, USAGE mentions "--output" as parameter definition while line 71 parses for "--outputDir".

So, using --output in "web3j solidity generate ... --output .... " generates an error while "--outputDir" is ok.

Either USAGE String or parsing shall be fixed. Thanks!

taivokasper:
Usage string and actual usage were not in sync. I think changing the usage to reflect actual usage is the only way to go for it to stay backwards compatible.


taivokasper:
Is the above fail a fluke? Running the same command locally passes.

PR #661, opened 2018-08-05
Fixed failing integration-tests

franz-see:
This is a fix for the failing integration tests. This modification also adds integration tests to travis. Here's my latest travis build - https://travis-ci.org/franz-see/web3j/builds/412356770

*Changes*
Some test data had to be changed in order to get the tests passing. Also, I've modified how some of the test works. 

For example I introduced a `TestParameters` class which can be configured via `integration-tests/local-test.properties` (_useful during development but not in travis_), via system property and via environment variable (_useful for travis_). 

```
public class TestParameters {
    public static final String ALICE_PRIVKEY;
    public static final String ALICE_PUBKEY;
    public static final boolean BENCHMARK_TESTING_ON;
    public static final String BOB_PRIVKEY;
    public static final String BOB_PUBKEY;
    public static final int FAST_RAW_TRANSACTION_MANAGER_IT_COUNT;
    public static final String TEST_KOVAN_URL;
    public static final String TEST_RINKEBY_URL;
    public static final String TEST_ROPSTEN_URL;
```

And for the tests that do transfers, I've modified them so that the tests now do transfers from ALICE to BOB (_how they always were_) and transfers from BOB to ALICE (_addition_) to minimize the changes in the balances of those accounts.

Lastly, another big change is that I added assumptions on some tests so that those tests will not run and fail if the assumptions are not met (i.e. if the test is running against infura, some methods are not supported, thus, the assumptions would fail and not even run the test).

_Note: Fails can still fail because of disconnection, etc_


franz-see:
I've configured my travis to have the following environment variables

 * WEB3J_ALICE_PRIVKEY  
 * WEB3J_ALICE_PUBKEY
 * WEB3J_BENCHMARK_TESTING_ON
 * WEB3J_BOB_PRIVKEY
 * WEB3J_BOB_PUBKEY
 * WEB3J_TEST_ROPSTEN_URL

I can email my configuration.

franz-see:
Sorry, forgot to mention - the build for this works if pointed to an infura URL ropsten. This probably need more work though to get it running on a local geth and parity.

franz-see:
I just updated the PR to include fixes on the tests so that they work against an actual geth and parity (instead of infura).

In my travis, I added an env variable WEB3J_TEST_KOVAN_URL and WEB3J_TEST_RINKEBY_URL to point to a remote geth and parity instances that I am hosting

franz-see:
This is the link to my travis-ci build for my last commit in this PR - https://travis-ci.org/franz-see/web3j/builds/413977338

And here's the report on my local
integration tests results

integration-tests/src/test/java/org/web3j/protocol/scenarios/BalanceHelper.java
```diff
@@ -0,0 +1,93 @@
...
+import org.web3j.utils.Convert;
+
+import static org.web3j.protocol.scenarios.Scenario.ALICE;
+import static org.web3j.protocol.scenarios.Scenario.BOB;
+
+/**
+ * This rebalances the balance of ALICE and BOB equally amongst themselves.
+ *
+ * <p>The tests should be created so that any transfer from ALICE to BOB should be offset by a
+ * transfer from BOB to ALICE to minimize the likelihood that one of them would ran out of
```
iikirilov: "run" instead of "ran"

integration-tests/src/test/java/org/web3j/ens/EnsIT.java
```diff
@@ -3,20 +3,35 @@
...
+                web3j.ethSyncing().send().isSyncing(),
+                allOf(
+                        notNullValue(),
+                        is(false)
+                ));
 
+        EnsResolver ensResolver = new EnsResolver(web3j);
         assertThat(ensResolver.resolve("web3j.test"),
-                is("0x19e03255f667bdfd50a32722df860b1eeaf4d635"));
+                is("0xe7fdbbac9325508269d772982cc68103f79d9693"));
```
iikirilov: http://docs.ens.domains/en/latest/introduction.html#ens-on-ethereum

web3j.test will expire every 28 days - how will this affect this test? is worth getting the permanent web3j.eth?

integration-tests/src/test/java/org/web3j/protocol/core/CoreIT.java
```diff
@@ -119,6 +146,10 @@ public void testEthSyncing() throws Exception {
 
     @Test
     public void testEthCoinbase() throws Exception {
+        assumeFalse("Infura does NOT support eth_coinbase - "
```
iikirilov: I like this pattern

integration-tests/src/test/java/org/web3j/protocol/core/CoreIT.java
```diff
@@ -207,27 +271,44 @@ public void testEthGetUncleCountByBlockHash() throws Exception {
     public void testEthGetUncleCountByBlockNumber() throws Exception {
         EthGetUncleCountByBlockNumber ethGetUncleCountByBlockNumber =
                 web3j.ethGetUncleCountByBlockNumber(
-                        DefaultBlockParameter.valueOf("latest")).send();
+                        DefaultBlockParameter.valueOf(new BigInteger("232"))).send();
```
iikirilov: test will fail if node has not synced block 232 - how likely is this? better make the test bulletproof

integration-tests/src/test/java/org/web3j/ens/EnsIT.java
iikirilov: the test will fail if ran against kovan or rinkeby

integration-tests/src/test/java/org/web3j/protocol/core/TestnetConfig.java
```diff
@@ -16,13 +16,13 @@
```
iikirilov: tests using this will fail on kovan and rinkeby

integration-tests/src/test/java/org/web3j/protocol/scenarios/EventFilterIT.java
```diff
@@ -35,11 +37,11 @@
 public class EventFilterIT extends Scenario {
 
     // Deployed Fibonacci contract instance in testnet
-    private static final String CONTRACT_ADDRESS = "0x3c05b2564139fb55820b18b72e94b2178eaace7d";
+    private static final String CONTRACT_ADDRESS = "0x581f4ef871aef69324cff1df9708678e2bff9870";
```
iikirilov: tests using this will fail on kovan and rinkeby

snazha-blkio:
This has been replaced with: https://github.com/web3j/web3j/pull/779

PR #679, opened 2018-08-23
make methods public

UnknownEnergy:
I am using some kind of hardware wallet and I need those methods to be public for signing/sending transactions via card. Is there any specific reason why they are private?

I need the encode method for the second encode after reading out r,s and v values and calling TransactionEncoder.createEip155SignatureData.

The recoverFromSignature is needed for getting the V byte.


conor10:
@UnknownEnergy I presume you made this change as the Android branch hasn't caught up to master yet? It should be brought up to date shortly.

UnknownEnergy:
@conor10 As far as I remember this methods were private but if Android branch now caught up with master were it is public then this issue can be closed yes.

snazha-blkio:
Closing this on the basis that it will be addressed in issue #769 and PR #618.

PR #701, opened 2018-09-10
WIP support for multiple services.

sebaraba:
### What does this PR do?
*required*
- [x] Change web3j to work with a list of services;
- [x] If one node does not work, web3j will automatically try to connect to the next one in the list so on until one node is available to connect to;

### To do: 
- [ ] Update tests;
- [ ] Request class has one method `next()` set as public. We need to change this to be at least protected, but it is used in the `WebSocketService` class, which is in a different package.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/Web3j.java
+++ b/core/src/main/java/org/web3j/protocol/Web3j.java
@@ -2,6 +2,7 @@ package org.web3j.protocol;
 
 import java.util.concurrent.ScheduledExecutorService;
 
+import org.web3j.abi.datatypes.Array;
 import org.web3j.protocol.core.Ethereum;
 import org.web3j.protocol.core.JsonRpc2_0Web3j;
 import org.web3j.protocol.rx.Web3jRx;
@@ -14,11 +15,11 @@ public interface Web3j extends Ethereum, Web3jRx {
     /**
      * Construct a new Web3j instance.
      *
-     * @param web3jService web3j service instance - i.e. HTTP or IPC
+     * @param web3jServices web3j service instances - i.e. HTTP or IPC
      * @return new Web3j instance
      */
-    static Web3j build(Web3jService web3jService) {
-        return new JsonRpc2_0Web3j(web3jService);
+    static Web3j build(Web3jService... web3jServices ) {
+        return new JsonRpc2_0Web3j(web3jServices);
     }
 
     /**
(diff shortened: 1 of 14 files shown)


conor10:
Hi @sebaraba, this change touches a lot of the web3j API. Can the change not simply be done via the [HTTPService](https://github.com/web3j/web3j/blob/master/core/src/main/java/org/web3j/protocol/http/HttpService.java)?  I'd have thought OkHttp will provide support for multiple connections. It doesn't explicitly say it [here](https://github.com/square/okhttp/wiki/Connections#connections), but it seems sensible.

I'm not keen on the Web3jService abstraction changing as that will be a major change.

core/src/main/java/org/web3j/protocol/Web3j.java
```diff
@@ -14,11 +15,11 @@
     /**
      * Construct a new Web3j instance.
      *
-     * @param web3jService web3j service instance - i.e. HTTP or IPC
+     * @param web3jServices web3j service instances - i.e. HTTP or IPC
      * @return new Web3j instance
      */
-    static Web3j build(Web3jService web3jService) {
-        return new JsonRpc2_0Web3j(web3jService);
+    static Web3j build(Web3jService... web3jServices) {
```
iikirilov: Instead of allowing a list of Web3jServices you can extend the Http implementations to accept multiple urls.

core/src/main/java/org/web3j/protocol/core/Request.java
```diff
@@ -68,14 +73,39 @@ public void setId(long id) {
     }
 
     public T send() throws IOException {
-        return web3jService.send(this, responseType);
+        try {
```
iikirilov: Move this logic to HttpService

core/src/test/java/org/web3j/protocol/http/HttpServiceTest.java
```diff
@@ -15,20 +15,37 @@
...
 import org.web3j.protocol.core.Request;
 import org.web3j.protocol.core.methods.response.EthBlockNumber;
 import org.web3j.protocol.core.methods.response.EthSubscribe;
 import org.web3j.protocol.exceptions.ClientConnectionException;
 import org.web3j.protocol.websocket.events.NewHeadsNotification;
 
+import static java.util.Collections.emptyList;
 import static org.junit.Assert.assertTrue;
 
 public class HttpServiceTest {
```
iikirilov: Have a look at the tests again after changing your approach

snazha-blkio:
@sebaraba are you able to create a new PR so this could be closed off, given Conor's latest comment?

iikirilov:
moved to #782

PR #708, opened 2018-09-13
Ignore logs without topics when parsing events

loriopatrick:
### What does this PR do?
Skips over anonymous logs when parsing events. 

Still need to add a test.

### Why is it needed?
Currently when a transaction has an anonymous log (log0) amongst event logs, an IndexOutOfBoundsException is thrown when filtering for a specific event.


snazha-blkio:
@loriopatrick , would you be able to rebase from release/4.0 and provide a test as per @conor10 's comment?

Thanks.

PR #711, opened 2018-09-17
Code generator correction

flefevre:
### What does this PR do?

- remove deprecated constructor method in generator
- change load method in generator

both of these changes introduce the ContractGasProvider in the generated
code

- have a look to SolidityFunctionWrapper
- test the generated code in the linked project web3j-maven-plugin

 ### Why is it needed?

 - the initial generated code is not compatible with FastRawTransactionManager

### DIFF
--- a/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
+++ b/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
@@ -55,6 +55,7 @@ import org.web3j.protocol.core.methods.response.Log;
 import org.web3j.protocol.core.methods.response.TransactionReceipt;
 import org.web3j.tx.Contract;
 import org.web3j.tx.TransactionManager;
+import org.web3j.tx.gas.ContractGasProvider;
 import org.web3j.utils.Collection;
 import org.web3j.utils.Strings;
 import org.web3j.utils.Version;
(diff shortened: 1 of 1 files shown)


codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -122,16 +124,13 @@ void generateJavaFiles(
 
         TypeSpec.Builder classBuilder = createClassBuilder(className, bin);
 
-        classBuilder.addMethod(buildConstructor(Credentials.class, CREDENTIALS));
         classBuilder.addMethod(buildConstructor(TransactionManager.class,
                 TRANSACTION_MANAGER));
         classBuilder.addFields(buildFuncNameConstants(abi));
         classBuilder.addMethods(
                 buildFunctionDefinitions(className, classBuilder, abi));
-        classBuilder.addMethod(buildLoad(className, Credentials.class, CREDENTIALS));
```
conor10: We should be able to provide a TransactionManager or Credentials in a generated class - this looks like it removes the credentials.

snazha-blkio: Let me roll back this commit and get the 3.6 out, given that more work needs to be done on this.

iikirilov:
cherry-picked this commit and applied comment in #725

PR #713, opened 2018-09-18
Fix invalid index error on node disconnection.

sebaraba:
### What does this PR do?
When the latest `BlockNumber` was 0 `Invalid index error` would be fired because we try to `CatchUpToLatestAndSubscribeToNewBlocksObservable` starting from `index = -1`. 
By deleting the `.subtract(BigInt.ONE)` the block number will never be 0 and nodes would restart correctly.

*required*

### Why is it needed?
*required*


core/src/main/java/org/web3j/protocol/core/DefaultBlockParameter.java
```diff
@@ -10,6 +10,8 @@
  */
 public interface DefaultBlockParameter {
     static DefaultBlockParameter valueOf(BigInteger blockNumber) {
+        if (BigInteger.ZERO.compareTo(blockNumber) >= 0)
+            blockNumber = BigInteger.ZERO;
```
iikirilov: Here you should throw an appropriate exception rather than modifying the input.

integration-tests/src/test/java/org/web3j/protocol/core/ObservableIT.java
```diff
@@ -21,8 +21,8 @@
  */
 public class ObservableIT {
 
-    private static final int EVENT_COUNT = 5;
-    private static final int TIMEOUT_MINUTES = 5;
+    private static final int EVENT_COUNT = 1;
```
iikirilov: You should leave these values as they were.
You can add the extra static variable for edge cases like the one you are testing. i.e.
`private static final int SINGLE_EVENT_COUNT = 1`
Then add specific tests for those edge cases.

integration-tests/src/test/java/org/web3j/protocol/core/ObservableIT.java
```diff
@@ -21,8 +21,8 @@
...
+    private static final int TIMEOUT_MINUTES = 1;
```
iikirilov: Timeout can stay the same

integration-tests/src/test/java/org/web3j/protocol/core/ObservableIT.java
```diff
@@ -64,7 +64,7 @@ public void testCatchUpToLatestAndSubscribeToNewBlocksObservable() throws Except
                 .send();
         BigInteger latestBlockNumber = ethBlock.getBlock().getNumber();
         run(web3j.catchUpToLatestAndSubscribeToNewBlocksObservable(
-                new DefaultBlockParameterNumber(latestBlockNumber.subtract(BigInteger.ONE)),
+                new DefaultBlockParameterNumber(latestBlockNumber),
```
iikirilov: Leave this test as it was and add another for your edge case.

core/src/main/java/org/web3j/protocol/core/DefaultBlockParameter.java
conor10: I'm not sure that this did fix the issue - @xaviarias did you have any feedback on it?

xaviarias:
This could be merged and closed after implementing the requested changes.

iikirilov:
There are also checkstyle violations.

sebaraba:
@iikirilov  Done with this.

PR #714, opened 2018-09-21
fixes issue with IndexOutOfBoundsException when trying to populate wo…

ShadowCatSteph:
…rd list

### What does this PR do?
Returns an empty list

Line 259

### Why is it needed?
Fixes an issue with IndexOutOfBoundsException when trying to populate word list for BIP39 wallet.  It seems that the problem is that the file path with the mnemonic words is incorrectly parsed.


iikirilov:
Agreed. This is not a problem with the code, rather the file does not exist or is missing.

We should definitely not just return an empty list!

PR #725, opened 2018-10-03
Remove deprecated solidity code from example contracts and fix code generation

iikirilov:
### What does this PR do?
Updates all of the sample contracts such that they are following the latest Solidity syntax. Removed all of the deprecated keywords and syntax. Also recompiled those contracts and regenerated the smart-contract wrappers.

Also provides a fix for code generation

The smart contracts are in the codegen module.
The regenerated smart contract wrappers are in the integration-tests module in the generated package.

See #711 for code generation

### Why is it needed?
Because the example contracts were mostly deprecated and would throw a number of warnings while compiling. Also, this library will be used in a workshop, so it is good to have the example up to date and fully functional generated code.

**N.B.** The web3j 3.5.0 binary does not generate the smart contract wrappers correctly, so the integration test will most probably fail. I am waiting for an updated binary which should include #711


iikirilov:
includes #711 with applied comments

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -310,10 +312,9 @@ private static MethodSpec buildConstructor(Class authType, String authName) {
                 .addParameter(String.class, CONTRACT_ADDRESS)
                 .addParameter(Web3j.class, WEB3J)
                 .addParameter(authType, authName)
-                .addParameter(BigInteger.class, GAS_PRICE)
```
conor10: We don't want to force people to use GasProviders in generated wrappers until we release version 4.0. I.e. generatd code should support both gas providers and the original wrappers (which are correctly marked as deprecated).

iikirilov: ofcourse

Issue #720, opened 2018-09-26
Release web3j-3.6.x


snazha-blkio:
We need to release 3.6.x with some important fixes.

snazha-blkio:
Released on 7 October 2018.

PR #726, opened 2018-10-03
#720: improve the build script to publish snapshots

snazha-blkio:
### What does this PR do?
Updates the gradle build scripts to:
1. externalise the build version into a separate properties file
2. support publishing of snapshots
3. not run publish tasks if no credentials are provided

build.gradle

### Why is it needed?
To support publishing of snapshots

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,8 +19,6 @@ hs_err_pid*
 */build/
 */out/
 
-gradle.properties
-
 # Ignore Gradle GUI config
 gradle-app.setting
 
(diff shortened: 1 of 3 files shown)


iikirilov:
Might be useful to provide a gradle.properties.example with stubbed fields

snazha-blkio:
Botched commit, file is now added in.
https://github.com/web3j/web3j/blob/feature/720-improve-build-script/gradle.properties

Thanks for picking up on it.

PR #731, opened 2018-10-05
fixed issue when we have error before listener assigned

dmikel:
### What does this PR do?
check listener before call

-
### Why is it needed?
to avoid NullPointerException

### DIFF
--- a/core/src/main/java/org/web3j/protocol/websocket/WebSocketClient.java
+++ b/core/src/main/java/org/web3j/protocol/websocket/WebSocketClient.java
@@ -44,13 +44,17 @@ public class WebSocketClient extends org.java_websocket.client.WebSocketClient {
     public void onClose(int code, String reason, boolean remote) {
         log.info("Closed WebSocket connection to {}, because of reason: '{}'."
                 + "Conection closed remotely: {}", uri, reason, remote);
-        listener.onClose();
+        if (null != listener) {
+            listener.onClose();
+        }
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/web3j/protocol/websocket/WebSocketClient.java
```diff
@@ -44,13 +44,17 @@ public void onMessage(String s) {
     public void onClose(int code, String reason, boolean remote) {
         log.info("Closed WebSocket connection to {}, because of reason: '{}'."
                 + "Conection closed remotely: {}", uri, reason, remote);
-        listener.onClose();
+        if (null != listener) {
```
iikirilov: code duplication

core/src/main/java/org/web3j/protocol/websocket/WebSocketClient.java
```diff
@@ -44,13 +44,17 @@ public void onMessage(String s) {
...
+            listener.onClose();
+        }
     }
 
     @Override
     public void onError(Exception e) {
         log.error(String.format("WebSocket connection to {} failed with error", uri), e);
-        listener.onError(e);
+        if (null != listener) {
```
iikirilov: code duplication

snazha-blkio:
Hi @dmikel,

I agree with @iikirilov , having null checks will just grow as we add more methods.
Could you please look at making the 'listener' a constructor argument, throwing an exception if one is not provided?

Thanks.

PR #746, opened 2018-10-19
fix WebSocketClient log error format bug

wenshenjun:
### What does this PR do?
Just the output log format is wrong.

core/src/main/java/org/web3j/protocol/websocket/WebSocketClient.java

### Why is it needed?
Good for debug.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/websocket/WebSocketClient.java
+++ b/core/src/main/java/org/web3j/protocol/websocket/WebSocketClient.java
@@ -36,7 +36,7 @@ public class WebSocketClient extends org.java_websocket.client.WebSocketClient {
             log.debug("Received message {} from server {}", s, uri);
             listener.onMessage(s);
         } catch (Exception e) {
-            log.error("Failed to process message '{}' from server {}", s, uri);
+            log.error(String.format("Failed to process message '%s' from server %s", s, uri),e);
         }
     }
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/web3j/protocol/websocket/WebSocketClient.java
```diff
@@ -36,7 +36,7 @@ public void onMessage(String s) {
             log.debug("Received message {} from server {}", s, uri);
             listener.onMessage(s);
         } catch (Exception e) {
-            log.error("Failed to process message '{}' from server {}", s, uri);
+            log.error(String.format("Failed to process message '%s' from server %s", s, uri),e);
```
snazha-blkio: @wenshenjun there is no need to use String.format, using the {} placeholders is equivalent.
Can you please retain those and only add in the exception? Moreover ...
```log.error("Failed to process message '{}' from server {}", s, uri, e);```

File: core/src/main/java/org/web3j/protocol/websocket/WebSocketClient.java [hunk 2]
```diff
@@ -49,7 +49,7 @@ public void onClose(int code, String reason, boolean remote) {
 
     @Override
     public void onError(Exception e) {
-        log.error(String.format("WebSocket connection to {} failed with error", uri), e);
+        log.error(String.format("WebSocket connection to %s failed with error", uri), e);
```
snazha-blkio: No need for String.format() here.

snazha-blkio:
This has been addressed in https://github.com/web3j/web3j/pull/785

PR #761, opened 2018-10-27
Prefix message signing according to eth_sign. (#575)

snazha-blkio:
** This is a copy of PR#650 to address some feedback prior to merging **
Original PR: https://github.com/web3j/web3j/pull/650

--
As we discussed before, I have been unable to run the ./gradlew integration tests. But opening a PR so it is easier to get this discussion going. :)

I've updated the test cases by using the key pair to sign a message using ethers. I will be adding more test cases to ethers soon, generated by parity and geth.

I am also concerned that any dapp that has previously signed a message with web3j will no longer get the correct key back when calling signedMessageToKey, so this is at best, something that should probably be a major version change. The other option is to add a signedPrefixedMessageToKey and a signPrefixedMessage and deprecate the non-prefixed version.


snazha-blkio:
I will approve and merge this given the approval on the original PR: https://github.com/web3j/web3j/pull/650

PR #782, opened 2018-11-12
[WIP] - Client side load balancing

sebaraba:
### Done:
- [x] Multiple http node support;
- [x] Old tests are running;
- [x] New test for multiple nodes;

### Mentions with respect to the comments: 
- [x] I had to pull in `release/4.0` branch in order to overcome the `CLEARTEXT error` and be able to run the tests.

- [x]  I noticed there is a rule which enforced me to keep the url/urls parameter on the last position in the constructor:
```java
public HttpService(OkHttpClient httpClient, boolean includeRawResponses, String... urls)
```
Please let me know if there is anyway to overcome this in order not to change the position of the parameter.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/http/HttpService.java
+++ b/core/src/main/java/org/web3j/protocol/http/HttpService.java
@@ -38,9 +38,12 @@ public class HttpService extends Service {
             CipherSuite.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,
             CipherSuite.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
 
-            // Note that the following cipher suites are all on HTTP/2's bad cipher suites list. We'll
-            // continue to include them until better suites are commonly available. For example, none
-            // of the better cipher suites listed above shipped with Android 4.4 or Java 7.
+            // Note that the following cipher suites are all on
+            // HTTP/2's bad cipher suites list. We'll
+            // continue to include them until better suites are
+            // commonly available. For example, none
+            // of the better cipher suites listed above shipped
+            // with Android 4.4 or Java 7.
             CipherSuite.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,
             CipherSuite.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
             CipherSuite.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,
--- a/core/src/main/java/org/web3j/protocol/http/HttpService.java
+++ b/core/src/main/java/org/web3j/protocol/http/HttpService.java
@@ -31,31 +31,31 @@ public class HttpService extends Service {
      * Copied from {@link ConnectionSpec#APPROVED_CIPHER_SUITES}.
      */
     private static final CipherSuite[] INFURA_CIPHER_SUITES = new CipherSuite[] {
-        CipherSuite.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
-        CipherSuite.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
-        CipherSuite.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
-        CipherSuite.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
-        CipherSuite.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,
-        CipherSuite.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
+            CipherSuite.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+            CipherSuite.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/web3j/protocol/http/HttpService.java
```diff
@@ -154,6 +169,7 @@ protected InputStream performIO(String request) throws IOException {
         }
     }
 
+
```
iikirilov: unnecessary whitespace

core/src/main/java/org/web3j/protocol/http/HttpService.java
```diff
@@ -70,41 +70,43 @@
...
+        this(httpClient, includeRawResponses, DEFAULT_URL);
     }
 
-    private HttpService(String url, OkHttpClient httpClient) {
-        this(url, httpClient, false);
+    private HttpService(OkHttpClient httpClient, String url) {
+        this(httpClient, false, url);
     }
 
     public HttpService(String url) {
```
iikirilov: You will need a constructor `HttpService(String... urls)` too

core/src/main/java/org/web3j/protocol/http/HttpService.java
```diff
@@ -70,41 +70,43 @@
...
     }
 
     public HttpService(OkHttpClient httpClient, boolean includeRawResponses) {
-        this(DEFAULT_URL, httpClient, includeRawResponses);
+        this(httpClient, includeRawResponses, DEFAULT_URL);
     }
 
-    private HttpService(String url, OkHttpClient httpClient) {
-        this(url, httpClient, false);
+    private HttpService(OkHttpClient httpClient, String url) {
```
iikirilov: You will need `HttpService(OkHttpClient httpClient, String... urls)`

core/src/main/java/org/web3j/protocol/http/HttpService.java
```diff
@@ -70,41 +70,43 @@
...
     }
 
     public HttpService(String url) {
-        this(url, createOkHttpClient());
+        this(createOkHttpClient(), url);
     }
 
-    public HttpService(String url, boolean includeRawResponse) {
-        this(url, createOkHttpClient(), includeRawResponse);
+    public HttpService(boolean includeRawResponse, String url) {
```
iikirilov: You will need `HttpService(boolean includeRawResponse, String... urls)`

core/src/main/java/org/web3j/protocol/http/HttpService.java
```diff
@@ -70,41 +70,43 @@
...
+
+    private int counter = 0;
 
-    private final boolean includeRawResponse;
+    private boolean includeRawResponse;
 
     private HashMap<String, String> headers = new HashMap<>();
 
-    public HttpService(String url, OkHttpClient httpClient, boolean includeRawResponses) {
+    public HttpService(OkHttpClient httpClient, boolean includeRawResponses, String... urls) {
```
iikirilov: You will need `HttpService(OkHttpClient httpClient, boolean includeRawResponses, String url)` too

core/src/main/java/org/web3j/protocol/http/HttpService.java
```diff
@@ -199,6 +215,14 @@ public void addHeaders(Map<String, String> headersToAdd) {
         return headers;
     }
 
+    public String next() {
```
iikirilov: call it `nextUrl()` - I think it should be a private method also.

core/src/main/java/org/web3j/protocol/http/HttpService.java
```diff
@@ -131,13 +133,26 @@ protected InputStream performIO(String request) throws IOException {
...
+        try {
+            httpRequest = new okhttp3.Request.Builder()
+                    .url(url)
+                    .headers(headers)
+                    .post(requestBody)
+                    .build();
+        } catch (IllegalStateException e) {
+            url = next();
+            if (url != null) {
+                httpRequest = new okhttp3.Request.Builder()
```
iikirilov: encapsulate in `private okhttp3.Request buildHttpRequest(...)` instead of copy pasta

core/src/main/java/org/web3j/protocol/http/HttpService.java
```diff
@@ -131,13 +133,26 @@ protected InputStream performIO(String request) throws IOException {
...
+                        .url(url)
+                        .headers(headers)
+                        .post(requestBody)
+                        .build();
+            } else {
```
iikirilov: it is good to do `log.error("message", e)` before rethrowing.

iikirilov:
Also, tests required.

core/src/main/java/org/web3j/protocol/http/HttpService.java
sebaraba: You mean moving the check implementation inside the `okhttp3.Rquest` class?

iikirilov: No. I mean write this method in HttpService class so you don't duplicate code.

iikirilov:
You have checkstyle violations

iikirilov:
> * [ ]  Also `includeRawResponses` and `url` were final before my changes. Is there a way to keep it final after the changes ?

You should be able to keep them as final.

fcorneli:
This feature will also mess with the nonce cache... and hence fast transactions.

iikirilov:
This feature has been dropped

PR #783, opened 2018-11-12
731: Fix NPE in WebSocketClient if an error occurs during connection

snazha-blkio:
### What does this PR do?
Addresses the case where WebSocketClient is used without a listener.

WebSocketClient.

### Why is it needed?
To safeguard against NPEs, and to address an issue raised via another PR: https://github.com/web3j/web3j/pull/731


snazha-blkio:
Although WebSocketClientTest.java already tests the functionality I added a dedicated test for the case where the listener is not set.

PR #784, opened 2018-11-13
fixed WebSocketClient onError log message

zacscoding:
### What does this PR do?
fixed error message format 

WebSocketClient

### Why is it needed?
Printed wrong message


snazha-blkio:
@zacscoding, well-spotted. Can you please change the message to:
`log.error("WebSocket connection to {} failed with error", uri, e);`

PR #786, opened 2018-11-13
Bugfix/784 websocketclient error message

snazha-blkio:
### What does this PR do?
Fixes incorrect log statement usage.

WebSocketClient.java

### Why is it needed?
Incorrect use of log statements.


snazha-blkio:
Addressed in https://github.com/web3j/web3j/pull/783.

PR #787, opened 2018-11-14
adds WebSocketListener at WebSocketService

zacscoding:
### What does this PR do?
added WebSocketListener at WebSocketService

WebSocketService, WebSocketServiceTest

### Why is it needed?
I want to handle websocket events such as onClose event.
I think that it will be better if WebSocketClient have multiple listeners and WebSocketSerivce`s listener
is added.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
+++ b/core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
@@ -14,6 +14,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
@@ -90,9 +91,14 @@ public class WebSocketService implements Web3jService {
      * @throws ConnectException thrown if failed to connect to the server via WebSocket protocol
      */
     public void connect() throws ConnectException {
+        connect(null, null, null);
+    }
+
+    public void connect(Consumer<String> onMessage, Consumer<Throwable> onError,
+            Consumer<Void> onClose) throws ConnectException {
         try {
             connectToWebSocket();
-            setWebSocketListener();
+            setWebSocketListener(onMessage, onError, onClose);
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             log.warn("Interrupted while connecting via WebSocket protocol");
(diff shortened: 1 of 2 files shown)


zacscoding:
how about this ?
```
    public void connect() throws ConnectException {
        connect(null, null, null);
    }

    public void connect(Consumer<String> onMessage, Consumer<Throwable> onError, Runnable onClose) throws ConnectException {
        try {
            connectToWebSocket();
            setWebSocketListener(onMessage, onError, onClose);
        } catch (InterruptedException e) {
            Thread.currentThread().interrupt();
            log.warn("Interrupted while connecting via WebSocket protocol");
        }
    }

    private void connectToWebSocket() throws InterruptedException, ConnectException {
        boolean connected = webSocketClient.connectBlocking();
        if (!connected) {
            throw new ConnectException("Failed to connect to WebSocket");
        }
    }

    private void setWebSocketListener(Consumer<String> onMessage, Consumer<Throwable> onError, Runnable onClose) {
        webSocketClient.setListener(new WebSocketListener() {
            @Override
            public void onMessage(String message) throws IOException {
                onWebSocketMessage(message);
                if (onMessage != null) {
                    onMessage.accept(message);
                }
            }

            @Override
                log.error("Received error from a WebSocket connection", e);
                if (onError != null) {
                    onError.accept(e);
                }
            }

            @Override
            public void onClose() {
                onWebSocketClose();
                if (onClose != null) {
                    onClose.run();
                }
            }
        });
    }
```

I don`t know how to maintain multiple WebSocketClient in WebSocketService TT

core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
```diff
@@ -106,21 +111,30 @@ private void connectToWebSocket() throws InterruptedException, ConnectException
         }
     }
 
-    private void setWebSocketListener() {
+    private void setWebSocketListener(Consumer<String> onMessage, Consumer<Throwable> onError, Runnable onClose) {
         webSocketClient.setListener(new WebSocketListener() {
```
snazha-blkio: @zacscoding , use of `Runnable` can be confusing as it would imply the implementation is invoked by another thread. Could you instead use `Consumer<Void>` please?
@iikirilov , thoughts?

zacscoding: If use Consumer<Void>, then have to called like onClose.accept(null);
is ok ?

iikirilov: yes

zacscoding:
appreciate ur kindness 👍

snazha-blkio:
> @snazha-blkio I updated source with checkstyle. is any problem?

Could you please add some unit tests for your changes? We can then look at getting this into 5.0

zacscoding:
@snazha-blkio  added some testing code !

snazha-blkio:
Thanks @zacscoding , this will go out in 5.0

olimsaidov:
@snazha-blkio is there any reason why this pull request cannot be released on 4.* versions? There is no breaking changes in public interface. 

We really need this feature. Thank you.

iikirilov:
@olimsaidov yes we can do this

PR #795, opened 2018-11-19
missing project dependency "crypto"

apetersson:
SolidityFunctionWrapper uses org.web3j.crypto.Credentials. Dependency to crpyto is needed (alternatively moving to core or utils)

### What does this PR do?
adds a dependency to the crypto subproject

build.grade

### Why is it needed?
to run the TruffleJsonFunctionWrapperGenerator in standalone mode

### DIFF
--- a/codegen/build.gradle
+++ b/codegen/build.gradle
@@ -2,8 +2,7 @@
 description 'web3j project code generators'
 
 dependencies {
-    compile project(':core'),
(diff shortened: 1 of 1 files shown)


apetersson:
i revisited the issue on my build, and was not able to reproduce the error that lead to this PR. closing, not an issue anymore.

PR #851, opened 2019-01-18
Bringing the android branch inline with the main branch - part 1

snazha-blkio:
### What does this PR do?
Bringing the android branch inline with the main branch


snazha-blkio:
Not needed

Issue #858, opened 2019-01-21
Retrieve revert reason for EthCall


MDhondt:
Since Solidity 0.4.22 the `require()` and `revert()` operation optionally take a revert reason as argument: [Revert with reason](https://github.com/ethereum/solidity/pull/3364).

This is accessible for `eth_call`s in the `result` field.

Have a method `public String getRevertReason()` on `org.web3j.protocol.core.methods.response.EthCall` that retrieves that revert reason.

PR #859, opened 2019-01-21
Retrieve revert reason for EthCall

MDhondt:
### What does this PR do?
Retrieves the revert reason from an `eth_call` if it is present. This is issue #858.

Check Solidity changes adding the `require()` and `revert()` reason at https://github.com/ethereum/solidity/pull/3364/files.

Use following contract to trigger a reverting call:

```solidity
pragma solidity ^0.4.22;

contract Web3j858 {
    function willRevert() public pure {
        require(1 > 2, "Solidity uses state-reverting exceptions to handle errors. The require function should be used to ensure valid conditions, such as inputs, or contract state variables are met, or to validate return values from calls to external contracts.");
    }
}
```

### Why is it needed?
Because the information is already there, web3j just doesn't interpret it.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/core/methods/response/EthCall.java
+++ b/core/src/main/java/org/web3j/protocol/core/methods/response/EthCall.java
@@ -1,14 +1,14 @@
 package org.web3j.protocol.core.methods.response;
 
+import java.util.Collections;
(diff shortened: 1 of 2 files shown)


iikirilov:
LGTM

Issue #874, opened 2019-02-20
Contract.isValid() shouldn't compare "Metadata Hash"


p-s-dev:
The "Metadata Hash" is some extra non-executable bytecode near the end of the contract bytecode.  This bytecode represent a hash of info such as the filename of the solidity file that was compiled.  Since it has no impact on the execution of the contract, it should be ignored curing the isValid() check.

here is an example where another code project excludes the metadata hash when checking contract validity.

https://github.com/ConsenSys/bytecode-verifier/blob/master/src/verifier.js#L87

fcorneli:
This would indeed be interesting. A recompilation of the same solidity contract should not mark previous contracts as invalid if the actual bytecode is the same.

antonydenyer:
The only methods that use it are `isValid` and `deploy` so it shouldn't cause any problems. But it would be better if there was a way to represent instances of contracts. e.g I have multiple deployments of the same contract.

PR #878, opened 2019-02-23
skip metadata for validating contract code

fcorneli:
This PR makes sure that we do not include code metadata when validating for contract validity.
See also: https://github.com/web3j/web3j/issues/874


conor10:
Please can we include a unit test?

snazha-blkio:
@fcorneli , thanks for this. As per Conor's comment could you please include a test so we can get this released?

fcorneli:
@snazha-blkio added a unit test.

PR #885, opened 2019-03-01
include call data when getting gasPrice & gasLimit in ContractGasProvider

p-s-dev:
### What does this PR do?
Update ContractGasProvider interface to allow passing call data.  

ContractGasProvider interface and Contract.executeTransaction() protected method.

### Why is it needed?
If ContractGasProvider only has function name, it is difficult to make informed decisions about gas price.  Alternatively Contract.executeTransaction() could be changed from protected to public, or the ContractGasProvider could be moved to ManagedTransaction .


p-s-dev:
maybe this is an anti-pattern.

PR #903, opened 2019-03-26
Change Chain ID type from byte to long

bartwell:
### What does this PR do?
Allow using long Chain ID.

public static byte[] signMessage(RawTransaction rawTransaction, long chainId, Credentials credentials)

### Why is it needed?
It needed because some network have long ID. For example PIRL Chain ID is 3125659152. Byte can't accept this value.


iikirilov:
Thanks for the input here but I feel #905 is more in line with what I believe needs to be changed.

PR #904, opened 2019-03-26
Generalize Hosted Providers Concept and Add Nodesmith

sdesmond46:
### What does this PR do?
This PR does some work to to generalize the concept of hosted web3 providers instead of just having Infura as a named provider. This will be useful as more competitors to Infura come online.

* Renames `infura` project to `hosted-providers`
* Adds [Nodesmith](https://nodesmith.io) as a provider.
* Adds an optional method to `HttpService` to read the headers which are sent back from the server allowing for additional processing
* Updates some of the documentation.

With the `NodesmithHttpService` class and the renaming of the `infura` subproject to `hosted-providers`

### Why is it needed?
Much has been written about the need for Ethereum to diversify away from heavily depending on Infura. This change makes it so other hosted providers can be on level footing with Infura.


sdesmond46:
> LGTM. You are legend!

Thanks for taking a look!

PR #905, opened 2019-03-27
TransactionEncoder to accept chainId as a long

rain-on:
### What does this PR do?
Allows users of Web3j to set a chainId greater than 255.

This change includes 2 commits:
1. A piece of work conducted in e-Contract whereby Signature is updated to use a byte array for "V".
(https://github.com/e-Contract/web3j/commit/6b704849b181480b69fbfb20602c4c921d3a4e4a)
2. TransactionEncoder to accept chainId as a long (and update chain constants from byte to long).

### Why is it needed?
Developers working with private chains (which theoretically have a chain ID greater than 1337) are unable to use web3j to sign transactions bound for their web3 provider.

### DIFF
--- a/core/src/test/java/org/web3j/ens/EnsResolverTest.java
+++ b/core/src/test/java/org/web3j/ens/EnsResolverTest.java
@@ -51,7 +51,7 @@ public class EnsResolverTest {
         configureLatestBlock(System.currentTimeMillis() / 1000);  // block timestamp is in seconds
 
         NetVersion netVersion = new NetVersion();
-        netVersion.setResult(Byte.toString(ChainId.MAINNET));
+        netVersion.setResult(Long.toString(ChainId.MAINNET));
 
         String resolverAddress =
                 "0x0000000000000000000000004c641fb9bad9b60ef180c31f56051ce826d21a9a";
@@ -81,7 +81,7 @@ public class EnsResolverTest {
         configureLatestBlock(System.currentTimeMillis() / 1000);  // block timestamp is in seconds
 
         NetVersion netVersion = new NetVersion();
-        netVersion.setResult(Byte.toString(ChainId.MAINNET));
+        netVersion.setResult(Long.toString(ChainId.MAINNET));
 
         String resolverAddress =
                 "0x0000000000000000000000004c641fb9bad9b60ef180c31f56051ce826d21a9a";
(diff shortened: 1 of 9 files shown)


iikirilov:
I have a concern that changing the type of `v` will break things downstream mainly in web3j-quorum. It may be the case that this PR is included in web3j 5.0

iikirilov:
Please can we merge this for 4.3

iikirilov:
I would like to get this merged ASAP - to that we will need to provide deprecated methods for backwards compatibility. @rain-on is this something you can do? - would be awesome

crypto/src/test/java/org/web3j/crypto/TransactionDecoderTest.java
```diff
@@ -75,7 +75,7 @@ public void testDecodingSignedChainId() throws Exception {
         BigInteger gasLimit = BigInteger.TEN;
         String to = "0x0add5355";
         BigInteger value = BigInteger.valueOf(Long.MAX_VALUE);
-        Integer chainId = 1;
+        Integer chainId = 46;
```
iikirilov: Why are you testing with 46 here? This is still below 256.

crypto/src/main/java/org/web3j/crypto/TransactionDecoder.java
```diff
@@ -20,7 +20,7 @@ public static RawTransaction decode(String hexTransaction) {
         BigInteger value = ((RlpString) values.getValues().get(4)).asPositiveBigInteger();
         String data = ((RlpString) values.getValues().get(5)).asString();
         if (values.getValues().size() > 6) {
-            byte v = ((RlpString) values.getValues().get(6)).getBytes()[0];
+            byte[] v = ((RlpString) values.getValues().get(6)).getBytes();
```
iikirilov: Can we test this too please - that we can decode a transaction that was encoded with a chainId > 256

core/src/test/java/org/web3j/ens/EnsResolverTest.java
```diff
@@ -51,7 +51,7 @@ public void testResolve() throws Exception {
         configureLatestBlock(System.currentTimeMillis() / 1000);  // block timestamp is in seconds
 
         NetVersion netVersion = new NetVersion();
-        netVersion.setResult(Byte.toString(ChainId.MAINNET));
+        netVersion.setResult(Long.toString(ChainId.MAINNET));
```
iikirilov: Can we make this backwards compatible? It would require make a new class `ChainIdLong` and deprecating the old one

iikirilov:
replaced by #913 thank you @rain-on for initial commit

rain-on:
Sorry for losing track of this - day-to-day got in the way badly :(

Issue #879, opened 2019-02-25
Support contracts with arrays containing more than 32 elements


snazha-blkio:
`StaticArray.checkValid` contains a max size check of 32 elements; we need to support arbitrary length arrays.

At this stage we need to support at least arrays with up to 584 elements.

```
Caused by: java.lang.UnsupportedOperationException: Invalid type provided org.web3j.abi.datatypes.generated.StaticArray32
    at org.web3j.abi.Utils.getParameterizedTypeName(Utils.java:74)
    at org.web3j.abi.Utils.getTypeName(Utils.java:33)
    at org.web3j.abi.EventEncoder.lambda$buildMethodSignature$0(EventEncoder.java:37)
    at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
    at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1382)
    at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
    at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
    at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
    at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
    at org.web3j.abi.EventEncoder.buildMethodSignature(EventEncoder.java:38)
    at org.web3j.abi.EventEncoder.encode(EventEncoder.java:23)
    at org.web3j.tx.Contract.staticExtractEventParameters(Contract.java:504)
    at org.web3j.tx.Contract.extractEventParametersWithLog(Contract.java:535)
    at tech.pegasys.artemis.pow.contract.ValidatorRegistrationContract.access$000(ValidatorRegistrationContract.java:39)
    at tech.pegasys.artemis.pow.contract.ValidatorRegistrationContract$7.apply(ValidatorRegistrationContract.java:97)
    at tech.pegasys.artemis.pow.contract.ValidatorRegistrationContract$7.apply(ValidatorRegistrationContract.java:94)
    at io.reactivex.internal.operators.flowable.FlowableMap$MapSubscriber.onNext(FlowableMap.java:63)
    ... 12 more
```

PR #910, opened 2019-03-31
#879: Added support for up to 1024 elements in (solidity) arrays.

snazha-blkio:
### What does this PR do?
Adds support for larger (> 32) arrays in the Java contract wrappers

All changes.

### Why is it needed?
There are some changes (eth2.0?) where arrays containing up to 584 elements are required.


snazha-blkio:
We may need to generate classes in increments of 8, @iikirilov will confirm if this is needed.

iikirilov:
I never got any more information for this. I think it is ok to proceed like until we conjure up a better solution

PR #911, opened 2019-04-01
Implement EIP712

### LINKED ISSUE #898: EIP-712 Signing structured data
We need support for [eip712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md)

A java implementation of eip-712 v1 exists [here](https://github.com/gjeanmart/eth_signTypedData). This implementation is a good starting point, however, is severely updated.

Where this will fit into web3j needs to be clarified.

stringray55:
### What does this PR do?
Fixes #898

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
+++ b/crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
@@ -1,0 +1,271 @@
+package org.web3j.crypto;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.lang.reflect.Constructor;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Vector;
+import java.util.regex.Pattern;
+import javax.xml.bind.ValidationException;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import org.web3j.abi.TypeEncoder;
+import org.web3j.abi.datatypes.Type;
+import org.web3j.abi.datatypes.generated.AbiTypes;
+import org.web3j.utils.Numeric;
+
+import static org.web3j.crypto.Hash.sha3;
+import static org.web3j.crypto.Hash.sha3String;
+
+public class StructuredDataEncoder {
+    public static StructuredData.EIP712Message jsonMessageObject;
+
+    public static HashSet<String> getDependencies(String primaryType) {
+        // Find all the dependencies of a type
+        HashMap<String, Vector<StructuredData.Entry>> types = jsonMessageObject.types;
+        HashSet<String> deps = new HashSet<>();
+
+        if (!types.containsKey(primaryType)) {
+            return deps;
+        }
+
+        List<String> remainingTypes = new ArrayList<>();
+        remainingTypes.add(primaryType);
+
+        while (remainingTypes.size() > 0) {
+            String structName = remainingTypes.get(remainingTypes.size() - 1);
+            remainingTypes.remove(remainingTypes.size() - 1);
+            deps.add(structName);
+
+            Iterator itr = types.get(primaryType).iterator();
+            while (itr.hasNext()) {
+                StructuredData.Entry entry = (StructuredData.Entry) itr.next();
+                if (!types.containsKey(entry.type)) {
+                    // Don't expand on non-user defined types
+                    continue;
+                } else if (deps.contains(entry.type)) {
+                    // Skip types which are already expanded
+                    continue;
+                } else {
+                    // Encountered a user defined type
+                    remainingTypes.add(entry.type);
+                }
+            }
+        }
+
+        return deps;
+    }
+
+    public static String encodeStruct(String structName) {
+        HashMap<String, Vector<StructuredData.Entry>> types = jsonMessageObject.types;
+
+        String structRepresentation = structName + "(";
+        for (StructuredData.Entry entry: types.get(structName)) {
+            structRepresentation += String.format("%s %s,", entry.type, entry.name);
+        }
+        structRepresentation = structRepresentation.substring(0, structRepresentation.length() - 1);
+        structRepresentation += ")";
+
+        return structRepresentation;
+    }
+
+    public static String encodeType(String primaryType) {
+        HashSet<String> deps = getDependencies(primaryType);
+        deps.remove(primaryType);
+
+        // Sort the other dependencies based on Alphabetical Order and finally add the primaryType
+        List<String> depsAsList = new ArrayList<>(deps);
+        Collections.sort(depsAsList);
+        depsAsList.add(0, primaryType);
+
+        String result = "";
+        for (String structName: depsAsList) {
+            result += encodeStruct(structName);
+        }
+
+        return result;
+    }
+
+    public static byte[] typeHash(String primaryType) {
+        return Numeric.hexStringToByteArray(sha3String(encodeType(primaryType)));
+    }
+
+    public static byte[] encodeData(
+            String primaryType,
+            LinkedHashMap<String, Object> data) throws Throwable {
+        HashMap<String, Vector<StructuredData.Entry>> types = jsonMessageObject.types;
+
+        List<String> encTypes = new ArrayList<>();
+        List<Object> encValues = new ArrayList<>();
+
+        // Add typehash
+        encTypes.add("bytes32");
+        encValues.add(typeHash(primaryType));
+
+        // Add field contents
+        for (StructuredData.Entry field: types.get(primaryType)) {
+            Object value = data.get(field.name);
+
+            if (field.type.equals("string")) {
+                encTypes.add("bytes32");
+                byte[] hashedValue = Numeric.hexStringToByteArray(sha3String((String) value));
+                encValues.add(hashedValue);
+            } else if (field.type.equals("bytes")) {
+                encTypes.add("bytes32");
+                byte[] hashedValue = sha3((byte[]) value);
+                encValues.add(hashedValue);
+            } else if (types.containsKey(field.type)) {
+                // User Defined Type
+                byte[] hashedValue = sha3(
+                        encodeData(field.type, (LinkedHashMap<String, Object>) value)
+                );
+                encTypes.add("bytes32");
+                encValues.add(hashedValue);
+            } else if (field.type.lastIndexOf(']') == field.type.length() - 1) {
+                // Still not implemented
+                throw new Throwable("TODO: Arrays currently unimplemented in encodeData");
+            } else {
+                encTypes.add(field.type);
+                encValues.add(value);
+            }
+        }
+
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        for (int i = 0; i < encTypes.size(); i++) {
+            Class<Type> typeClazz = (Class<Type>) AbiTypes.getType(encTypes.get(i));
+
+            boolean atleastOneConstructorExistsForGivenParametersType = false;
+            // Using the Reflection API to get the types of the parameters
+            Constructor[] constructors = typeClazz.getConstructors();
+            for (Constructor constructor: constructors) {
+                // Check which constructor matches
+                try {
+                    Class[] parameterTypes = constructor.getParameterTypes();
+                    byte[] temp = Numeric.hexStringToByteArray(
+                            TypeEncoder.encode(
+                                    typeClazz
+                                            .getDeclaredConstructor(parameterTypes)
+                                            .newInstance(encValues.get(i))
+                            )
+                    );
+                    baos.write(temp, 0, temp.length);
+                    atleastOneConstructorExistsForGivenParametersType = true;
+                    break;
+                } catch (IllegalArgumentException e) {
+                    continue;
+                }
+            }
+
+            if (!atleastOneConstructorExistsForGivenParametersType) {
+                throw new ValidationException(
+                        String.format(
+                                "Received an invalid argument for which no constructor"
+                                        + " exists for the ABI Class %s",
+                                typeClazz.getSimpleName()
+                        )
+                );
+            }
+        }
+        byte[] result = baos.toByteArray();
+
+        return result;
+    }
+
+    public static byte[] hashMessage(
+            String primaryType,
+            LinkedHashMap<String, Object> data) throws Throwable {
+        return sha3(encodeData(primaryType, data));
+    }
+
+    public static byte[] hashDomain() throws Throwable {
+        ObjectMapper oMapper = new ObjectMapper();
+        LinkedHashMap<String, Object> data = oMapper.convertValue(
+                jsonMessageObject.domain,
+                LinkedHashMap.class
+        );
+
+        data.put(
+                "chainId",
+                ((LinkedHashMap<String, Object>) data.get("chainId")).get("value")
+        );
+        data.put(
+                "verifyingContract",
+                ((LinkedHashMap<String, Object>) data.get("verifyingContract")).get("value")
+        );
+        return sha3(encodeData("EIP712Domain", data));
+    }
+
+    public static void validateStructuredData() throws ValidationException {
+        // Fields of Entry Objects need to follow a regex pattern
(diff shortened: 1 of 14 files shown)


stringray55:
@iikirilov could you please give me a preliminary review? The hashes are matching as per the spec. I haven't written the test cases specifically, but will do it. I just need to know the style of code that you would recommend (Hence a preliminary review). Thankyou.

iikirilov:
@stringray55 I am happy that your implementation seems to work. Let's get some tests going. That will help to restructure the code in the future.

stringray55:
@iikirilov I have added the test cases taken from the spec. Could you please review it? Thankyou.

iikirilov:
I would also like to have a failing test so we can see how your code reacts if some tries to encode/decode some garbage.

iikirilov:
Also please rebase from the 4.3 branch

stringray55:
@iikirilov I can split into the `Data Class` and the `Encoder Class`, but what is the `Decoder Class` here? Because I don't think that there is a concept of `Decdoing` in the EIP (Please correct me if wrong).

iikirilov:
> @iikirilov I can split into the `Data Class` and the `Encoder Class`, but what is the `Decoder Class` here? Because I don't think that there is a concept of `Decdoing` in the EIP (Please correct me if wrong).

This may be my lack of understanding but I assume that if you can encode a struct to a string you should be able to decode the string back to a struct

iikirilov:
Nevermind - of-course you cannot "decode" a hash so a "decoder" does not make sense here

stringray55:
@iikirilov I have added more testcases and refactored the code as you suggested. Please take a look at it.

stringray55:
@iikirilov please review. I have completely finished the implementation of `EIP712`. `WIP` has been removed now also.

iikirilov:
@snazha-blkio can I have a second pair of eyes here

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -97,9 +106,106 @@ public static String encodeType(String primaryType) {
...
+                );
+            }
+            dimensions.add(setOfDimensionsInParticularDepth.stream().findFirst().get());
+        }
+
+        return dimensions;
+    }
+
+    public static ArrayList<Object> flattenMultidimensionalArray(Object data) {
+        if (!ArrayList.class.isInstance(data)) {
```
serso: What if it's not an ArrayList but, say, LinkedList? Wouldn't a more generic List be better here? Or even Iterable?

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -109,6 +215,8 @@ public static String encodeType(String primaryType) {
         encTypes.add("bytes32");
         encValues.add(typeHash(primaryType));
 
+        String arrayTypeRegex = "^([a-zA-Z_$][a-zA-Z_$0-9]*)((\\[([1-9]\\d*)?\\])+)$";
```
serso: I guess if regexp works it works but if it doesn't somebody else would spend some time decoding it. I would at least  add a comment explaining what it matches with some examples.

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -97,9 +106,106 @@ public static String encodeType(String primaryType) {
         return Numeric.hexStringToByteArray(sha3String(encodeType(primaryType)));
     }
 
+    public static List<Integer> getArrayDimensionsFromDeclaration(String declaration) {
+        String arrayTypeRegex = "^[a-zA-Z_$][a-zA-Z_$0-9]*((\\[([1-9]\\d*)?\\])+)$";
```
serso: If you use the same regexp constant in encodeData maybe it makes sense to extract it to a class constant.

stringray55:
@serso I have made the changes, please review again. Thankyou.

stringray55:
@serso  Basically `ArrayList` is inherited from `List`  so why would it crash and why would we need to cast it . (pls correct if iam wrong )

serso:
@stringray55 it doesn't have to be an `ArrayList`, it might be `LinkedList` for example.

stringray55:
@serso I think you got the use case wrong. We initially have the message in the `JSON string format`. And then the `Jackson` library converts the whole string into a suitable `JSON Object` (which would have `LinkedHashMap` and `ArrayList` only). So I don't see any case where the data would be given as a `LinkedList`. I know that it is possible to give an argument of the type `LinkedList` for that function in general, but on the overall flow, I don't think that passing a `LinkedList` would ever happen.

serso:
@stringray55 if it's possible to make it private - I'd do so. If you expect the argument to be `ArrayList` only, then I think you should revert the initial fix and check `ArrayList.class.isInstanceOf`. I wouldn't rely on `Jackson` as it's a separate library and if they suddenly decide to switch to some other `List` implementation then the code wouldn't work.

stringray55:
@iikirilov @snazha-blkio Sorry for the delay, but the new changes have been pushed which address the above mentioned comments. Please review again. Thankyou.

/cc @ceresstation

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,450 @@
...
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import javax.xml.bind.ValidationException;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import javafx.util.Pair;
```
iikirilov: We should avoid this

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,450 @@
...
+                dimensions.add(Integer.parseInt("-1"));
+            } else {
+                dimensions.add(Integer.parseInt(currentDimension));
+            }
+        }
+
+        return dimensions;
+    }
+
+    public ArrayList<Pair> getDepthsAndDimensions(Object data, int depth) {
```
iikirilov: Can you use a map here?

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,450 @@
...
+        // groupedByDepth has key as depth and value as ArrayList(pair(Depth, Dimension))
+        Map<Object, List<Pair>> groupedByDepth = depthsAndDimensions.stream().collect(
+                Collectors.groupingBy(
+                        depthDimensionPair -> depthDimensionPair.getKey()
+                )
+        );
+
+        // depthDimensionsMap is aimed to have key as depth and value as ArrayList(Dimension)
+        Map<Integer, List<Integer>> depthDimensionsMap = new HashMap<>();
+        for (Map.Entry<Object, List<Pair>> entry : groupedByDepth.entrySet()) {
```
iikirilov: avoid using Pair

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
snazha-blkio: Or even create your own `Pair` class if there isn't already one on the classpath outside java.fx?

stringray55:
@iikirilov @snazha-blkio I would like to create my own `Pair class`, as I feel that using a `Pair` would make sense in all of the cases. Please tell me where to place the `pair.java` class file. Would it be ok if I put it in the directory `web3j/crypto/src/main/java/org/web3j/crypto/` (which is basically along with the `StructuredData.java` files)?

iikirilov:
@stringray55 you can add it in utils under the crypto package

stringray55:
@iikirilov @snazha-blkio all set. Please review again. Thankyou.

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,448 @@
...
+
+        return result;
+    }
+
+    public byte[] typeHash(String primaryType) {
+        return Numeric.hexStringToByteArray(sha3String(encodeType(primaryType)));
+    }
+
+    public List<Integer> getArrayDimensionsFromDeclaration(String declaration) {
+        Pattern arrayTypePattern = Pattern.compile(arrayTypeRegex);
```
snazha-blkio: Could you make this a field to optimize?

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,448 @@
...
+
+        // This regex tries to extract the dimensions from the
+        // square brackets of an array declaration using the ``Regex Groups``.
+        // Eg- It extracts ``5, 6, 7`` from ``[5][6][7]``
+        String arrayDimensionRegex = "\\[([1-9]\\d*)?\\]";
+        Pattern arrayDimensionPattern = Pattern.compile(arrayDimensionRegex);
```
snazha-blkio: Could you make this a field, too? No point compiling the (static) pattern each time the function is called.

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,448 @@
...
+        ArrayList<Object> dataAsArray = (ArrayList<Object>) data;
+        list.add(new Pair(depth, dataAsArray.size()));
+        for (Object subdimensionalData: dataAsArray) {
+            list.addAll(getDepthsAndDimensions(subdimensionalData,depth + 1));
+        }
+
+        return list;
+    }
+
+    public ArrayList<Integer> getArrayDimensionsFromData(
```
snazha-blkio: Could you please use `List` in your return types?

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,448 @@
...
+                dimensions.add(Integer.parseInt("-1"));
+            } else {
+                dimensions.add(Integer.parseInt(currentDimension));
+            }
+        }
+
+        return dimensions;
+    }
+
+    public ArrayList<Pair> getDepthsAndDimensions(Object data, int depth) {
```
snazha-blkio: Could you please use `List` in your return types?

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,448 @@
...
+    // Eg- arr[0][5] is not matched.
+    static String arrayTypeRegex = "^([a-zA-Z_$][a-zA-Z_$0-9]*)((\\[([1-9]\\d*)?\\])+)$";
+
+    public StructuredDataEncoder(String jsonMessageInString)
+            throws IOException, ValidationException {
+        // Parse String Message into object and validate
+        this.jsonMessageObject = parseJSONMessage(jsonMessageInString);
+    }
+
+    public HashSet<String> getDependencies(String primaryType) {
```
snazha-blkio: Could you please use `Set` in your return types?

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,448 @@
...
+                        )
+                );
+            }
+            dimensions.add(setOfDimensionsInParticularDepth.stream().findFirst().get());
+        }
+
+        return dimensions;
+    }
+
+    public ArrayList<Object> flattenMultidimensionalArray(Object data) {
```
snazha-blkio: Could you please use `List` in your return types?

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,448 @@
...
+        if (!ArrayList.class.isInstance(data)) {
```
snazha-blkio: Can you please use `List` instead? The `isAssignableFrom` method is your friend here.

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,448 @@
...
+                flattenedArray.add(otherArrayItem);
+            }
+        }
+
+        return flattenedArray;
+    }
+
+    public byte[] encodeData(
+            String primaryType,
+            LinkedHashMap<String, Object> data) throws ValidationException {
```
snazha-blkio: Any reason why you shouldn't pass in `HashMap` instead of `LinkedHashMap`?

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,448 @@
...
+                byte[] hashedValue = sha3((byte[]) value);
+                encValues.add(hashedValue);
+            } else if (types.containsKey(field.getType())) {
+                // User Defined Type
+                byte[] hashedValue = sha3(
+                        encodeData(field.getType(), (LinkedHashMap<String, Object>) value)
+                );
+                encTypes.add("bytes32");
+                encValues.add(hashedValue);
+            } else if (Pattern.matches(arrayTypeRegex, field.getType())) {
```
snazha-blkio: Are we better off pulling this out into a field so the pattern doesn't have to get compiled each time this function is run?

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,448 @@
...
+            }
+        }
+        byte[] result = baos.toByteArray();
+
+        return result;
+    }
+
+    public byte[] hashMessage(
+            String primaryType,
+            LinkedHashMap<String, Object> data) throws ValidationException {
```
snazha-blkio: Could you please use `HashMap` in your argument types?

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -0,0 +1,448 @@
...
+        // Identifier Regex matches to a valid name, but can't be an array declaration.
+        String identifierRegex = "^[a-zA-Z_$][a-zA-Z_$0-9]*$";
+        Iterator typesIterator = jsonMessageObject.getTypes().keySet().iterator();
+        while (typesIterator.hasNext()) {
+            String structName = (String) typesIterator.next();
+            List<StructuredData.Entry> fields = jsonMessageObject.getTypes().get(structName);
+            Iterator<StructuredData.Entry> fieldsIterator = fields.iterator();
+            while (fieldsIterator.hasNext()) {
+                StructuredData.Entry entry = fieldsIterator.next();
+                if (!Pattern.matches(identifierRegex, entry.getName())) {
```
snazha-blkio: Convert the compiled `Pattern` into a field?

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
stringray55: `isAssignableFrom` is giving wrong results for some reasons. Hence I haven't changed this part of the code.

stringray55:
@snazha-blkio the new review has been addressed in the latest push. Please review. Thankyou.

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
snazha-blkio: > `isAssignableFrom` is giving wrong results for some reasons. Hence I haven't changed this part of the code.

Really? Normally `List.class.isAssignableFrom(otherObject.getClass())` does the job. I can accept this for now, I presume you have some tests covering this?

snazha-blkio:
Thanks @stringray55 , looks good now.
@iikirilov , 4.3.1? 4.4.0? or 5.0?

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
stringray55: Yes @snazha-blkio there are some tests covering this, which happened to fail when I used `isAssignableFrom` method.

PR #913, opened 2019-04-11
Feature/long chainId

iikirilov:
Replaces #905  

What does this PR do?
Allows users of Web3j to set a chainId greater than 255.

This change includes 2 commits:

A piece of work conducted in e-Contract whereby Signature is updated to use a byte array for "V".
(e-Contract@6b70484)
TransactionEncoder to accept chainId as a long (and update chain constants from byte to long).
Why is it needed?
Developers working with private chains (which theoretically have a chain ID greater than 1337) are unable to use web3j to sign transactions bound for their web3 provider.

Thanks to @rain-on for initial PR

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,29 @@
+## How to contribute to web3j
+
+Thank you for reading this! If you'd like to report a bug or join in the buidl-ing of web3j, then here are some notes on how to do that.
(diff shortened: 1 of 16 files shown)


conor10:

https://community.web3labs.com/t/web3j-android-branch-harmonisation-with-master/24/6

PR #922, opened 2019-04-29
Create Type, TypeReference, Function on the fly from ABI without compiling java code

isentropy:
### What does this PR do?
Exposes methods to create TypeReference from solidity type (eg "uint256[5]") and Type from (solidity type, value Object). **This enables Function and Event encoding and decoding without translating ABI into java code with FunctionWrapperGenerator.** Includes support for static and dynamic arrays of multiple dimensions.

See the static utility methods
```
TypeReference.java: public static TypeReference makeTypeReference(String solidityType, final boolean indexed). Can handle complex array types like uint[][10].
TypeDecoder.java:   public static Type instantiateType(String solidityType, Object value)  [uses makeTypeReference() and common sense rules to coerce value to correct type]
FunctionEncoder.java: Function makeFunction(String fnname, List<String> solidityInputtypes, List<Object> arguments, List<String> solidityOutputTypes) [uses instantiateType]
```

### Why is it needed?
Currently there is no way to make function calls and decode events given an ABI without first translating the ABI to java code with FunctionWrapperGenerator and then compiling the java code. Our product, Streamr, needed to do this on the fly without compiling so we wrote these helper methods. This PR adds this functionality, which is similar to the functionality of web3 js lib.

### DIFF
--- a/abi/src/main/java/org/web3j/abi/FunctionEncoder.java
+++ b/abi/src/main/java/org/web3j/abi/FunctionEncoder.java
@@ -1,6 +1,9 @@
 package org.web3j.abi;
 
+import java.lang.reflect.InvocationTargetException;
 import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -10,6 +13,9 @@ import org.web3j.abi.datatypes.Type;
 import org.web3j.abi.datatypes.Uint;
 import org.web3j.crypto.Hash;
 import org.web3j.utils.Numeric;
+
+import static org.web3j.abi.TypeDecoder.instantiateType;
+import static org.web3j.abi.TypeReference.makeTypeReference;
 
 /**
  * <p>Ethereum Contract Application Binary Interface (ABI) encoding for functions.
@@ -87,5 +93,17 @@ public class FunctionEncoder {
         byte[] input = methodSignature.getBytes();
         byte[] hash = Hash.sha3(input);
         return Numeric.toHexString(hash).substring(0, 10);
+    }
+    public static Function encodeFunction(String fnname, List<String> solidity_inputtypes, List<Object> arguments, List<String> solidity_output_types) throws ClassNotFoundException, NoSuchMethodException, InstantiationException, IllegalAccessException, InvocationTargetException {
+        List<Type> encoded_input = new ArrayList<>();
+        Iterator argit = arguments.iterator();
+        for (String st : solidity_inputtypes) {
+            encoded_input.add(instantiateType(st, argit.next()));
+        }
+        List<TypeReference<?>> encoded_output = new ArrayList<>();
+        for (String st : solidity_output_types) {
+            encoded_output.add(makeTypeReference(st));
+        }
+        return new Function(fnname, encoded_input, encoded_output);
     }
 }
--- a/abi/src/main/java/org/web3j/abi/FunctionEncoder.java
+++ b/abi/src/main/java/org/web3j/abi/FunctionEncoder.java
@@ -43,6 +43,22 @@ public class FunctionEncoder {
         return encodeParameters(parameters, new StringBuilder());
     }
 
+    public static Function makeFunction(String fnname, List<String> solidityInputTypes,
+                                          List<Object> arguments, List<String> solidityOutputTypes)
+            throws ClassNotFoundException, NoSuchMethodException, InstantiationException,
+            IllegalAccessException, InvocationTargetException {
+        List<Type> encodedInput = new ArrayList<>();
+        Iterator argit = arguments.iterator();
+        for (String st : solidityInputTypes) {
+            encodedInput.add(instantiateType(st, argit.next()));
+        }
+        List<TypeReference<?>> encodedOutput = new ArrayList<>();
+        for (String st : solidityOutputTypes) {
+            encodedOutput.add(makeTypeReference(st));
+        }
+        return new Function(fnname, encodedInput, encodedOutput);
+    }
+
     private static String encodeParameters(List<Type> parameters, StringBuilder result) {
         int dynamicDataOffset = getLength(parameters) * Type.MAX_BYTE_LENGTH;
         StringBuilder dynamicData = new StringBuilder();
@@ -93,21 +109,5 @@ public class FunctionEncoder {
         byte[] input = methodSignature.getBytes();
         byte[] hash = Hash.sha3(input);
         return Numeric.toHexString(hash).substring(0, 10);
-    }
-
-    public static Function encodeFunction(String fnname, List<String> solidityInputtypes,
-            List<Object> arguments, List<String> solidityOutputTypes)
-            throws ClassNotFoundException, NoSuchMethodException, InstantiationException,
-            IllegalAccessException, InvocationTargetException {
-        List<Type> encodedInput = new ArrayList<>();
-        Iterator argit = arguments.iterator();
-        for (String st : solidityInputtypes) {
-            encodedInput.add(instantiateType(st, argit.next()));
-        }
-        List<TypeReference<?>> encodedOutput = new ArrayList<>();
-        for (String st : solidityOutputTypes) {
-            encodedOutput.add(makeTypeReference(st));
-        }
-        return new Function(fnname, encodedInput, encodedOutput);
     }
 }
--- a/abi/src/main/java/org/web3j/abi/FunctionEncoder.java
+++ b/abi/src/main/java/org/web3j/abi/FunctionEncoder.java
@@ -94,16 +94,20 @@ public class FunctionEncoder {
         byte[] hash = Hash.sha3(input);
         return Numeric.toHexString(hash).substring(0, 10);
     }
-    public static Function encodeFunction(String fnname, List<String> solidity_inputtypes, List<Object> arguments, List<String> solidity_output_types) throws ClassNotFoundException, NoSuchMethodException, InstantiationException, IllegalAccessException, InvocationTargetException {
-        List<Type> encoded_input = new ArrayList<>();
+
+    public static Function encodeFunction(String fnname, List<String> solidityInputtypes,
+            List<Object> arguments, List<String> solidityOutputTypes)
+            throws ClassNotFoundException, NoSuchMethodException, InstantiationException,
+            IllegalAccessException, InvocationTargetException {
+        List<Type> encodedInput = new ArrayList<>();
         Iterator argit = arguments.iterator();
-        for (String st : solidity_inputtypes) {
-            encoded_input.add(instantiateType(st, argit.next()));
+        for (String st : solidityInputtypes) {
+            encodedInput.add(instantiateType(st, argit.next()));
         }
-        List<TypeReference<?>> encoded_output = new ArrayList<>();
-        for (String st : solidity_output_types) {
-            encoded_output.add(makeTypeReference(st));
+        List<TypeReference<?>> encodedOutput = new ArrayList<>();
+        for (String st : solidityOutputTypes) {
+            encodedOutput.add(makeTypeReference(st));
         }
-        return new Function(fnname, encoded_input, encoded_output);
+        return new Function(fnname, encodedInput, encodedOutput);
     }
 }
--- a/abi/src/main/java/org/web3j/abi/TypeDecoder.java
+++ b/abi/src/main/java/org/web3j/abi/TypeDecoder.java
@@ -374,10 +374,11 @@ public class TypeDecoder {
         } else if (Utf8String.class.isAssignableFrom(rc)) {
             constructorArg = value.toString();
         } else if (Address.class.isAssignableFrom(rc)) {
-            if(value instanceof BigInteger || value instanceof Uint160)
+            if (value instanceof BigInteger || value instanceof Uint160) {
                 constructorArg = value;
-            else
+            } else {
                 constructorArg = value.toString();
+            }
         } else if (Bool.class.isAssignableFrom(rc)) {
             if (value instanceof Boolean) {
                 constructorArg = value;
--- a/abi/src/main/java/org/web3j/abi/TypeDecoder.java
+++ b/abi/src/main/java/org/web3j/abi/TypeDecoder.java
@@ -42,15 +42,32 @@ public class TypeDecoder {
 
     static final int MAX_BYTE_LENGTH_FOR_HEX_STRING = Type.MAX_BYTE_LENGTH << 1;
 
-    static <T extends Type> int getSingleElementLength(String input, int offset, Class<T> type) {
-        if (input.length() == offset) {
-            return 0;
-        } else if (DynamicBytes.class.isAssignableFrom(type)
-                || Utf8String.class.isAssignableFrom(type)) {
-            // length field + data value
-            return (decodeUintAsInt(input, offset) / Type.MAX_BYTE_LENGTH) + 2;
+    public static Type instantiateType(String solidityType, Object value) throws
+            InvocationTargetException, NoSuchMethodException, InstantiationException,
+            IllegalAccessException, ClassNotFoundException {
+        return instantiateType(makeTypeReference(solidityType), value);
+    }
+
+    public static Type instantiateType(TypeReference ref, Object value) throws
+            NoSuchMethodException, IllegalAccessException, InvocationTargetException,
+            InstantiationException, ClassNotFoundException {
+        Class rc = ref.getClassType();
+        if (Array.class.isAssignableFrom(rc)) {
+            return instantiateArrayType(ref, value);
+        }
+        return instantiateAtomicType(rc, value);
+    }
+
+    public static <T extends Array> T decode(
+            String input, int offset, TypeReference<T> typeReference) {
+        Class cls = ((ParameterizedType) typeReference.getType()).getRawType().getClass();
+        if (StaticArray.class.isAssignableFrom(cls)) {
+            return decodeStaticArray(input, offset, typeReference, 1);
+        } else if (DynamicArray.class.isAssignableFrom(cls)) {
+            return decodeDynamicArray(input, offset, typeReference);
         } else {
-            return 1;
+            throw new UnsupportedOperationException("Unsupported TypeReference: "
+                    + cls.getName() + ", only Array types can be passed as TypeReferences");
         }
     }
 
@@ -74,19 +91,6 @@ public class TypeDecoder {
         } else {
             throw new UnsupportedOperationException(
                     "Type cannot be encoded: " + type.getClass());
-        }
-    }
-
-    public static <T extends Array> T decode(
-            String input, int offset, TypeReference<T> typeReference) {
-        Class cls = ((ParameterizedType) typeReference.getType()).getRawType().getClass();
-        if (StaticArray.class.isAssignableFrom(cls)) {
-            return decodeStaticArray(input, offset, typeReference, 1);
-        } else if (DynamicArray.class.isAssignableFrom(cls)) {
-            return decodeDynamicArray(input, offset, typeReference);
-        } else {
-            throw new UnsupportedOperationException("Unsupported TypeReference: "
-                    + cls.getName() + ", only Array types can be passed as TypeReferences");
         }
     }
 
@@ -145,6 +149,103 @@ public class TypeDecoder {
         }
         return Type.MAX_BIT_LENGTH;
     }
+
+
+    static Type instantiateArrayType(TypeReference ref, Object value) throws
+            NoSuchMethodException, IllegalAccessException, InvocationTargetException,
+            InstantiationException, ClassNotFoundException {
+        List values;
+        if (value instanceof List) {
+            values = (List) value;
+        } else if (value.getClass().isArray()) {
+            values = arrayToList(value);
+        } else {
+            throw new ClassCastException("Arg of type " + value.getClass()
+                    + " should be a list to instantiate web3j Array");
+        }
+        Constructor listcons;
+        int arraySize = ref instanceof TypeReference.StaticArrayTypeReference
+                ? ((TypeReference.StaticArrayTypeReference) ref).getSize() : -1;
+        if (arraySize <= 0) {
+            listcons = DynamicArray.class.getConstructor(new Class[]{Class.class, List.class});
+        } else {
+            Class arrayClass = Class.forName("org.web3j.abi.datatypes.generated.StaticArray"
+                    + arraySize);
+            listcons = arrayClass.getConstructor(new Class[]{Class.class, List.class});
+        }
+        //create a list of transformed arguments
+        ArrayList transformedList = new ArrayList(values.size());
+        java.lang.reflect.Type subtype = ((ParameterizedType) ref.getType())
+                .getActualTypeArguments()[0];
+        for (Object o : values) {
+            TypeReference elementTR;
+            //array of arrays
+            if (subtype instanceof ParameterizedType) {
+                elementTR = new TypeReference<Array>() {
+                    @Override
+                    public java.lang.reflect.Type getType() {
+                        return subtype;
+                    }
+                };
+            } else {
+                //array of basic types
+                elementTR = TypeReference.create((Class) subtype);
+            }
+            transformedList.add(instantiateType(elementTR, o));
+        }
+        return (Type) listcons.newInstance(ref.getClassType(), transformedList);
+    }
+
+    static Type instantiateAtomicType(Class referenceClass, Object value) throws
+            NoSuchMethodException, IllegalAccessException, InvocationTargetException,
+            InstantiationException, ClassNotFoundException {
+        Object constructorArg = null;
+        if (NumericType.class.isAssignableFrom(referenceClass)) {
+            constructorArg = asBigInteger(value);
+        } else if (BytesType.class.isAssignableFrom(referenceClass)) {
+            if (value instanceof byte[]) {
+                constructorArg = value;
+            } else if (value instanceof BigInteger) {
+                constructorArg = ((BigInteger) value).toByteArray();
+            } else if (value instanceof String) {
+                constructorArg = Numeric.hexStringToByteArray((String) value);
+            }
+        } else if (Utf8String.class.isAssignableFrom(referenceClass)) {
+            constructorArg = value.toString();
+        } else if (Address.class.isAssignableFrom(referenceClass)) {
+            if (value instanceof BigInteger || value instanceof Uint160) {
+                constructorArg = value;
+            } else {
+                constructorArg = value.toString();
+            }
+        } else if (Bool.class.isAssignableFrom(referenceClass)) {
+            if (value instanceof Boolean) {
+                constructorArg = value;
+            } else {
+                BigInteger bival = asBigInteger(value);
+                constructorArg = bival == null ? null : !bival.equals(BigInteger.ZERO);
+            }
+        }
+        if (constructorArg == null) {
+            throw new InstantiationException("Could not create type " + referenceClass
+                    + " from arg " + value.toString() + " of type " + value.getClass());
+        }
+        Constructor cons = referenceClass.getConstructor(new Class[]{constructorArg.getClass()});
+        return (Type) cons.newInstance(constructorArg);
+    }
+
+    static <T extends Type> int getSingleElementLength(String input, int offset, Class<T> type) {
+        if (input.length() == offset) {
+            return 0;
+        } else if (DynamicBytes.class.isAssignableFrom(type)
+                || Utf8String.class.isAssignableFrom(type)) {
+            // length field + data value
+            return (decodeUintAsInt(input, offset) / Type.MAX_BYTE_LENGTH) + 2;
+        } else {
+            return 1;
+        }
+    }
+
 
     static int decodeUintAsInt(String rawInput, int offset) {
         String input = rawInput.substring(offset, offset + MAX_BYTE_LENGTH_FOR_HEX_STRING);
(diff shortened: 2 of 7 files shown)


isentropy:
I see some style errors. Will fix.

abi/src/main/java/org/web3j/abi/TypeDecoder.java
```diff
@@ -278,4 +282,116 @@ static Utf8String decodeUtf8String(String input, int offset) {
...
+        return rslt;
+    }
+
+    public static Type instantiateType(String solidityType, Object value) throws
+            InvocationTargetException, NoSuchMethodException, InstantiationException,
+            IllegalAccessException, ClassNotFoundException {
+        return instantiateType(makeTypeReference(solidityType), value);
+    }
+
+    public static Type instantiateType(TypeReference ref, Object value) throws
```
xaviarias: I'd suggest to split this method into smaller ones to make it more readable.

abi/src/test/java/org/web3j/abi/TypeDecoderTest.java
```diff
@@ -177,11 +215,19 @@ public void testStaticBytes() {
         assertThat(TypeDecoder.decodeBytes(
                 "6461766500000000000000000000000000000000000000000000000000000000", Bytes4.class),
                 is(dave));
+        try {
```
xaviarias: Please remove all try/catch blocks from tests (see comment below) and also `printStackTrace` calls.

abi/src/main/java/org/web3j/abi/FunctionEncoder.java
```diff
@@ -88,4 +94,20 @@ static String buildMethodId(String methodSignature) {
         byte[] hash = Hash.sha3(input);
         return Numeric.toHexString(hash).substring(0, 10);
     }
+
+    public static Function encodeFunction(String fnname, List<String> solidityInputtypes,
```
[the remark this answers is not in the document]
isentropy: done

abi/src/main/java/org/web3j/abi/TypeDecoder.java
isentropy: done. split into 2 methods: instantiateArrayType and instantiateAtomicType

abi/src/test/java/org/web3j/abi/TypeDecoderTest.java
isentropy: done

abi/src/test/java/org/web3j/abi/TypeDecoderTest.java
```diff
@@ -227,13 +273,35 @@ public void testDynamicBytes() {
...
+                is(new Address("0xbe5422d15f39373eb0a97ff8c10fbd0e40e29338")));
+        try {
+            assertThat(TypeDecoder.instantiateType("address",
+                    "0xbe5422d15f39373eb0a97ff8c10fbd0e40e29338"),
+                    is(new Address("0xbe5422d15f39373eb0a97ff8c10fbd0e40e29338")));
+            assertThat(TypeDecoder.instantiateType("address",
+                    BigInteger.ONE),
+                    is(new Address("0x0000000000000000000000000000000000000001")));
+
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
isentropy: done

abi/src/test/java/org/web3j/abi/TypeDecoderTest.java
```diff
@@ -35,6 +40,21 @@ public void testBoolDecode() {
         assertThat(TypeDecoder.decodeBool(
                 "0000000000000000000000000000000000000000000000000000000000000001", 0),
                 is(new Bool(true)));
+        try {
+            assertThat(TypeDecoder.instantiateType("bool",true),
```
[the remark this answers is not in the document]
isentropy: ok, I added line breaks between tests.

abi/src/main/java/org/web3j/abi/FunctionEncoder.java
```diff
@@ -88,4 +94,20 @@ static String buildMethodId(String methodSignature) {
...
+            List<Object> arguments, List<String> solidityOutputTypes)
+            throws ClassNotFoundException, NoSuchMethodException, InstantiationException,
+            IllegalAccessException, InvocationTargetException {
+        List<Type> encodedInput = new ArrayList<>();
+        Iterator argit = arguments.iterator();
+        for (String st : solidityInputtypes) {
```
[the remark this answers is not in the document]
isentropy: yes, changed

isentropy:
Hmmm Travis integration is failing when trying to install JDK:
`Installing oraclejdk8
$ export JAVA_HOME=~/oraclejdk8
$ export PATH="$JAVA_HOME/bin:$PATH"
$ ~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"
install-jdk.sh 2019-06-25
Expected feature release number in range of 9 to 14, but got: 8
The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .`

I think this is a Travis config issue. Should it be using jdk9?

isentropy:
> I have done a very quick review and LGTM apart from comments.
> 
> I wonder though wouldn't be possible to use this new code in the old code to avoid repeating all the type type logic?

Hi @xaviarias . Yes I think this code could replace the existing TypeReference generation code, but I just wanted to submit some static helper methods and not perform major surgery to web3j. Maybe that could be a separate PR?

**Also**: I think it would be useful to have a class that reads an ABI from string with methods to easily generate Function and Event objects. So given ABI:

`contract foo{
	function bar(address x, uint256 y) public view returns (uint balance){
		return 123;
	}

    event testEvent(
    	address indexed indexedEventParam,
    	uint nonIndexedEventParam
    )
}
`
You could call:
`
ABI foo = new ABI(fooAbiString);
Function bar = foo.getFunction("bar", asList("address", "uint256"), asList("0xabc",1));
Event testEvent = foo.getEvent("testEvent","address", "uint");
`
Maybe another PR?

xaviarias:
> @xaviarias I made the changes you suggested.
> 
> Travis integration is failing when trying to install JDK:
> `Installing oraclejdk8 $ export JAVA_HOME=~/oraclejdk8 $ export PATH="$JAVA_HOME/bin:$PATH" $ ~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL" install-jdk.sh 2019-06-25 Expected feature release number in range of 9 to 14, but got: 8 The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .`
> 
> I think this is a Travis config issue. Should it be using jdk9?

Hi @isentropy I'm experiencing the same error in other branches, not sure what's the issue but I'll check it. Will ping when is ready to merge this PR.

xaviarias:
> > I have done a very quick review and LGTM apart from comments.
> > I wonder though wouldn't be possible to use this new code in the old code to avoid repeating all the type type logic?
> 
> Hi @xaviarias . Yes I think this code could replace the existing TypeReference generation code, but I just wanted to submit some static helper methods and not perform major surgery to web3j. Maybe that could be a separate PR?
> 
> **Also**: I think it would be useful to have a class that reads an ABI from string with methods to easily generate Function and Event objects. So given ABI:
> 
> contract foo{
> function bar(address x, uint256 y) public view returns (uint balance){
> return 123;
> }
> 
> event testEvent(
> address indexed indexedEventParam,
> uint nonIndexedEventParam
> )
> }
> 
> You could call:
> 
> ABI foo = new ABI(fooAbiString);
> Function bar = foo.getFunction("bar", asList("address", "uint256"), asList("0xabc",1));
> Event testEvent = foo.getEvent("testEvent","address", "uint");
> 
> Maybe another PR?

Ideally this "major surgery" should be done in this PR or immediately after.

IMO adding so much code and pushing the refactoring for later rarely results in any real action because the scope of the PR will be lost with the time. I fear we'll end up having many duplicated logic and no-one will remember doing something about it. 

We should also consider that potentially new networks will be integrated with web3 (Aion and others on the way), so we need a very concise and generic way for encoding and decoding types. Could you propose this refactoring PR to simplify the logic?

About the proposal to have a class that reads an ABI from string, I think is out of scope of web3j. The code you've shown is not an ABI, is Solidity code, and the work of parsing and extracting the ABI is done by the Solidity compiler:
```
contract foo{
function bar(address x, uint256 y) public view returns (uint balance){
return 123;
}
```
We might get into muddy waters if we want to do everything within web3j, because there are many Solidity versions and things to consider before doing a parser that works properly. Also we should bear in mind that code generation from an ABI should be kept independent of the language, for that same reason of generality. Other smart contract languages might be coming into play in the near future.

isentropy:
@xaviarias oops, I meant the ABI described by that Solidity code:
`[
  {
    "constant": true,
    "inputs": [
      {
        "name": "x",
        "type": "address"
      },
      {
        "name": "y",
        "type": "uint256"
      }
    ],
    "name": "bar",
    "outputs": [
      {
        "name": "balance",
        "type": "uint256"
      }
    ],
    "payable": false,
    "stateMutability": "view",
    "type": "function"
  },
  {
    "anonymous": false,
    "inputs": [
      {
        "indexed": true,
        "name": "indexedEventParam",
        "type": "address"
      },
      {
        "indexed": false,
        "name": "nonIndexedEventParam",
        "type": "uint256"
      }
    ],
    "name": "testEvent",
    "type": "event"
  }
]
`
so my proposed ABI class would take in the ABI string above, not compile Solidity

isentropy:
> Ideally this "major surgery" should be done in this PR or immediately after.
> 
> IMO adding so much code and pushing the refactoring for later rarely results in any real action because the scope of the PR will be lost with the time. I fear we'll end up having many duplicated logic and no-one will remember doing something about it.
> 
> We should also consider that potentially new networks will be integrated with web3 (Aion and others on the way), so we need a very concise and generic way for encoding and decoding types. Could you propose this refactoring PR to simplify the logic?

OK, I'll look into replacing existing TypeReference generation with this code. I think that's healthy in the long term. Just not sure how much needs to be refactored.

abi/src/test/java/org/web3j/abi/TypeDecoderTest.java
xaviarias: The `try`...`catch` comment relates to all tests and there are many of them still there.

isentropy: Eh, looks like all try/catch were removed from https://github.com/streamr-dev/web3j/blob/master/abi/src/test/java/org/web3j/abi/TypeDecoderTest.java. The code snippet above is outdated. This PR only modified TypeDecoderTest.

What are you referring to?

xaviarias: Sorry my bad, if there are no try/catch blocks in the test that's fine for me.

xaviarias:
> @xaviarias oops, I meant the ABI described by that Solidity code:
> `[ { "constant": true, "inputs": [ { "name": "x", "type": "address" }, { "name": "y", "type": "uint256" } ], "name": "bar", "outputs": [ { "name": "balance", "type": "uint256" } ], "payable": false, "stateMutability": "view", "type": "function" }, { "anonymous": false, "inputs": [ { "indexed": true, "name": "indexedEventParam", "type": "address" }, { "indexed": false, "name": "nonIndexedEventParam", "type": "uint256" } ], "name": "testEvent", "type": "event" } ] `
> so my proposed ABI class would take in the ABI string above, not compile Solidity. Do you have a problem with ABI class as described above that takes in ABI string?

Is not that I have a problem with such class, the thing is that adding code is fantastic as long as its functionality is clearly defined and targeting and issue or a specific feature or enhancement, and does not overlap with existing code.

If a new class overlaps with existing code, like in this case happens with `org.web3j.codegen.SolidityFunctionWrapperGenerator#loadContractDefinition(File absFile)` returning a `List<org.web3j.protocol.core.methods.response.AbiDefinition>`, then the PR has to replace the old code with the new one.

This changes also impact other open PRs that are currently being reviewed, so before adding more code  we need to make sure we cover all requirements. I agree that ABI loading should be encapsulated but I'd think more about an abstract `AbiDefinitionParser` or loader that can convert `AbiDefinition`s from a string, input stream or a file. Loading a specific ABI format ABI would be implemented by subclasses like `SolidityAbiDefinitionParser`, etc.

So if we want want to encapsulate ABI loading, then we should use an ABI parser or loader; if we want to create `Function` instances easily, then consider using a FunctionBuilder or something similar, but we must keep the concerns clearly separated or we'll end up having a potpourri that will make very difficult to generify in the future to support many ABI definition types.

This ABI class would be fine as a facade for an end-user API, but not for internal use because such functionality exists already.

isentropy:
> > Ideally this "major surgery" should be done in this PR or immediately after.
> > IMO adding so much code and pushing the refactoring for later rarely results in any real action because the scope of the PR will be lost with the time. I fear we'll end up having many duplicated logic and no-one will remember doing something about it.
> > We should also consider that potentially new networks will be integrated with web3 (Aion and others on the way), so we need a very concise and generic way for encoding and decoding types. Could you propose this refactoring PR to simplify the logic?
> 
> OK, I'll look into replacing existing TypeReference generation with this code. I think that's healthy in the long term. Just not sure how much needs to be refactored.
> 
> I do see value in merging this as is tho. There isn't really a way to make function calls or decode Event logs given ABI without compiling Java code at the moment, and these helpers would solve that issue.

@xaviarias I think this is much easier than I expected. I can just replace the following method in SolidityFunctionWrapper:

`static TypeName buildTypeName(String typeDeclaration) {
            typeDeclaration = typeDeclaration.split("\\s+")[0];
            return TypeName.get(TypeReference.makeTypeReference(typeDeclaration).getType());
}`

all tests pass except for 1 about an oversized static array (length 33 when only StaticArray32 is defined). I will commit when I have the oversized array issue fixed.

xaviarias:
> > @xaviarias I made the changes you suggested.
> > Travis integration is failing when trying to install JDK:
> > `Installing oraclejdk8 $ export JAVA_HOME=~/oraclejdk8 $ export PATH="$JAVA_HOME/bin:$PATH" $ ~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL" install-jdk.sh 2019-06-25 Expected feature release number in range of 9 to 14, but got: 8 The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .`
> > I think this is a Travis config issue. Should it be using jdk9?
> 
> Hi @isentropy I'm experiencing the same error in other branches, not sure what's the issue but I'll check it and ping you when is ready to merge this PR.

Replacing the existing JDK configuration in `.travis.yml` from `oraclejdk8` to `openjdk8` as per [this issue](https://travis-ci.community/t/the-travis-ci-build-could-not-be-completed-due-to-an-error/1345/13) seems to fix the build.

isentropy:
and fixing SolidityFunctionWrapper.buildTypeName would probably add support for arrays of dimension greater than 2

xaviarias:
> and fixing SolidityFunctionWrapper.buildTypeName would probably add support for arrays of dimension greater than 2

We are planning changing a few things on arrays implementation to support Aion network and n-sized static arrays, so we'll take the opportunity to fix this also if not done yet.

isentropy:
All tests pass now and SolidityFunctionWrapper.buildTypeName() uses TypeReference.makeTypeReference() as it should. 

N dimensional arrays are supported by TypeReference.makeTypeReference() and TypeDecoder.instantiateType(), **but unfortunately they won't work in generated java code without some more involved changes to SolidityFunctionWrapper**:  SolidityFunctionWrapper.createMappedParameterTypes() must swap out Utils.typeMap(), which takes Class reference type, for TypeDecoder.instantiateType(), which takes TypeReference as reference type.

I think this is a good stopping point for this PR. Everything works, and there's a clear path forward for N-dimensional array support deprecating Utils.typeMap() in favor of TypeDecoder.instantiateType().

isentropy:
@xaviarias I have working code (not committed) to patch SolidityFunctionWrapper to use  TypeDecoder.instantiateType() in code generation instead of Utils.typeMap(), **which adds N dimensional array support to java code generation.** An example autogenerated Solidity foo(string x, uint[][][]) is shown here:
`
public RemoteCall<BigInteger> bar(String x, List<List<List<BigInteger>>> y) {
  final Function function = new Function(FUNC_BAR, 
                Arrays.<Type>asList(new org.web3j.abi.datatypes.Address(x), 
                org.web3j.abi.TypeDecoder.instantiateTypeNoThrow("uint256[][][]", y)), 
                Arrays.<TypeReference<?>>asList(new TypeReference<Uint256>() {}));
        return executeRemoteCallSingleValueReturn(function, BigInteger.class);
    }
'

But there are lots of unit tests that would need to be rewritten and other codegen pieces still use Utils.typeMap(). I would prefer doing this change in another PR.

xaviarias:
> @xaviarias I have working code (not committed) to patch SolidityFunctionWrapper to use TypeDecoder.instantiateType() in code generation instead of Utils.typeMap(), **which adds N dimensional array support to java code generation.** An example autogenerated Solidity `foo(string x, uint[][][])` is shown here:
> 
> ```java
> public RemoteCall<BigInteger> bar(String x, List<List<List<BigInteger>>> y) {
>   final Function function = new Function(FUNC_BAR, 
>                 Arrays.<Type>asList(new org.web3j.abi.datatypes.Address(x), 
>                 org.web3j.abi.TypeDecoder.instantiateTypeNoThrow("uint256[][][]", y)), 
>                 Arrays.<TypeReference<?>>asList(new TypeReference<Uint256>() {}));
>         return executeRemoteCallSingleValueReturn(function, BigInteger.class);
>     }
> ```
> 
> But there are lots of unit tests that would need to be rewritten and other codegen pieces still use Utils.typeMap(). I would prefer doing this change in another PR.

Hi @isentropy and sorry for the late reply. Great job, it looks very good now! I think unit tests should be passing before merging into master. We need to make sure that those changes do not introduce regression into the existing code. Is that very long to implement ?

isentropy:
> Update tests
You mean you want me to integrate instantiateType() into to Java wrapper generation (as in `instantiateTypeNoThrow("uint256[][][]", y)`) in this PR? makeTypeReference() is already integrated in this PR with tests passing. @xaviarias **can we discuss on a short phone call?** Streamr doesn't use the java wrappers at all so they don't want to sponsor work on that. It might require quite a bit of work because there are a lot of unit tests to update and other codegen pieces to work on to fully deprecate Utils.typeMap(). That's why I prefer to submit another PR for that. Can you email me contact info at jwolff@isentropy.com? I'm on Signal and WhatsApp.

Re recursion: Yes, there is recursion in makeTypeRefernce but it only recurses the # of array dimensions so I didn't think it's a huge issue. I'll think about how to remove. Simple solution is just to limit the recursion depth by some constant [maybe like 26 for those doing [string theory](https://en.wikipedia.org/wiki/String_theory) calculations in EVM]. Does that work for you?

abi/src/main/java/org/web3j/abi/TypeReference.java
```diff
@@ -16,6 +25,7 @@
  */
 public abstract class TypeReference<T extends org.web3j.abi.datatypes.Type>
         implements Comparable<TypeReference<T>> {
+    protected static Pattern ARRAY_SUFFIX = Pattern.compile("\\[(\\d*)\\]$");
```
xaviarias: IntelliJ reports a warning: `Redundant character scape \\`

abi/src/main/java/org/web3j/abi/TypeReference.java
```diff
@@ -88,4 +122,81 @@ public int getSize() {
...
+    public static TypeReference makeTypeReference(String solidityType) throws
+            ClassNotFoundException {
+        return makeTypeReference(solidityType,false);
+    }
+
+    public static TypeReference makeTypeReference(String solidityType, final boolean indexed)
+            throws ClassNotFoundException {
+        Matcher m = ARRAY_SUFFIX.matcher(solidityType);
+        if (!m.find()) {
+            final Class tc = getAtomicTypeClass(solidityType);
```
xaviarias: Review variable naming `tc`

abi/src/main/java/org/web3j/abi/TypeReference.java
```diff
@@ -88,4 +122,81 @@ public int getSize() {
...
+            return create(tc, indexed);
```
xaviarias: Unchecked assignment warning

abi/src/main/java/org/web3j/abi/TypeReference.java
```diff
@@ -88,4 +122,81 @@ public int getSize() {
...
+        }
+        String arraySize = m.group(1);
+        TypeReference baseTr = makeTypeReference(solidityType.substring(0,solidityType.length()
+                - m.group(0).length()));
```
xaviarias: Review formatting, there are missing spaces here and other places

abi/src/main/java/org/web3j/abi/TypeReference.java
```diff
@@ -88,4 +122,81 @@ public int getSize() {
...
+        } else {
+            final Class arrayclass;
+            if (Integer.parseInt((arraySize)) <= StaticArray.MAX_SIZE_OF_STATIC_ARRAY) {
+                arrayclass = Class.forName("org.web3j.abi.datatypes.generated.StaticArray"
+                        + arraySize);
+            } else {
+                arrayclass = Class.forName("org.web3j.abi.datatypes.StaticArray");
+            }
+            ref = new TypeReference
+                    .StaticArrayTypeReference<StaticArray>(Integer.parseInt(arraySize)) {
```
xaviarias: `Integer.parseInt(arraySize)` is already computed a few lines above.

abi/src/main/java/org/web3j/abi/TypeReference.java
```diff
@@ -88,4 +122,81 @@ public int getSize() {
...
+                    };
+                }
+            };
+        } else {
+            final Class arrayclass;
+            if (Integer.parseInt((arraySize)) <= StaticArray.MAX_SIZE_OF_STATIC_ARRAY) {
+                arrayclass = Class.forName("org.web3j.abi.datatypes.generated.StaticArray"
+                        + arraySize);
+            } else {
+                arrayclass = Class.forName("org.web3j.abi.datatypes.StaticArray");
```
xaviarias: Should not an error be reported instead of using and abstract type?

abi/src/main/java/org/web3j/abi/TypeReference.java
isentropy: hmmm that regex looks correct to me and IntelliJ shows error with "\[(\\d*)\]$" and "\\[(\d*)\\]$". I think you need "\\" there to created the regex string `\[(\d*)\]$`.

abi/src/main/java/org/web3j/abi/TypeReference.java
isentropy: ok, I'll rename typeClass

abi/src/main/java/org/web3j/abi/TypeReference.java
xaviarias: `\[(\d*)]$` does not show any error for me and removing `\\` removes the warning.

isentropy: hmmm that looks correct to me because we want to construct the regex string `\[(\d*)\]$`. What do you suggest?

abi/src/main/java/org/web3j/abi/TypeReference.java
isentropy: ha that's what I think, but there's an existing unit test that was failing because uint[33] should return StaticArray! That was the existing behavior. I can remove the existing unit test.

abi/src/main/java/org/web3j/abi/TypeReference.java
isentropy: ok, if that works. I guess ] without [ is interpreted literally.

xaviarias:
> > Update tests
> 
> You mean you want me to integrate instantiateType() into to Java wrapper generation (as in `instantiateTypeNoThrow("uint256[][][]", y)`) in this PR? makeTypeReference() is already integrated in this PR with tests passing. @xaviarias **can we discuss on a short phone call?** Streamr doesn't use the java wrappers at all so they don't want to sponsor work on that. It might require quite a bit of work because there are a lot of unit tests to update and other codegen pieces to work on to fully deprecate Utils.typeMap(). That's why I prefer to submit another PR for that. Can you email me contact info at [jwolff@isentropy.com](mailto:jwolff@isentropy.com)? I'm on Signal and WhatsApp.
> 
> Re recursion: Yes, there is recursion in makeTypeReference() but it only recurses the # of array dimensions so I didn't think it's a huge issue. I'll think about how to remove. Simple solution is just to limit the recursion depth by some constant [maybe like 26 for those doing [string theory](https://en.wikipedia.org/wiki/String_theory) calculations in EVM]. Does that work for you?
> 
> To rewrite without recursion probably involves working in the other direction, ie makeTypeRefence("uint[1][]") would make Uint, then StaticArray1, then DynamicArray. It's doable.

Ok I see your point, code generation could be done in another PR yes. The problem is that: 

1) Still I have added a few comments to review, and `makeTypeReference` is still very complex method without any comment or Javadoc. Probably going deeply through the code I could find more things, and right now is not a priority to support streaming.

2) We are facing a problem with web3j type system in general and we are rethinking the approach for a next release. Also the Aion branch has some changes conflicting with this PR and we need to release that very soon.

So to be honest I really think we should merge your PR, but I don't feel comfortable in doing it right now. There are too many open questions to add such a big amount of code that will need to be reviewed and merged very soon. But is now going to be long, I reckon during the next 1-2 months we can re-open the discussion. What do you think, is there an urgency for you?

isentropy:
> > > Update tests
> > 
> > 
> > You mean you want me to integrate instantiateType() into to Java wrapper generation (as in `instantiateTypeNoThrow("uint256[][][]", y)`) in this PR? makeTypeReference() is already integrated in this PR with tests passing. @xaviarias **can we discuss on a short phone call?** Streamr doesn't use the java wrappers at all so they don't want to sponsor work on that. It might require quite a bit of work because there are a lot of unit tests to update and other codegen pieces to work on to fully deprecate Utils.typeMap(). That's why I prefer to submit another PR for that. Can you email me contact info at [jwolff@isentropy.com](mailto:jwolff@isentropy.com)? I'm on Signal and WhatsApp.
> > Re recursion: Yes, there is recursion in makeTypeReference() but it only recurses the # of array dimensions so I didn't think it's a huge issue. I'll think about how to remove. Simple solution is just to limit the recursion depth by some constant [maybe like 26 for those doing [string theory](https://en.wikipedia.org/wiki/String_theory) calculations in EVM]. Does that work for you?
> > To rewrite without recursion probably involves working in the other direction, ie makeTypeRefence("uint[1][]") would make Uint, then StaticArray1, then DynamicArray. It's doable.
> 
> Ok I see your point, code generation could be done in another PR yes. The problem is that:
> 
> 1. Still I have added a few comments to review, and `makeTypeReference` is still very complex method without any comment or Javadoc. Probably going deeply through the code I could find more things, and right now is not a priority to support streaming.
> 2. We are facing a problem with web3j type system in general and we are rethinking the approach for a next release. Also the Aion branch has some changes conflicting with this PR and we need to release that very soon.

Yes the type system in web3j is complicated and not ideal! That's why I think it's important to include these helper methods.

> 
> So to be honest I really think we should merge your PR, but I don't feel comfortable in doing it right now. There are too many open questions to add such a big amount of code that will need to be reviewed and merged very soon. But is now going to be long, I reckon during the next 1-2 months we can re-open the discussion. What do you think, is there an urgency for you?

Well we're using these helper methods now in our core service in production and have launches coming up in the next weeks. It would be helpful for us to have them incorporated into web3j, but of course it's your decision. I think there's value to just adding the static helpers, which could then be built upon later [eg improvements to java wrapper]. They add a lot of useful functionality and I'm sure others would appreciate too. I think the existing java wrapper generation and compilation are dealbreakers for many use cases.

isentropy:
I'll do the minor changes Xavi suggested soon

isentropy:
@xaviarias I made the minor changes you requested

isentropy:
@xaviarias  I reworked makeTypeReference() to **remove recursion**. It works in the reverse order as described in above [comment](https://github.com/web3j/web3j/pull/922#issuecomment-510849446), ie makeTypeRefence("uint[1][]") makes Uint, then StaticArray1, then DynamicArray.

isentropy:
@xaviarias  I improved the way in which instantiateType coerces the arguments to the Array's constructor, and added a test case to show that **3D array support is working in makeTypeReference() and instantiateType()**. Which is another benefit of this PR.

I'm not planning any more changes to this PR, but tell me if you need something.

iikirilov:
@xaviarias are you happy with the changes? I can pick up anything else that needs to be done here.

isentropy:
Will anyone from Web3 Labs attend the Web3 Summit in Berlin in Aug? I'll be there and will demo Streamr's Ethereum workflow that uses this code at the Chainlink Hackernode on Aug 21.

iikirilov:
Unfortunately no. We will be at devcon though :)

isentropy:
> Unfortunately no. We will be at devcon though :)

I'm working on that :)

iikirilov:
replaced by #997

PR #926, opened 2019-05-04
925: initial cut of issue template

snazha-blkio:
### What does this PR do?
Specifies the code template that will be presented to the user when raising issues.

### Why is it needed?
We are getting too many issues with very little information besides an exception.


iikirilov:
Anything else that needs to be done to get this merged?

iikirilov:
replaced by #949

Issue #896, opened 2019-03-20
Error while compiling wrapper class generated from a crowdsale contract extending from RefundablePostDeliveryCrowdsale. Error: cannot override finalize().


sowmyakannan:
I have created a crowdsale contract that extends from RefundablePostDeliveryCrowdsale. web3j generates the wrapper class successfully.  However the wrapper classes do not compile and give an error like so:

Error:(204, 43) java: finalize() in com.reit.resreitsmartcontractwrapper.TokenSale cannot override finalize() in java.lang.Object
  return type org.web3j.protocol.core.RemoteCall<org.web3j.protocol.core.methods.response.TransactionReceipt> is not compatible with void.

I am overriding the finalize() method in my crowdsale contract that extends from RefundablePostDeliveryCrowdsale which extends Finalizable that has a finalize() method to be overridden. 

Any suggestions or workarounds for this issue?

iikirilov:
finalize() is a protected method in `Object.java`. The workaround would be to rename the method in the generated class. It will still call finalize() in the contract.

A long-term solution would be to modify the code generator such that it prepends protected methods from Object.java with an underscore

AlexandrouR:
Looks like this was solved by #776

iikirilov:
#776 is related but `finalize()` is a method and is not covered (i.e. it is not a keyword in java). So this needs to check against the method list of `Object.java` and probably the method list of `Contract.java`

AlexandrouR:
The Object class has a finalize method that is protected. If it is within your control you can change the method name in the contract or write a wrapper around it. This is the only workaround I can think of.

PR #945, opened 2019-06-23
Fix bug that contract method is duplicate with names as ones from Object.java

leejh903:
### What does this PR do?
This PR solved the issue(#896) 

just SolidityFunctionWrapper.java

### Why is it needed?
As you can see issue(#896), there is compile error when contract method name is duplicate with names as ones from Object.java, especially there are no parameters. This PR adds more condition which append '_' in front of method name in this situation.


codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -98,6 +99,8 @@
     private static final Pattern pattern = Pattern.compile(regex);
     private final GenerationReporter reporter;
 
+    private static Set<String> parentsMethodNames = getParentsMethodNames();
```
iikirilov:
```suggestion
    private static final Set<String> restrictedMethodNames = new HashSet<>(Arrays.asList(Object.class.getDeclaredMethods()));
```

We only need to build this set once

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -686,7 +689,7 @@ MethodSpec buildFunction(
 
         // If the solidity function name is a reserved word
         // in the current java version prepend it with "_"
-        if (!SourceVersion.isName(functionName)) {
+        if (!SourceVersion.isName(functionName) || duplicateWithParentsMethods(functionName)) {
```
iikirilov:
```suggestion
        if (!SourceVersion.isName(functionName) || restrictedMethodNames.contains(functionName)) {
```

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -708,6 +711,19 @@ MethodSpec buildFunction(
         return methodBuilder.build();
     }
 
+    private boolean duplicateWithParentsMethods(String functionName) {
```
iikirilov: these two methods can be removed

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
leejh903: I think suggested code is more better than previous one!

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
leejh903: I removed two methods. Thanks

leejh903:
@iikirilov Thanks for your review. The suggested code has a compile error because the return type of `new HashSet<>(Arrays.asList(Object.class.getDeclaredMethods()));` is HashSet<Method> and it is not HashSet<String>. So I changed your suggested code and add test code related with this code.

leejh903:
@iikirilov I recovered some conflicts

codegen/src/test/java/org/web3j/codegen/SolidityFunctionWrapperTest.java
```diff
@@ -12,11 +12,10 @@
  */
 package org.web3j.codegen;
 
+import java.lang.reflect.Method;
 import java.math.BigInteger;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
```
iikirilov: please remove wildcard

leejh903: I reflected your feedback!

leejh903:
Close this PR because function name collision solved adding prefix in front of function name.

PR #952, opened 2019-07-05
Mozzy. Bpllane

RLlane9:
### What does this PR do?
*required*
Import export
### Where should the reviewer start?true
*required*
True
### Why is it needed?
*required*
To do and get sucess


iikirilov:
Hi @RLlane9. This PR contains many changes and is forked from an old version of web3j.

Could you please let me know what your intentions are with this?

Right now I don't see any benefits...

Issue #576, opened 2018-06-05
EthBlock throws MessageDecodingException if size is null and getSize() is called


assafY:
In many early uncle blocks the size field is null.

This is the code in `EthBlock` for `getSize()`:

`return Numeric.decodeQuantity(size);`

This throws a `MessageDecodingException` if size is null.

PR #969, opened 2019-07-25
Issue576

AlexandrouR:
### What does this PR do?
Changed getSize method of ethblock to return zero if the block size is null.

EthBlock get size method.

### Why is it needed?
To address issue #576

### DIFF
--- a/core/src/main/java/org/web3j/protocol/core/methods/response/EthBlock.java
+++ b/core/src/main/java/org/web3j/protocol/core/methods/response/EthBlock.java
@@ -266,7 +266,7 @@ public class EthBlock extends Response<EthBlock.Block> {
         }
 
         public BigInteger getSize() {
-            return size != null ? Numeric.decodeQuantity(size) : Numeric.decodeQuantity("0x0") ;
+            return size != null ? Numeric.decodeQuantity(size) : Numeric.decodeQuantity("0x0");
         }
 
         public String getSizeRaw() {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/web3j/protocol/core/methods/response/EthBlock.java
```diff
@@ -267,7 +267,7 @@ public void setExtraData(String extraData) {
         }
 
         public BigInteger getSize() {
-            return Numeric.decodeQuantity(size);
+            return size != null ? Numeric.decodeQuantity(size) : Numeric.decodeQuantity("0x0");
```
iikirilov:
```suggestion
            return size != null ? Numeric.decodeQuantity(size) : BigInteger.ZERO;
```

core/src/test/java/org/web3j/protocol/core/methods/response/EthBlockTest.java
```diff
@@ -0,0 +1,42 @@
...
+public class EthBlockTest {
+
+    @Test
+    public void testEthBlockNullSize() {
+        EthBlock.Block ethBlock =
+                new EthBlock.Block(
+                        null, null, null, null, null, null, null, null, null, null, null, null,
+                        null, null, null, null, null, null, null, null, null, null);
+
+        assertEquals(ethBlock.getSize(), BigInteger.valueOf(0));
```
iikirilov:
```suggestion
        assertEquals(ethBlock.getSize(), BigInteger.ZERO);
```

PR #971, opened 2019-07-25
Support kotlin format in tuples

alex-forshtat-tbk:
We use web3j in both our Java and Kotlin code. There exists a built-in format for tuples in Kotlin, which prettifies the decomposing adssignment, and it is nice to have it in web3j code: 
val (a,b,c) = myContract.methodReturnTupple().send();

This required adding "componentN()" methods to Tuple class. The original getValueN()) methods were left as-is for backward-compatibility and they can still be used in java and Kotlin.

### DIFF
--- a/tuples/src/main/java/org/web3j/tuples/generated/Tuple10.java
+++ b/tuples/src/main/java/org/web3j/tuples/generated/Tuple10.java
@@ -48,7 +48,15 @@ public final class Tuple10<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10> implements T
         return value1;
     }
 
+    public T1 component1() {
+        return value1;
+    }
+
     public T2 getValue2() {
+        return value2;
+    }
+
+    public T2 component2() {
         return value2;
     }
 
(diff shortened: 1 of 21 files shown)


tuples/src/main/java/org/web3j/tuples/generated/Tuple10.java
```diff
@@ -48,42 +48,82 @@ public T1 getValue1() {
         return value1;
     }
 
+    public T1 component1() {
+        return value1;
+    }
+
     public T2 getValue2() {
```
iikirilov: It is worth modifying the code generator to make these methods @Deprecated and move across to the `componentN` convention for web3j 5.0

alex-forshtat-tbk: Addressed, please see the last commit.

iikirilov:
waiting for my colleague @sebaraba to review and we can merge

alex-forshtat-tbk:
@iikirilov @sebaraba @conor10  Hi! I just saw that it is approved. Thanks a lot! I am just wondering if there is anything else I need to do now to help you get it merged?

iikirilov:
Nope, just some patience - will be merged in the next couple of days.

PR #972, opened 2019-07-25
Support encoding a function without actually calling it

alex-forshtat-tbk:
There is an easy way in Web3.js to perform an 'encode with selector' operation with the contract object, like this:
let encodedAbi = myContract.methods.myMethod(myParameter).encodeABI()

Unfortunately, this was not possible with web3j, as encoding and performing the call were highly coupled. This PR wraps the returned 'RemoteCall' object with a 'RemoteFunctionCall' that allows accessing and encoding the function if needed, without breaking the existing API.


iikirilov:
this will be merged after #922

PR #973, opened 2019-07-25
Estimated gas provider

forshtat:
### What does this PR do?
Creates an EstimatedGasProvider class that provides (arguably) the most common strategy for determining the gas price and gas limit in Ethereum, i.e. asking the node for both values. This is also the behaviour that all Web3.js customers are used to.

This does require a change to the existing ContractGasProvider interface, which does not seem logical as 'contract function name' is not what usually defines the gas price or limit.

See if EstimatedGasProvider makes more sense than the DefaultGasProvider. 

### Why is it needed?
Dapp developers must have a way to estimate the gas cost that is built into the tools they use. Currently, the provided default gives a hard-coded value, which is:
GAS_LIMIT = 4_300_000
GAS_PRICE = 22_000_000_000
That equals to $20.98 on Ethereum mainnet

### DIFF
--- a/core/src/main/java/org/web3j/tx/Contract.java
+++ b/core/src/main/java/org/web3j/tx/Contract.java
@@ -197,26 +197,6 @@ public abstract class Contract extends ManagedTransaction {
     }
 
     /**
-     * Allow {@code gasPrice} to be set.
-     *
-     * @param newPrice gas price to use for subsequent transactions
-     * @deprecated use ContractGasProvider
-     */
-    public void setGasPrice(BigInteger newPrice) {
-        this.gasProvider = new StaticGasProvider(newPrice, gasProvider.getGasLimit());
-    }
-
-    /**
-     * Get the current {@code gasPrice} value this contract uses when executing transactions.
-     *
-     * @return the gas price set on this contract
-     * @deprecated use ContractGasProvider
-     */
-    public BigInteger getGasPrice() {
-        return gasProvider.getGasPrice();
-    }
-
-    /**
      * Check that the contract deployed at the address associated with this smart contract wrapper
      * is in fact the contract you believe it is.
      *
@@ -358,13 +338,15 @@ public abstract class Contract extends ManagedTransaction {
             String data, BigInteger weiValue, String funcName, boolean constructor)
             throws TransactionException, IOException {
 
+        BigInteger gasPrice = gasProvider.getGasPrice();
+        BigInteger gasLimit = gasProvider.getGasLimit(transactionManager.getFromAddress(), gasPrice, contractAddress, weiValue, data);
         TransactionReceipt receipt =
                 send(
                         contractAddress,
                         data,
                         weiValue,
-                        gasProvider.getGasPrice(funcName),
-                        gasProvider.getGasLimit(funcName),
+                        gasPrice,
+                        gasLimit,
                         constructor);
 
         if (!receipt.isStatusOK()) {
--- a/core/src/main/java/org/web3j/tx/gas/ContractGasProvider.java
+++ b/core/src/main/java/org/web3j/tx/gas/ContractGasProvider.java
@@ -12,16 +12,27 @@
  */
 package org.web3j.tx.gas;
 
+import java.io.IOException;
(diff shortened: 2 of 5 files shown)


core/src/main/java/org/web3j/tx/Contract.java
```diff
@@ -333,13 +333,14 @@ private TransactionReceipt executeTransaction(Function function, BigInteger weiV
     TransactionReceipt executeTransaction(String data, BigInteger weiValue, String funcName)
             throws TransactionException, IOException {
 
+        BigInteger gasPrice = gasProvider.getGasPrice(funcName);
```
iikirilov:
```suggestion
```

core/src/main/java/org/web3j/tx/Contract.java
```diff
@@ -333,13 +333,14 @@ private TransactionReceipt executeTransaction(Function function, BigInteger weiV
...
         TransactionReceipt receipt =
                 send(
                         contractAddress,
                         data,
                         weiValue,
-                        gasProvider.getGasPrice(funcName),
-                        gasProvider.getGasLimit(funcName));
+                        gasPrice,
```
iikirilov:
```suggestion
                        gasProvider.getGasPrice(funcName),
```

core/src/main/java/org/web3j/tx/gas/ContractGasProvider.java
```diff
@@ -20,8 +20,11 @@
     @Deprecated
```
iikirilov: I don't understand why this method is `@Deprecated`. There shouldn't be a link between `contractFunc` and `gasPrice`.

This can be modified in this PR also.

alex-forshtat-tbk: So what is the desired interface of the ContractGasProvider? I am not familiar with how do other developers implement or use it. In my understanding, the 'getGasPrice' does not need any parameters as it is not something that depends on the kind of transaction you are about to make. 
On the other hand, 'getGasLimit' needs all the parameters, and it can throw the exception here (did not add it to the interface yet, but I am not a big fan of catching it inside the EstimatedGasProvider).

So what do you think?  Should I 'undepricate' the "getGasPrice()" and deprecate the "getGasPrice(String contractFunc)" here?

iikirilov: Yes agreed. You can change the deprecated `getGasPrice`.

Regarding the exception handling, this should be handled by the caller. It is ok to remove the `try/catch` and declare that the method `throws` an exception

iikirilov:
@alex-forshtat-tbk bump to keep this open as the stale bot is hunting you down.

Looking forward to seeing this get merged one day.

alex-forshtat-tbk:
@iikirilov addressed some things we talked about and keeping it up-to-date with the master branch. Please hit me up once there is an integration test running or if you need any help with that so that this can be eventually merged.
P.S. As you can see I propose here to finally remove 'setGasPrice/Limit' from the Contract class, these two are deprecated and they forcefully replace passed Gas Provider with a Static Gas Provider.

iikirilov:
waiting on IT refactor, marking needs-review to stop going stale

lsh-silpion:
@forshtat @iikirilov any news on this one?

iikirilov:
@lsh-silpion I have not forgotten about this, however I cannot give you specific timeframe for when this will be integrated.

robertvescio1 (approved):

core/src/main/java/org/web3j/tx/gas/ContractGasProvider.java
georgmittendorfer: I think it is good that the funcName is in the method signature because:
* it doesn't hurt if you don't use it
* it is important for me as I adjust the gas price for some functions (some need faster handling, some can live with slower execution).

gtebrean:
In web3j the gas provider has been changed for several times and few month ago web3j started to support EIP-1559 which has the role to better estimate the gas price, due this I'm closing thins PR.

Issue #933, opened 2019-05-19
ReadOnlyTransactionManager can't make read only calls


pablovillot:
As of Web3j version 4.3.0, ReadOnlyTransactionManager can't make calls. Code was added to throw a UnsupportedOperationException("Not implemented") in sendCall

AlexandrouR:
PR is welcome.

PR #976, opened 2019-07-29
Implemented ReadonlyTransactionManager sendCall

AlexandrouR:
### What does this PR do?
Implemented sendCall method in the ReadonlyTransactionManager

ReadonlyTransactionManager class , sendCall method.

### Why is it needed?
Addresses issue #933

### DIFF
--- a/core/src/test/java/org/web3j/tx/ReadonlyTransactionManagerTest.java
+++ b/core/src/test/java/org/web3j/tx/ReadonlyTransactionManagerTest.java
@@ -13,12 +13,16 @@
 package org.web3j.tx;
 
 import java.io.IOException;
+import java.math.BigInteger;
+import java.util.Arrays;
 
 import org.junit.Test;
 
 import org.web3j.protocol.Web3j;
-import org.web3j.protocol.core.DefaultBlockParameterName;
+import org.web3j.protocol.Web3jService;
+import org.web3j.protocol.core.DefaultBlockParameter;
 import org.web3j.protocol.core.Request;
+import org.web3j.protocol.core.methods.request.Transaction;
 import org.web3j.protocol.core.methods.response.EthCall;
 
 import static org.hamcrest.CoreMatchers.is;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/web3j/tx/ReadonlyTransactionManagerTest.java
```diff
@@ -0,0 +1,47 @@
...
+import static org.junit.Assert.assertThat;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class ReadonlyTransactionManagerTest {
+
+    EthCall response = mock(EthCall.class);
+    Request request = mock(Request.class);
+    Web3j web3j = mock(Web3j.class);
```
josh-richardson: It's probably worth mentioning that everything here is mocked, so as far as I can tell nothing is actually being tested. That said this stuff is hard to test in an atomic way.

iikirilov: Agreed. I think the approach should mock the `Web3jService` instance passed to `Web3j.build(...)`

iikirilov: Something like:

```
Web3jService service = mock(Web3jService.class);
when(service.send()).thenReturn("test");
Web3j web3j = web3j.build(service);
```

Then you can create the ReadonlyTransactionManager with a concrete instance of web3j.

core/src/test/java/org/web3j/tx/ReadonlyTransactionManagerTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+public class ReadonlyTransactionManagerTest {
+
+    Web3jService service = mock(Web3jService.class);
+    Web3j web3j = Web3j.build(service);
+    Transaction transaction = mock(Transaction.class);
+    DefaultBlockParameter defaultBlockParameter = mock(DefaultBlockParameter.class);
+    EthCall response = mock(EthCall.class);
+
+    Request<?, EthCall> request =
```
iikirilov: Is this required? I don't see it being used in the test.

PR #987, opened 2019-08-11
Interactive ethCall and ethSendTransaction

mesudip:
### Inspiration
 - There's no way (other than using `ethSendTransaction`)  to get transactionHash right after transaction submission.
   Contract functions only return `transactionReceipt` which requires the transaction to appear in a block.
 - Thre's no way (other than using `ethCall`) to execute a view function with a blockParameter.
   Contract View functions have no way to pass blockParameter to the call.
 - Every contract function throws a plain `Exception` (You won't know what they are unless you dig deep into web3j code).

### Notes for reviewer
-  package `org.web3j.tx.interactions` contains the classes that will be returned for calls and transactions.
-  `org.web3j.tx.TransactionManager` and `org.web3j.tx.ManagedTransactions` have additional methods to send interactive calls and transactions.
- Changes on `Contract` class , codegen package and tests all revolve around the above change.
##### Interactions with `view` functions.
-  `erc20Contract.getBalance("0xabcd").send();` // like before
-  `erc20Contract.getBalance("0xabcd").atBlockHight(100).send()`

##### Interactions with non `view` functions.

-  `receipt = erc20Contract.transfer().send()`  is deprecated. came effect can be achieved using `erc20Contract.transfer(...).submit().waitForReceipt()`

- ```
    erc20Contract.transfer().send(new SendTransactionCallback{
    
        // callback after receiving transactionHash
        transactionHash(String transactionHash){
        }
        
        // callback aftter the tranasction appears in a block
        success(TransactionReceipt receipt,Optional<Response.Error> error){
        }
        
        // callback for failure
        exception(Exception ex){
        }
     });
     ```
     ```
  // example usage in webApps
   public function doTransaction() throws IOExcpetion{
       submitTransaction = erc20Contract.transfer("0",BigDecimal.ZERO).send();
       submitTransaction.sendAsync(new TransactionReceiptCallback{ //
    
       // we got the receipt we might also have an error.
       success(TransactionRecipt receipt, Optional<Response.Error> error){
    
       }
       // we didn't get receipt
       error(Exception ex){

      }})
     // return transaction hash to the client
     return submitTransaction.getTransacitonHash();
  }
  ```

##### Exceptions
Everybody hates a plain `java.lang.Exception`, so
- `contract.viewFunction().send()` or `contract.transactionFunction().send()` will throw only `IOException`
- when waiting for receipt, `contract.transact().send().send()` will throw `IOExcpeiton`, `TransactionException`, `FailedTransactionReceipt` or `InterruptedException`

### DIFF
--- a/besu/src/main/java/org/web3j/tx/PrivateTransactionManager.java
+++ b/besu/src/main/java/org/web3j/tx/PrivateTransactionManager.java
@@ -27,6 +27,7 @@ import org.web3j.protocol.core.methods.response.EthSendTransaction;
 import org.web3j.protocol.core.methods.response.TransactionReceipt;
 import org.web3j.protocol.eea.crypto.PrivateTransactionEncoder;
 import org.web3j.protocol.eea.crypto.RawPrivateTransaction;
+import org.web3j.protocol.exceptions.EthCallException;
 import org.web3j.protocol.exceptions.TransactionException;
 import org.web3j.tx.gas.BesuPrivacyGasProvider;
 import org.web3j.tx.response.PollingPrivateTransactionReceiptProcessor;
@@ -99,7 +100,7 @@ public abstract class PrivateTransactionManager extends TransactionManager {
     @Override
     protected TransactionReceipt executeTransaction(
             BigInteger gasPrice, BigInteger gasLimit, String to, String data, BigInteger value)
-            throws IOException, TransactionException {
+            throws IOException, TransactionException, EthCallException {
 
         EthSendTransaction ethSendTransaction =
                 sendTransaction(gasPrice, gasLimit, to, data, value);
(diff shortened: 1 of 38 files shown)


iikirilov:
Thank you for the submission.

1. I believe you can get the `transactionHash` immediately if you use the `NoOpProcessor.`
2. I agree that you cannot pass in a `DefaultBlockParameter` to call view functions from the contract wrapper. I think a simpler way to achieve this is to generate overloaded wrapper function for `view/pure` functions that accept a `DefaultBlockParameter`.
3. Agree about the exception handling

This PR addresses multiple things - It would be beneficial to focus PR's on one issue only. This makes it easier to review and give feedback so we can accept it. It would also be beneficial to open an issue first to discuss the approach.

I am strongly against modifying the `RemoteCall` interface. You can achieve the same results by wrapping the interface in a `ViewRemoteCall` or `TransactionRemoteCall` etc.

mesudip:
> I believe you can get the `transactionHash` immediately if you use the `NoOpProcessor`.
There's no way to do that when you are using generated contract methods like `erc20.transfer(...)`.
All you can do is `send()` or `sendAsync()`. Both of them will directly return a `TransactionReceipt`, not `transactionHash`.

> I am strongly against modifying the `RemoteCall` interface
currently `RemoteCall` is not an Interface, It's `abstract class` for some reason. 
It is justified to remove the callable field and make it an Interface. `new RemoteCall( new Callable(){ ... })` can always be replaced with `new RemoteCall(){ ...}` if we change it to an Interface. 

Interfaces are easy to implement compared to abstract classes.

This is not a big issue, It only makes extending RemoteCall weird. Instead of simply overriding `send()` one has to provide a `Callable` as constructor param.

iikirilov:
> There's no way to do that when you are using generated contract methods like erc20.transfer(...).
All you can do is send() or sendAsync(). Both of them will directly return a TransactionReceipt, not transactionHash

If you use the `NoOpProcessor` it will return a receipt immediately with only the transaction hash populated

r.e. RemoteCall being an interface - you are right I misread it.

I will need some time to do a more thorough review.

iikirilov:
In the meantime it would be good to fix the conflicts

iikirilov:
@mesudip do you plan on advancing this proposal? You will need to fix the conflicts...

mesudip:
@iikirilov can you start the review process?

PR #988, opened 2019-08-13
Prevent the ClassCastExeption for events that contain arrays

forshtat:
### What does this PR do?
There is a bug in the 'buildTypedResponse' method that leads to a generation of invalid contract code.
In case the event emits any array type, and 'useNativeJavaTypes' mode is selected, the generated code would look like the following:

In MyContract.sol :
```
event Myevent(uint8[] myparam);
```

In MyContract.java :
```
    public static List<MyeventEventResponse> staticGetMyeventEvents(TransactionReceipt transactionReceipt) {
...
        for (Contract.EventValuesWithLog eventValues : valueList) {
            MyeventEventResponse typedResponse = new MyeventEventResponse();
            typedResponse.myparam = **(List<BigInteger>)** eventValues.getNonIndexedValues().get(0).getValue();
            responses.add(typedResponse);
        }
        return responses;
    }
```
The issue with this is that 'getNonIndexedValues().get(0)' is a DynamicArray<> of solidity types, and it's 'getValue()' just returns an underlying array (this method is used to 'unwrap' the non-array types like `Type<Uint256>`).
On an attempt to read from such an incorrectly-cast array the exception is thrown.

Reproduce the bug and please tell me if it is bound to happen in other scenarios I did not think of. Also, please advise on how this functionality can be tested.

### Why is it needed?
Because bugs and crashes are generally not as good as no bugs and crashes :-) Also, emitting arrays is an important solidity feature for our contracts.

### DIFF
--- a/codegen/src/test/java/org/web3j/codegen/SolidityFunctionWrapperTest.java
+++ b/codegen/src/test/java/org/web3j/codegen/SolidityFunctionWrapperTest.java
@@ -508,6 +508,88 @@ public class SolidityFunctionWrapperTest extends TempFileProvider {
         assertThat(builder.build().toString(), is(expected));
     }
 
+
+    @Test
+    public void testBuildEventArrayParameterJavaTypesReturn() throws Exception {
+
+        AbiDefinition.NamedType paramUint8 = new AbiDefinition.NamedType("paramUint8", "uint8[]");
+        AbiDefinition.NamedType paramInt256 = new AbiDefinition.NamedType("paramInt256", "int256[]");
+        AbiDefinition.NamedType indexedArrayParam = new AbiDefinition.NamedType("indexedArrayParam", "uint8[]");
+        indexedArrayParam.setIndexed(true);
+
+        AbiDefinition functionDefinition =
+                new AbiDefinition(
+                        false,
+                        Arrays.asList(paramUint8, paramInt256, indexedArrayParam),
+                        "Transfer",
+                        new ArrayList<>(),
+                        "event",
+                        false);
+        TypeSpec.Builder builder = TypeSpec.classBuilder("testClass");
+
+        builder.addMethods(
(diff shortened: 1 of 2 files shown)


alex-forshtat-tbk:
@iikirilov Me too, and I will add some tests once I return to the office. But the issue for me is that the crash happens when executing the generated code, and tests that can check it are located in 'integration-tests' AFAIK. The tests that I will addin the meantime are the string comparison tests for the code generator, but I think that is not exactly the issue here.

alex-forshtat-tbk:
@iikirilov added a string comparison test for that. Integration tests still TBD.

iikirilov:
waiting on IT refactor

codegen/src/test/java/org/web3j/codegen/SolidityFunctionWrapperTest.java
```diff
@@ -508,6 +508,88 @@ public void testBuildEventConstantMultipleValueReturn() throws Exception {
...
+                "      }\n" +
+                "      typedResponse.paramInt256 = new ArrayList<>();\n" +
+                "      for (org.web3j.abi.datatypes.generated.Int256 element : (List<org.web3j.abi.datatypes.generated.Int256>) eventValues.getNonIndexedValues().get(1).getValue()) {\n" +
+                "        typedResponse.paramInt256.add(element.getValue());\n" +
+                "      }\n      responses.add(typedResponse);\n" +
+                "    }\n" +
+                "    return responses;\n" +
+                "  }\n" +
+                "\n" +
+                "  public io.reactivex.Flowable<TransferEventResponse> transferEventFlowable(org.web3j.protocol.core.methods.request.EthFilter filter) {\n" +
```
robertvescio1, line 555: 2nd filter?

robertvescio1 (approved):

gtebrean:
Issues fixed

PR #990, opened 2019-08-13
remove javafx exception

iikirilov:
### What does this PR do?
Replaces `javax.xml.bind.ValidationException` with `RuntimeException`

`StructuredDataEncoder.java`

### Why is it needed?
We don't want to depend on javafx


iikirilov:
@josh-richardson

PR #993, opened 2019-08-14
Release/4.5 aion

xaviarias:
### What does this PR do?
Implements required changes to support the Aion network and potentially other networks being integrated soon.

Review the changes from @xaviarias.

### Why is it needed?
Mainly because the current implementation does not support code generation subclasses and alternative encoding/decoding, making not possible the support for new networks such the Aion blockchain.


core/src/main/java/org/web3j/tx/RawTransactionManager.java
```diff
@@ -107,7 +107,12 @@ public void setTxHashVerifier(TxHashVerifier txHashVerifier) {
 
     @Override
     public EthSendTransaction sendTransaction(
-            BigInteger gasPrice, BigInteger gasLimit, String to, String data, BigInteger value)
+            BigInteger gasPrice,
+            BigInteger gasLimit,
+            String to,
+            String data,
+            BigInteger value,
+            boolean constructor)
```
iikirilov: this breaks backward compatibility

abi/src/main/java/org/web3j/abi/datatypes/primitive/PrimitiveType.java
```diff
@@ -0,0 +1,57 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package org.web3j.abi.datatypes.primitive;
+
+import java.util.Objects;
+
+import org.web3j.abi.datatypes.Type;
+
+public abstract class PrimitiveType<T> implements Type<T> {
```
iikirilov: Is there any sense in restricting `T` ?

abi/src/main/java/org/web3j/abi/DefaultFunctionEncoder.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.math.BigInteger;
+import java.util.List;
+
+import org.web3j.abi.datatypes.Function;
+import org.web3j.abi.datatypes.StaticArray;
+import org.web3j.abi.datatypes.Type;
+import org.web3j.abi.datatypes.Uint;
+
+public class DefaultFunctionEncoder extends FunctionEncoder {
```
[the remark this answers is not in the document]
xaviarias: I would leave it as Default since all web3j is Ethereum-based. Agree that would make sense if we had separate a separate module for Ethereum.

abi/src/main/java/org/web3j/abi/datatypes/IntType.java
```diff
@@ -17,19 +17,26 @@
...
+    public int getBitSize() {
+        return bitSize;
     }
 
     boolean valid(int bitSize, BigInteger value) {
         return isValidBitSize(bitSize) && isValidBitCount(bitSize, value);
     }
 
-    static boolean isValidBitSize(int bitSize) {
+    private static boolean isValidBitSize(int bitSize) {
```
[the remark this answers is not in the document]
xaviarias: I would leave them as static because they are simple predicates.

Changed instead the method `valid` to use fields instead of parameters.

core/src/main/java/org/web3j/tx/RawTransactionManager.java
xaviarias: Should we wait for release 5?

iikirilov: release 5 will happen before devcon - if it is ok to wait then I suggest we do

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -695,7 +722,13 @@ private String createMappedParameterTypes(ParameterSpec parameterSpec) {
                         + "))";
             }
         } else {
-            return "new " + parameterSpec.type + "(" + parameterSpec.name + ")";
+            String constructor = "new " + parameterSpec.type + "(";
+            if (Address.class.getCanonicalName().equals(parameterSpec.type.toString())
+                    && addressLength != Address.DEFAULT_LENGTH) {
```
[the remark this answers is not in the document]
iikirilov: agreed

PR #1003, opened 2019-08-20
run spotless in pipeline

### LINKED ISSUE #944: use web3j case The number of Threads has increased dramatically
When my program does not use web3j, the number of processes is stable. But when I use web3j requests, there will be a growing number of processes, resulting in insufficient memory in my program and Java heap space.

I use web3j like this demo:

image

I have use : web3j.shutdown()

iikirilov:
### What does this PR do?
runs spotlessCheck in travis pipeline

`build.gradle`

### Why is it needed?
code quality

fixes #944

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,9 @@ cache:
 
 # Don't run integrationTests as they require a live Ethereum client to be running
 # Add --info to view details of test failures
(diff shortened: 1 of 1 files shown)


AlexandrouR:
LGTM

PR #1011, opened 2019-08-25
adds functional websocket listener at WebSocketService

olimsaidov:
Backport of #787

What does this PR do?
added WebSocketListener at WebSocketService

WebSocketService, WebSocketServiceTest

Why is it needed?
I want to handle websocket events such as onClose event.
I think that it will be better if WebSocketClient have multiple listeners and WebSocketSerivce`s listener
is added.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
+++ b/core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
@@ -24,6 +24,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
```diff
@@ -114,22 +120,33 @@ private void connectToWebSocket() throws InterruptedException, ConnectException
         }
     }
 
-    private void setWebSocketListener() {
+    private void setWebSocketListener(Consumer<String> onMessage, Consumer<Throwable> onError,
+                                      Consumer<Void> onClose) throws ConnectException {
```
iikirilov: onClose can be a Runnable?

olimsaidov: > onClose can be a Runnable?

@zacscoding Can we use Runnable?

zacscoding: @olimasaidov
Sure :)
why requested pull again ?

olimsaidov: @zacscoding your pull request was merged into 5.x branch. We [need this functionality](https://github.com/web3j/web3j/pull/787#issuecomment-521595515) in 4.x versions.

core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
```diff
@@ -98,9 +99,14 @@ public WebSocketService(WebSocketClient webSocketClient, boolean includeRawRespo
      * @throws ConnectException thrown if failed to connect to the server via WebSocket protocol
      */
     public void connect() throws ConnectException {
+        connect(null, null, null);
```
iikirilov: I prefer to pass empty lambdas here and not check for null in the listener methods.

iikirilov:
@olimsaidov Please update to pass in empty lambdas and remove the null checks in the handlers so we can merge this :)

olimsaidov:
@iikirilov Thanks for your patience. Done ;)

Issue #902, opened 2019-03-26
ethCallResult.reverts is false if the call is reverted without any message


jimy74:
This boolean should still return true if the call execution been reverted without any message.

(I will implement it myself in Web3j if needed, anyone know how to conclude that the call been reverted?)

iikirilov:
Good spot. A PR is welcome

jimy74:
I still don't know how to conclude that a call to a contract has revert without message.

If the method should return something it's easy, because when it's revert without message it the return value is empty.

 But **if the method should not return anything, then good luck. 

How to detect a revert without message in this case?**

iikirilov:
@jimy74 maybe check the transaction receipt?

jimy74:
A "call" with web3 is made to be not mined at the end.
So there is nothing to do with the transaction receipt I guess unfortunately.

iikirilov:
@jimy74 ofcourse, my bad. This is tricky because on the client side there is no information to indicate whether a call has failed or not if it does not return any values. The node does have this information though - it is not transmitted back to the client. Maybe an EIP is required for this if you have a solid use-case...

AlexandrouR:
What is the use case for using `ethCall` on a function that does not return anything? My understanding is that `ethCall` is used to get contract state - i.e. calling a `view` function which returns a value of a variable in the contract. From this point of view if the function does not return a value then we can never know if the call reverted or not.

Maybe the resolution here is to remove the `reverts` function. What would be the implication of doing this?

kostind:
@AlexandrouR @iikirilov 
We can use ethCall for executing contract methods without transaction and getting the reason for the revert.
It can be done for an already failed transaction using input data.
Please take a look at PR https://github.com/web3j/web3j/pull/1022

oasis-zhou:
The function executeCall of Contract class gets the value which include the revert message,but the decode function ingores the message and returns "".
private List<Type> executeCall(Function function) throws IOException {
        String encodedFunction = FunctionEncoder.encode(function);
        EthCall ethCall = (EthCall)this.web3j.ethCall(Transaction.createEthCallTransaction(this.transactionManager.getFromAddress(), this.contractAddress, encodedFunction), this.defaultBlockParameter).send();
        String value = ethCall.getValue();
        return FunctionReturnDecoder.decode(value, function.getOutputParameters());
    }`

oasis-zhou:
[...]

PR #1022, opened 2019-09-04
Fix for EthCall reverts (#902)

kostind:
### What does this PR do?
Fix EthCall reverts and getRevertReason methods.

EthCall
EthCallIT

### Why is it needed?
If you need to execute your contract methods which update the state without transaction and get the reason for the transaction failure.


kostind:
@iikirilov Should I rebase my fork and recreate pull request in order to avoid commits like " Merge pull request #1 from web3j/master "?

iikirilov:
> @iikirilov Should I rebase my fork and recreate pull request in order to avoid commits like " Merge pull request #1 from web3j/master "?

yes

core/src/main/java/org/web3j/protocol/core/methods/response/EthCall.java
```diff
@@ -38,15 +38,21 @@ public String getValue() {
     }
 
     public boolean reverts() {
```
iikirilov: I prefer this method to be renamed to `isReverted`

kostind:
> LGTM - one suggestion for method name change.

Done, thank you.

iikirilov:
Yes, I prefer a rebase and force-push, but either way is fine as I will squash and merge.

I think this will clash with #1026. Is it possible to extend this for ethSendTransaction also?

kostind:
> Yes, I prefer a rebase and force-push, but either way is fine as I will squash and merge.
> 
> I think this will clash with #1026. Is it possible to extend this for ethSendTransaction also? Maybe I just need to merge both PRs?

I will do rebase before creating PR next time. Thanks for the clarification.

Looks like it will clash only with method `EthCall.reverts` which has been renamed to `isReverted`.
I think we have a few options.

1) Rename method back to `reverts`. Probably, it's a good option because all projects which depend on web3j's EthCall have to use `isReverted` instead of `reverts` after web3j update.
2) Add method `reverts` which simple use `isReverted` for backward compatibility.
3) Merge both PRs as is (#1026 first), fix errors (TransactionManager.assertCallNotReverted, ReadonlyTransactionManagerTest.sendCallRevertedTest)

I can help with either variant.

iikirilov:
ok cool. let's go with option 4. don't change anything in the PR :)

kostind:
Done.
I'm not sure how to fix the formatting issue.

iikirilov:
@kostind formatting fix with `./gradlew spotlessApply`

kostind:
@iikirilov Thanks.

Issue #1030, opened 2019-09-13
Release web3j-android


angerborn:
There's a branch for the 4.3.0-android release, but there's nothing under https://github.com/web3j/web3j/releases and nothing on maven central https://search.maven.org/search?q=g:org.web3j%20AND%20a:core&core=gav

Was there a problem creating the release, was it forgotten or was it buried below more important tasks? (Or am I just not looking in the right place?)

iikirilov:
Mainly buried - partly forgotten. I would like someone from the community to pick up this work - it requires merging all PRs labelled "merge-to-android" to an android branch.

It will mean we skip a few version on the android side.

Water-bamboo:
Would you please release 4.3-android?

om26er:
I see a new tag for Android, so I guess the Android version has finally progressed https://github.com/web3j/web3j/releases/tag/v4.6.0-android ?

AlexandrouR:
4.6.0 android version can be found [here](https://mvnrepository.com/artifact/org.web3j/core/4.6.0-android)

PR #901, opened 2019-03-26
Add ScheduledExecutorService to Parity constructor like in org/web3j/…

Alexey-N-Chernyshov:
…protocol/Web3j.java

### What does this PR do?
Add `ScheduledExecutorService` to `Parity` so now it is possible to build the class like a 'org.web3j.protocol.Web3j'.

### Where should the reviewer start?
Compare changes in `org/web3j/protocol/parity/Parity.java` with `org/web3j/protocol/Web3j.java`.

### Why is it needed?
To specify `ScheduledExecutorService` and be compatible with `org/web3j/protocol/Web3j.java`.

PR #1055, opened 2019-09-27
bytes, bytes1..32 fix; chainId is not restrictly required anymore (issues #1054, #1063)

avaness:
### What does this PR do?
it fixes https://github.com/web3j/web3j/issues/1054, https://github.com/web3j/web3j/issues/1063

### Where should the reviewer start?
- StructuredDataTest.java:155
- ValidStructuredDataWithBytesTypes.json - bytes and bytes3 datatypes

### Why is it needed?
StructuredDataTest.testBytestAndBytes32Encoding() fails on "bytes" and "bytes1..32" types

PR #1062, opened 2019-10-02
Updating android web3j with PRs added for 4.3.3 release 

MarkOSullivan94:
### What does this PR do?

Cherry picking two of the PRs which have to be merged in with Android branch

https://github.com/web3j/web3j/pull/901
https://github.com/web3j/web3j/pull/911

Not entirely sure if this is the approach which was desired in this issue here https://github.com/web3j/web3j/issues/1030 so thought I'd open with two cherry picked PRs first.

Next cherry picked PR I added caused some merge conflicts I guess this will do for now 😅 

Would be super useful if the PRs with the `merge-to-android` label all had info on what milestone / release it belonged to so it'd be easier to know what order they should be merged in

### Why is it needed?

To update android branch before next android release

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/StructuredData.java
+++ b/crypto/src/main/java/org/web3j/crypto/StructuredData.java
@@ -1,0 +1,111 @@
+package org.web3j.crypto;
+
+import java.util.HashMap;
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import org.web3j.abi.datatypes.Address;
+import org.web3j.abi.datatypes.generated.Uint256;
+
+public class StructuredData {
+    static class Entry {
+        private final String name;
+        private final String type;
+
+        @JsonCreator
+        public Entry(
+                @JsonProperty(value = "name") String name,
+                @JsonProperty(value = "type") String type) {
+            this.name = name;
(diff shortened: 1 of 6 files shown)


iikirilov:
@MarkOSullivan94 thank you for beginning this effort. The plan in my head right now is to merge all of the outstanding PRs and bring the android branch to feature parity with master before making a release (and skipping all the releases in the middle).

@antonydenyer Is it possible to auto-run CI for PRs going to android branch using java 6 - which is the current target. Having said that it might be a good time to question whether we want to upgrade the android branch to be java 8 and what the impact of that would be. Also it might make sense to spin the android stuff out into a separate repo?

MarkOSullivan94:
@iikirilov opened this with a couple done initially to make sure I was approaching this the correct way. 

Would it be possible to get this merged into one of the existing android branch first and then the rest of the PRs merged later before a new Android release?

antonydenyer:
> @antonydenyer Is it possible to auto-run CI for PRs going to android branch using java 6 - which is the current target. Having said that it might be a good time to question whether we want to upgrade the android branch to be java 8 and what the impact of that would be. Also it might make sense to spin the android stuff out into a separate repo?

Yes, you need to have a different https://github.com/web3j/web3j/blob/master/.travis.yml that does what you want.

iikirilov:
#911 will need to be followed by #1055 before we can merge this.

PR #1064, opened 2019-10-05
Adds admin_nodeInfo

josh-richardson:
### What does this PR do?
Adds admin_nodeInfo because Quorum broke some stuff.

My request test doesn't pass but my body at 3am doesn't want to understand why, it might also need a response test

### Why is it needed?
Quorum broke stuff.


core/src/main/java/org/web3j/protocol/core/methods/response/management/AdminNodeInfo.java
```diff
@@ -0,0 +1,58 @@
...
+    public static class ResponseDeserialiser extends JsonDeserializer<AdminNodeInfo.Result> {
+
+        @Override
+        public AdminNodeInfo.Result deserialize(
+                JsonParser jsonParser, DeserializationContext deserializationContext)
+                throws IOException {
+
+            TreeNode treeNode = jsonParser.readValueAsTree();
+            return new Result(treeNode.at("/name").toString());
+        }
```
josh-richardson: This probably isn't the right way of going about things, but it was the easiest way to get it going. Maybe I'm due a quick refresher on serialization with jackson?

core/src/main/java/org/web3j/protocol/core/JsonRpc2_0Web3j.java
```diff
@@ -140,6 +141,15 @@ public JsonRpc2_0Web3j(
                 "net_peerCount", Collections.<String>emptyList(), web3jService, NetPeerCount.class);
     }
 
+    @Override
+    public Request<?, AdminNodeInfo> adminNodeInfo() {
+        return new Request<>(
+                "admin_nodeInfo",
+                Collections.<String>emptyList(),
```
iikirilov: type of empty list can be inferred?

core/src/test/java/org/web3j/protocol/core/RequestTest.java
```diff
@@ -73,6 +73,13 @@ public void testNetPeerCount() throws Exception {
         verifyResult("{\"jsonrpc\":\"2.0\",\"method\":\"net_peerCount\",\"params\":[],\"id\":1}");
     }
 
+    @Test
+    public void testAdminNodeInfo() throws Exception {
+        web3j.adminNodeInfo().send();
+
+        verifyResult("{\"jsonrpc\":\"2.0\",\"method\":\"admin_NodeInfo\",\"params\":[],\"id\":1}");
```
iikirilov: is it supposed to be `admin_nodeInfo`? - it is probably why the test is failing :)

core/src/main/java/org/web3j/protocol/core/methods/response/management/AdminNodeInfo.java
iikirilov: looks ok - you are not handling `null` case though - not sure if it is something to consider

core/src/test/java/org/web3j/protocol/core/RequestTest.java
iikirilov:
```suggestion
        verifyResult("{\"jsonrpc\":\"2.0\",\"method\":\"admin_nodeInfo\",\"params\":[],\"id\":1}");
```

core/src/main/java/org/web3j/protocol/core/methods/response/management/AdminNodeInfo.java
josh-richardson: If it's null then the name field will just return null.

josh-richardson:
This now seems to be failing due to an issue with travis building master.

josh-richardson:
```Execution failed for task ':abi:uploadArchives'.

> Could not publish configuration 'archives'

   > Failed to deploy artifacts: Could not transfer artifact org.web3j:abi:jar:4.6.0-20191008.051455-59 from/to remote (https://oss.sonatype.org/content/repositories/snapshots/): Failed to transfer file: https://oss.sonatype.org/content/repositories/snapshots/org/web3j/abi/4.6.0-SNAPSHOT/abi-4.6.0-20191008.051455-59.jar. Return code is: 401, ReasonPhrase: Unauthorized.```

PR #1077, opened 2019-10-22
Update web3j to be compatible with new private transaction receipt

josh-richardson:
### What does this PR do?
Adds some fields to web3j private transaction receipts

### Why is it needed?
Besu will soon have a more extensive private transaction receipt.

### DIFF
--- a/besu/src/test/java/org/web3j/protocol/besu/ResponseTest.java
+++ b/besu/src/test/java/org/web3j/protocol/besu/ResponseTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Web3 Labs LTD.
+ * Copyright 2019 Web3 Labs Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
--- a/besu/src/test/java/org/web3j/protocol/besu/ResponseTest.java
+++ b/besu/src/test/java/org/web3j/protocol/besu/ResponseTest.java
@@ -280,6 +280,10 @@ public class ResponseTest extends ResponseTester {
                         + "        \"to\":\"0x85h43d8a49eeb85d32cf465507dd71d507100c1\",\n"
                         + "        \"output\":\"myRlpEncodedOutputFromPrivateContract\",\n"
                         + "        \"status\":\"0x1\",\n"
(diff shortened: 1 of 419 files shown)


josh-richardson:
@iikirilov This is ready for review.

besu/src/test/java/org/web3j/protocol/besu/ResponseTest.java
```diff
@@ -312,7 +313,9 @@ public void testPrivGetTransactionReceipt() {
                                         "0x0000000000000000000000000000000000000000000000000000000000000000",
                                         "mined",
                                         Collections.singletonList(
-                                                "0x59ebeb90bc63057b6515673c3ecf9438e5058bca0f92585014eced636878c9a5"))));
+                                                "0x59ebeb90bc63057b6515673c3ecf9438e5058bca0f92585014eced636878c9a5"))),
```
iikirilov: nit: extra line?

core/src/main/java/org/web3j/tx/Contract.java
```diff
@@ -372,9 +372,11 @@ TransactionReceipt executeTransaction(
                     String.format(
                             "Transaction has failed with status: %s. "
                                     + "Gas used: %d. (not-enough gas?)",
-                            receipt.getStatus(), receipt.getGasUsed()));
+                            receipt.getStatus(),
+                            receipt.getGasUsedRaw() != null
+                                    ? receipt.getGasUsed()
+                                    : BigInteger.valueOf(0)));
```
iikirilov: this should be implemented differently

besu/src/test/java/org/web3j/protocol/besu/ResponseTest.java
josh-richardson: Not that I can see

core/src/main/java/org/web3j/tx/Contract.java
iikirilov: can you do something like "Gas used: unknown" if it is not available

iikirilov: It would also be useful to log the revert reason here if it is available

josh-richardson: Updated this to return unknown gas used, showing a revert reason would introduce a cyclic dependency.

PR #1084, opened 2019-10-31
Updated Junit and mockito refactored the tests.

AlexandrouR:
### What does this PR do?
Updated junit from 4 to 5 .
Updated mockito and mockito-junit
All files
### Why is it needed?
*required*


codegen/src/test/java/org/web3j/codegen/ContractJsonParseTest.java
```diff
@@ -69,11 +69,11 @@ public void testParseConvertLib() throws Exception {
...
 
         assertEquals("Unexpected contract name", "ConvertLib", mc.getContractName());
-        assertEquals("Unexpected number of functions", 1, mc.abi.size());
+        assertEquals(1, mc.abi.size());
         AbiDefinition abi = mc.abi.get(0);
         assertEquals("Unexpected function name", "convert", abi.getName());
-        assertTrue("Expected function to be 'constant'", abi.isConstant());
-        assertFalse("Expected function to not be 'payable'", abi.isPayable());
+        assertTrue(abi.isConstant());
+        assertFalse(abi.isPayable());
```
josh-richardson: Does junit 5 not support reasons?

codegen/src/test/java/org/web3j/codegen/SolidityFunctionWrapperGeneratorTest.java
```diff
@@ -247,7 +249,7 @@ private void verifyGeneratedCode(String sourceFile) throws IOException {
             boolean result = task.call();
 
             System.out.println(diagnostics.getDiagnostics());
-            assertTrue("Generated contract contains compile time error", result);
+            assertTrue(result);
```
josh-richardson: Ditto here

core/src/test/java/org/web3j/protocol/websocket/WebSocketServiceTest.java
```diff
@@ -68,20 +67,18 @@
                     service,
                     Web3ClientVersion.class);
 
-    @Rule public ExpectedException thrown = ExpectedException.none();
```
josh-richardson: Guessing that this is just analogous to expect annotator from JUnit 4?

core/src/test/java/org/web3j/tx/ContractTest.java
```diff
@@ -168,45 +164,56 @@ public void testIsValidEmptyCode() throws Exception {
...
+                RuntimeException.class,
+                () -> {
+                    TransactionReceipt transactionReceipt = new TransactionReceipt();
+                    transactionReceipt.setTransactionHash(TRANSACTION_HASH);
+
+                    prepareTransaction(transactionReceipt);
+
+                    String encodedConstructor =
+                            FunctionEncoder.encodeConstructor(
+                                    Arrays.<Type>asList(new Uint256(BigInteger.TEN)));
```
josh-richardson: Not sure if it's considered good practice to wrap entire function in `assertThrows`, or just the line that we actually expect to have issues with?

integration-tests/src/test/java/org/web3j/protocol/scenarios/FastRawTransactionManagerIT.java
```diff
@@ -36,17 +33,15 @@
...
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.web3j.tx.TransactionManager.DEFAULT_POLLING_ATTEMPTS_PER_TX_HASH;
 
 public class FastRawTransactionManagerIT extends Scenario {
 
     private static final int COUNT = 10; // don't set too high if using a real Ethereum network
     private static final long POLLING_FREQUENCY = 15000;
 
-    @Rule public TestRule benchmarkRun = new BenchmarkRule();
```
josh-richardson: ?

codegen/src/test/java/org/web3j/codegen/ContractJsonParseTest.java
AlexandrouR: For whatever reason in JUnit 5 the second argument is the reason. Fixed in the latest commit I did.

integration-tests/src/test/java/org/web3j/protocol/scenarios/FastRawTransactionManagerIT.java
AlexandrouR: @BenchmarkOptions(concurrency = 2, warmupRounds = 0, benchmarkRounds = 5)
This would be the equivalent of that rule and can be used at the class level. Would like some opinions on the values that should be passed here.
[Docs about that here.](https://shantonusarker.blogspot.com/2017/01/junit-benchmark.html)

cfelde:
Besides the comments on FastRawTransactionManagerIT.java, on which I don't know enough to comment, it seems good.

integration-tests/src/test/java/org/web3j/protocol/scenarios/FastRawTransactionManagerIT.java
xaviarias: Why this benchmark is used for, concurrency issues or actual benchmarking? 

I think concurrency and warmups are ok, maybe we don't need 5 rounds to have an approximate idea of the test execution time. IMO 2-3 would suffice but I'm a total newbie in benchmarking.

core/src/test/java/org/web3j/tx/ContractTest.java
AlexandrouR: Fixed in the latest commit.

PR #1102, opened 2019-11-18
Add support for eth_chainId

magooster:
### What does this PR do?
Add support the getchainid JSON-RPC method, https://web3js.readthedocs.io/en/v1.2.3/web3-eth.html#getchainid which returns the chain ID of the current connected node as described in the EIP-695.

Simple change hopefully obvious...

### Why is it needed?
To enable support for EIP-555 simple replay attack protection currently have to configure the chainId externally, this would allow lookup from connected 'node'


magooster:
Missed a commit for the interface

PR #1107, opened 2019-11-20
Add multiple event subscribers in filters

zacscoding:
### What does this PR do?
Adds filter that subscribe multiple logs

BlocksFilter, LogsFilter, PendingTransactionsFilter in org.web3j.protocol.core.filter package.

### Why is it needed?
We want to handle multiple filter changes without additional requests.
For example, we want to know latest block hash after request with polling interval.


zacscoding:
@antonydenyer @iikirilov  thx 👍

PR #1110, opened 2019-11-25
Fix WebSocketService connect

zacscoding:
### What does this PR do?
- fixed WebSocketService:connect() method
- fixed broken test testSendSubscriptionReply() in WebSocketServiceTest
- fixed typo verifyStartedSubscriptionHadnshake to verifyStartedSubscriptionHandshake

WebSocketService, WebSocketServiceTest in core module

### Why is it needed?
If we connect again after connected at once, IllegalStateException occur because WebSocketClient's
thread is already exist. So have to use reconnect method to reset WebSocketClient


zacscoding:
@AlexandrouR Thx :)

PR #1118, opened 2019-12-03
Abstracts getCode to transactionmanager

josh-richardson:
### What does this PR do?
Adds priv_getCode method to web3j, abstracts getCode method away from the Contract class and into the TransactionManager so we can have multiple implementations depending on the transaction types we're dealing with.

*required*

### Why is it needed?
*required*


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Tue Dec 03 15:43:36 GMT 2019
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.3-all.zip
```
iikirilov: how come these have moved around?

iikirilov: also we want only the binary

PR #1120, opened 2019-12-05
Removed URLClassLoader usage.

AlexandrouR:
### What does this PR do?
Instead of using URLClassLoader that causes problems in Java 11, 
 `System.getProperty("java.class.path");` can be used to achieve the same thing.
*required*

### Why is it needed?
*required*

### DIFF
--- a/codegen/src/main/java/org/web3j/codegen/unit/gen/CompilerClassLoader.java
+++ b/codegen/src/main/java/org/web3j/codegen/unit/gen/CompilerClassLoader.java
@@ -15,7 +15,6 @@ package org.web3j.codegen.unit.gen;
 import java.io.File;
 import java.io.IOException;
 import java.net.URL;
-import java.net.URLClassLoader;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


codegen/src/main/java/org/web3j/codegen/unit/gen/CompilerClassLoader.java
```diff
@@ -93,8 +92,7 @@
     }
 
     private String buildClassPath() {
-        final URL[] systemUrls = ((URLClassLoader) getClass().getClassLoader()).getURLs();
-        return buildClassPath(urls) + ':' + buildClassPath(systemUrls);
+        return buildClassPath(urls) + ':' + System.getProperty("java.class.path");
```
cfelde: This might need testing on Windows? From docs:

Path used to find directories and JAR archives containing class files. Elements of the class path are separated by a platform-specific character specified in the path.separator property.

platform-specific character

AlexandrouR: ill test it on my windows machine now.

AlexandrouR: This works on windows

PR #1122, opened 2019-12-07
Generate EVMTest annotation based on the Java version.

AlexandrouR:
### What does this PR do?
When generating the unit tests if the java version is lower than 11 use EVMTest(type = NodeType.Geth)
otherwise, generate the default EVMTest annotation.

### Why is it needed?
To ensure that the tests run out of the box with Java 8 and 11.


cfelde:
LGTM

Puneetha17:
LGTM

PR #1129, opened 2019-12-19
Update admin RPC endpoints

antonydenyer:
### What does this PR do?

Expose more NodeInfo properties from admin_nodeInfo RPC endpoint
Wire up admin_peers RPC endpoint


core/src/main/java/org/web3j/protocol/core/methods/response/admin/AdminNodeInfo.java
```diff
@@ -10,48 +10,81 @@
...
-    public static class Result {
-        public String getName() {
-            return name;
-        }
-
+    public static class NodeInfo {
+        private String enode;
+        private String id;
+        private String ip;
+        private String lisstenAddr;
```
iikirilov:
```suggestion
        private String listenAddr;
```

core/src/main/java/org/web3j/protocol/core/methods/response/admin/AdminNodeInfo.java
```diff
@@ -10,48 +10,81 @@
...
         private String name;
 
-        Result(String name) {
+        public NodeInfo() {}
+
+        public NodeInfo(String enode, String id, String ip, String lisstenAddr, String name) {
+            this.enode = enode;
+            this.id = id;
+            this.ip = ip;
+            this.lisstenAddr = lisstenAddr;
```
iikirilov:
```suggestion
            this.listenAddr = listenAddr;
```

core/src/main/java/org/web3j/protocol/core/methods/response/admin/AdminNodeInfo.java
```diff
@@ -10,48 +10,81 @@
...
+
+        public String getId() {
+            return id;
+        }
+
+        public String getIp() {
+            return ip;
+        }
+
+        public String getLisstenAddr() {
```
iikirilov:
```suggestion
        public String getListenAddr() {
```

core/src/main/java/org/web3j/protocol/core/methods/response/admin/AdminNodeInfo.java
```diff
@@ -10,48 +10,81 @@
...
+            return lisstenAddr;
```
iikirilov:
```suggestion
            return listenAddr;
```

antonydenyer:
> also some fields are missing from the node info and peers

I'm a bit wary of adding too many things in, there are no standards around this yet so need to do some more analysis before adding more fields in.

PR #1134, opened 2019-12-31
Fix junit5 failing tests

lyotam:
### What does this PR do?
It seems that due to the switch to Junit5 the tests were not run and multiple of them have been failing.
This PR restores the tests to the build task and fixes the failing tests. 

build.gradle

### Why is it needed?
Because tests are ignored and failing.


AlexandrouR:
LGTM

PR #1137, opened 2020-01-03
Fix the getType method for int40 and uint40

jimy74:
### What does this PR do?
This is a bug fix.
It make the int40 a int40 and not a uint40 and the uint40 a uint40 and not a int40.

just check the file changed.

### Why is it needed?
getType method return the wrong type for int40 and uint40 because of a typo in some previous development.

This should be fast released as it fix a real bug in web3j that prevent users to update to the latest.

### DIFF
--- a/abi/src/main/java/org/web3j/abi/datatypes/AbiTypes.java
+++ b/abi/src/main/java/org/web3j/abi/datatypes/AbiTypes.java
@@ -187,9 +187,9 @@ public final class AbiTypes {
             case "int32":
                 return primitives ? Int.class : Int32.class;
             case "uint40":
-                return primitives ? Long.class : Int40.class;
-            case "int40":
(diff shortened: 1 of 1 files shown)


cfelde:
Maybe we need to additional unit tests for this?

jimy74:
@cfelde,+1 the coverage is definitely to low.

If anyone want to do it with non regression tests feel free do to so, there is only two lines changed for the moment, but it need to get merged I insist that it's critical, "who want to use web3j for financial solutions if it return such non sense values from smart contracts". The coverage was already an issue before and after the bug so it should be a separated ticket as it's a lot of work.

jimy74:
I say that because definitely the problem of missing tests exist for almost every solidity types, I don't know when/if I will have the time so the issue is still open if someone want to tackle that more quickly feel free, at least the bug is identified AND hopefully fixed soon.

PR #1140, opened 2020-01-13
Bumped okhttp version.

AlexandrouR:
Changed test to comply with new okhttp version.

### What does this PR do?
Bumps okhttp to the latest version

Changed Request tester

### Why is it needed?
Fixes a bug in java where after a request the thread does not stop.


cfelde:
LGTM

PR #1142, opened 2020-01-13
Added Tests in Type Decoder and Encoder for each Int type and Uint type.

AlexandrouR:
### What does this PR do?
Adds tests for each Int type and Uint type 

*required*

### Why is it needed?
*required*


cfelde:
LGTM

PR #1146, opened 2020-01-20
Adds add to & remove from privacy group

josh-richardson:
### What does this PR do?
Adds support for on-chain privacy groups in Besu. This is ready for an initial review, but still requires some changes in the Besu permissioning contracts for all tests to pass. Unit tests for certain methods are omitted due to method complexity not yet being accounted for in the test suite, however all methods are tested in integration tests.

*required*

### Why is it needed?
*required*


core/src/main/java/org/web3j/protocol/core/methods/response/TransactionReceipt.java
```diff
@@ -158,10 +158,10 @@ public void setStatus(String status) {
     }
 
     public boolean isStatusOK() {
-        if (null == status) {
+        if (null == getStatus()) {
```
iikirilov: this is also in another PR - either remove it from here or add a test for this

josh-richardson: Changed various things to get CI to pass in the last few mins, however unclear as to how we'd test this change specifically - just redirects `isStatusOK()` to potentially use overridden method if it exists in child classes. Not an ideal situation, but equally nicer than having to basically entirely restructure control flow & transaction class structure as we previously tried to to fix this.

iikirilov: you would have a mock and verify the method was correctly delegated

josh-richardson: Done

besu/src/main/java/org/web3j/protocol/besu/response/privacy/PrivacyGroup.java
```diff
@@ -76,8 +77,8 @@ public boolean equals(Object o) {
         }
         PrivacyGroup that = (PrivacyGroup) o;
         return getPrivacyGroupId().equals(that.getPrivacyGroupId())
-                && getName().equals(that.getName())
-                && getDescription().equals(that.getDescription())
+                && Objects.equals(getName(), that.getName())
```
iikirilov: any reason this changed?

josh-richardson: Description & getName are now nullable.

iikirilov:
LGTM

PR #1147, opened 2020-01-20
Fix private transaction receipt 1132

### LINKED ISSUE #1132: PrivateTransactionReceipt declares new fields instead of inheriting them
# PrivateTransactionReceipt declares new fields instead of inheriting them
PrivateTransactionReceipt declares again the following fields instead of inheriting them:
private final String contractAddress;
private final String from;
private final String to;
private final List logs;
private final String transactionHash;
private final String status;

This creates two copies of the same field for the superclass (TransactionReceipt) & subclass (PrivateTransactionReceipt), and creates various issues.

## Steps To Reproduce
Calling privateTransactionReceipt.isStatusOk()

### Expected behavior
Result should be evaluated by the actual receipt status

### Actual behavior
Result will always be true because superclass's status is always null, even though subclass status is not.

## Environment
Running Besu 1.31 Quickstart network.

lyotam:
### What does this PR do?
Fixes #1132 by inheriting the fields from superclass instead of creating a new copy of them in the subclass.

PrivateTransactionReceipt

### Why is it needed?
By creating two copies of the same field for the superclass (TransactionReceipt) & subclass (PrivateTransactionReceipt), this creates various issues - for example, Calling privateTransactionReceipt.isStatusOk() Result will always be true because superclass's status is always null, even though subclass status is not.

### DIFF
--- a/besu/src/main/java/org/web3j/protocol/besu/response/privacy/PrivateTransactionReceipt.java
+++ b/besu/src/main/java/org/web3j/protocol/besu/response/privacy/PrivateTransactionReceipt.java
@@ -24,86 +24,61 @@ import org.web3j.protocol.core.methods.response.TransactionReceipt;
 
 public class PrivateTransactionReceipt extends TransactionReceipt {
 
-    private final String contractAddress;
-    private final String from;
-    private final String to;
     private final String output;
-    private final List<Log> logs;
     private final String commitmentHash;
-    private final String transactionHash;
     private final String privateFrom;
     private final ArrayList<String> privateFor;
     private final String privacyGroupId;
-    private final String status;
-    private final String revertReason;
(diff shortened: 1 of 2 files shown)


iikirilov:
We will look at this after 1.4 Besu because some fields will change

yotamleshem:
> We will look at this after 1.4 Besu because some fields will change

Besu 1.4 Release doesn't seem to have any changes for PrivateTransactionReceiptResult.java. 
Did I miss something?

iikirilov:
@josh-richardson I think you covered this in you PR right? Specifically about the duplicate fields?

@yotamleshem yes RevertReason and Status fields were added to the private transaction receipt. This might not affect web3j because of composition.

PR #1165, opened 2020-02-28
Fix private transaction receipt status

josh-richardson:
### What does this PR do?
Makes `isStatusOk()` on private transaction receipts return the actual value, instead of always defaulting to true. Also fixes a bug whereby Web3j will throw if the block number on a transaction receipt is null.

*required*

### Why is it needed?
*required*

### DIFF
--- a/core/src/main/java/org/web3j/protocol/core/methods/response/TransactionReceipt.java
+++ b/core/src/main/java/org/web3j/protocol/core/methods/response/TransactionReceipt.java
@@ -158,10 +158,10 @@ public class TransactionReceipt {
     }
 
     public boolean isStatusOK() {
-        if (null == status) {
+        if (null == getStatus()) {
             return true;
         }
-        BigInteger statusQuantity = Numeric.decodeQuantity(status);
+        BigInteger statusQuantity = Numeric.decodeQuantity(getStatus());
         return BigInteger.ONE.equals(statusQuantity);
     }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/web3j/protocol/core/methods/response/TransactionReceipt.java
```diff
@@ -158,10 +158,10 @@ public void setStatus(String status) {
     }
 
     public boolean isStatusOK() {
-        if (null == status) {
+        if (null == getStatus()) {
```
iikirilov: needs a test

core/src/main/java/org/web3j/utils/RevertReasonExtractor.java
```diff
@@ -65,7 +65,7 @@ public static String extractRevertReason(
     public static String retrieveRevertReason(
             TransactionReceipt transactionReceipt, String data, Web3j web3j) throws IOException {
 
-        if (transactionReceipt.getBlockNumber() == null) {
+        if (transactionReceipt.getBlockNumberRaw() == null) {
```
iikirilov: needs a test

core/src/main/java/org/web3j/protocol/core/methods/response/TransactionReceipt.java
iikirilov: this was fixed in another PR

iikirilov:
@josh-richardson status on this?

iikirilov:
closing due to inactivity

PR #1170, opened 2020-03-04
Online services implementation

AlexandrouR:
### What does this PR do?
Adds web3j local account implementation 

all modified files 

### Why is it needed?
to integrate with upcoming web3j auth

### DIFF
--- a/core/src/main/java/org/web3j/account/LocalWeb3jAccount.java
+++ b/core/src/main/java/org/web3j/account/LocalWeb3jAccount.java
@@ -14,7 +14,7 @@ public class LocalWeb3jAccount {
     private static final Path web3jConfigPath =
             Paths.get(System.getProperty("user.home"), ".web3j", ".config");
 
-    static String SERVICES_ENDPOINT = "http://localhost/api/geth/%s/%s/";
+    static String SERVICES_ENDPOINT = "http://localhost/api/rpc/%s/%s/";
 
     public static HttpService getOnlineServicesHttpService(final Network network) throws Exception {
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/web3j/account/LocalWeb3jAccount.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.web3j.protocol.Network;
+import org.web3j.protocol.http.HttpService;
+
+public class LocalWeb3jAccount {
+
+    private static final Path WEB3J_CONFIG_PATH =
+            Paths.get(System.getProperty("user.home"), ".web3j", ".config");
+
+    private static String SERVICES_ENDPOINT = "https://%s-eth.epirus.io/%s";
```
antonydenyer: Node RPC endpoint

core/src/main/java/org/web3j/account/LocalWeb3jAccount.java
```diff
@@ -0,0 +1,66 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import org.web3j.protocol.Network;
+import org.web3j.protocol.http.HttpService;
+
+public class LocalWeb3jAccount {
+
+    private static final Path WEB3J_CONFIG_PATH =
+            Paths.get(System.getProperty("user.home"), ".web3j", ".config");
```
antonydenyer: epirus config

PR #1174, opened 2020-03-13
Handle decoding values that are long values

antonydenyer:


utils/src/main/java/org/web3j/utils/Numeric.java
```diff
@@ -39,6 +40,10 @@ public static String encodeQuantity(BigInteger value) {
     }
 
     public static BigInteger decodeQuantity(String value) {
+        if (isLongValue(value)) {
+            return BigInteger.valueOf(Long.parseLong(value));
+        }
```
josh-richardson: The only thing that worries me here is that decodeQuantity basically now decodes longs as well as hex values - as long as we're mindful of this in the future and don't rely on decodeQuantity being able to validate an input as a hex input, I think we're okay here, but it does smell a bit.

antonydenyer: Agreed, it's bitten us a couple of times when client implementations have returned numbers instead of hex.

cfelde: it also makes the behavior closer to what people are used to from web3js I believe..

iikirilov:
needs a spotless

PR #1183, opened 2020-04-12
[Bugfix] NPE in JsonRpc2_0Rx

tau3:
### What does this PR do?
Fixes an NPE (verified by a unit test) 

`JsonRpc2_0RxTest#testReplayBlocksFlowableWhenIOExceptionOnBlockResolving`

### Why is it needed?
fixes bug, simplifies code, improves readability

### DIFF
--- a/core/src/main/java/org/web3j/protocol/rx/JsonRpc2_0Rx.java
+++ b/core/src/main/java/org/web3j/protocol/rx/JsonRpc2_0Rx.java
@@ -12,11 +12,18 @@
  */
 package org.web3j.protocol.rx;
 
+import java.io.IOException;
+import java.math.BigInteger;
+import java.util.List;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.stream.Collectors;
+
 import io.reactivex.BackpressureStrategy;
 import io.reactivex.Flowable;
 import io.reactivex.FlowableEmitter;
 import io.reactivex.Scheduler;
 import io.reactivex.schedulers.Schedulers;
+
 import org.web3j.protocol.Web3j;
 import org.web3j.protocol.core.DefaultBlockParameter;
 import org.web3j.protocol.core.DefaultBlockParameterName;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/web3j/protocol/rx/JsonRpc2_0Rx.java
```diff
@@ -138,35 +139,21 @@ public JsonRpc2_0Rx(Web3j web3j, ScheduledExecutorService scheduledExecutorServi
     private Flowable<EthBlock> replayBlocksFlowableSync(
             DefaultBlockParameter startBlock,
             DefaultBlockParameter endBlock,
-            boolean fullTransactionObjects,
-            boolean ascending) {
-
-        BigInteger startBlockNumber = null;
-        BigInteger endBlockNumber = null;
+            boolean isFullObjects,
```
iikirilov, line 142: i don't like this name change. maybe `containsFullTransactionObjects` ?

`isFullObjects` is too abstract here

core/src/main/java/org/web3j/protocol/rx/JsonRpc2_0Rx.java
```diff
@@ -138,35 +139,21 @@ public JsonRpc2_0Rx(Web3j web3j, ScheduledExecutorService scheduledExecutorServi
...
-        } else {
-            return Flowables.range(startBlockNumber, endBlockNumber, false)
-                    .flatMap(
-                            i ->
-                                    web3j.ethGetBlockByNumber(
-                                                    new DefaultBlockParameterNumber(i),
-                                                    fullTransactionObjects)
-                                            .flowable());
-        }
+        return Flowables.range(startBlockNumber, endBlockNumber, isAscending)
```
iikirilov, line 153: v. nice :)

tau3: thank you! 😄

core/src/main/java/org/web3j/protocol/rx/JsonRpc2_0Rx.java
tau3: yes, i agree)

PR #809, opened 2018-11-29
Pull request for #769

serso:
What does this PR do?
Makes web3j compatible with Android.
The project is built against the reduced JDK to ensure that it works fine on Android.
Fixes #769.

Where should the reviewer start?
First commit. I tried to make commits atomic and simple.

Why is it needed?
To make web3j available on Android.

PR #1188, opened 2020-04-16
4.5.17 android

Alex009:
### What does this PR do?
it's update for android version of library to current main version (4.5.17)

Please select correct target branch - i start from `release/4.2.1-android`, but maybe it not correct

### Why is it needed?
On android we can't use main java version because it use api introduced in API 24 and API 28 levels of android. With this request we can use it start from API 21.


iikirilov:
Awesome!! thanks for this. Can you confirm the current API version of web3j-android:4.2.0 - I believe we target API 16 but I am not sure.

I think we want to aim for a lower API version to support more devices. Does this even make sense? I mean is it worth it? I do not really have a preference...

N.B. I do not work with android on a daily basis so I am not up to date with what is considered acceptable...

Alex009:
if we consider the entire play market, then the current active devices have the following API levels
Снимок экрана 2020-04-15 в 19 49 42
i think library will support API 16 now too, but i not check it. i check on API 21, 24, 28

iikirilov:
@Alex009 you are legend! I am happy with API 21 :)

Considering this is a massive change - I will go through this in more detail over the weekend. The main consideration is to make sure we did not miss a feature.

Thank you for the effort so far!

Some more questions I have:

1. around the use of sopngeycastle/bouncycastle in android. Is the current set-up (whatever it is) good to go? i.e. could you test wallet generation on android. 

2. around the streams support I know we have some dependencies in the android branch so we can compile with java 6. Is this still necessary? Can we compile with java 8 and still be compliant with API 21?

iikirilov:
also you need to run `./gradlew spotlessApply` to fix code style

README.rst
```diff
@@ -145,7 +160,11 @@ Java 8:
 
 .. code-block:: groovy
 
+<<<<<<< HEAD
```
conor10, line 163: We shouldn't have diffs here

Alex009: sorry, bad merge was :( it fixed now

Alex009:
>around the use of sopngeycastle/bouncycastle in android. Is the current set-up (whatever it is) good to go? i.e. could you test wallet generation on android.

we use generation of mnemonics and generation of wallet from mnemonics for etherium wallets - it's work as we see.

>around the streams support I know we have some dependencies in the android branch so we can compile with java 6. Is this still necessary? Can we compile with java 8 and still be compliant with API 21?

- streams available on android only from API 24 - https://developer.android.com/reference/java/util/stream/Stream
- optional too API 24 - https://developer.android.com/reference/java/util/Optional
- java nio files API 26 - https://developer.android.com/reference/java/nio/file/Files

so it necessary for now. when API 24 will be more used (90% and more) of market - dependencies can be removed and min API level set to API 24

Alex009:
i fix issues of merge, now request is ready for review

Alex009:
now i merge `master-android` into this branch. some points:

module `eea` should be completed (i can't implement it self):
- org.web3j.tx.PrivateTransactionManager#getCode
- org.web3j.tx.PrivateTransactionManager#sendTransaction
- org.web3j.tx.PrivateTransactionManager#sendCall

also `org.web3j.account.LocalWeb3jAccount` updated to version without `java.nio` usage (API 26 in android)

iikirilov:
ok thanks, I will look tomorrow

iikirilov:
@serso any comments you might have on this?

I will build locally and test with https://github.com/serso/web3a 

We aim to stick to API 21 as the minimum required.

conor10:

https://community.web3labs.com/t/web3j-android-4-5-17-call-for-review/60/1

serso:
> @serso any comments you might have on this?
I think it would have been much easier to review the PR if the branch was based on master and then only fixes needed on Android applied at the top (basically what I did here https://github.com/web3j/web3j/pull/809).

iikirilov:
noted - I think that makes more sense too because this PR is missing significant code

iikirilov:
replaced by #1213

PR #1195, opened 2020-04-21
Added Besu privacy getLogs and filters methods

lucassaldanha:
### What does this PR do?
Added support for private logs/filters methods in Besu.
• priv_getLogs
• priv_newFilter
• priv_uninstallFilter
• priv_getFilterLogs
• priv_getFilterChanges

All of these methods work in the exact same way as their "eth" equivalent. The biggest difference is that they all have the privacyGroupId as request parameter index 0.

• Start by checking the that the method names are correct (using the prefix `priv` instead of `eth`)
• The order of the parameters is important. Parameter 0 for all requests is the `privacyGroupId`.

### Why is it needed?
To support people using web3j + Besu to consume logs.

### DIFF
--- a/besu/src/test/java/org/web3j/protocol/besu/RequestTest.java
+++ b/besu/src/test/java/org/web3j/protocol/besu/RequestTest.java
@@ -22,11 +22,13 @@ import org.junit.jupiter.api.Test;
 import org.web3j.protocol.RequestTester;
 import org.web3j.protocol.core.DefaultBlockParameter;
 import org.web3j.protocol.core.DefaultBlockParameterName;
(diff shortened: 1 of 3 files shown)


besu/src/test/java/org/web3j/protocol/besu/RequestTest.java
```diff
@@ -214,7 +216,7 @@ public void testPrivGetCode() throws Exception {
     }
 
     @Test
-    public void testEthCall() throws Exception {
+    public void testPrivCall() throws Exception {
```
lucassaldanha, line 219: I'm renaming this method because I believe it had the wrong name :)

PR #1197, opened 2020-04-22
Align type of constructor parameter chainId with RawTransactionManager

magooster:
### What does this PR do?
Align the type of FastRawTransactionManager constructor parameter chainId with RawTransactionManager

See RawTransactionManger constructor

### Why is it needed?
So can pass chainId > 127 to the constructor


magooster:
> please can we keep backward compatibility for this.
> 
> It will require that the old constructors remain - you can mark them `@Deprecated`
> 
> You can put new constructors that take the `chinId` as a `long`

FastRawTransactionManager extends RawTransactionManager in which the parameter type for chainId Id is defined as a long -so I see the current constructor definitions in FastRawTransactionManager to be bugs...

Changing the type to a long is fully backward compatible through automatic widening conversion.

PR #1206, opened 2020-05-05
feature/abiv2

iikirilov:
### What does this PR do?
Introduces support for generating contract wrappers with ABI v2 features

see https://solidity.readthedocs.io/en/v0.5.10/abi-spec.html#handling-tuple-types

SolidityFunctionWrapperGenerator
FunctionEncoder
FunctionDecoder

### Why is it needed?
Allows generating contract wrappers for contracts that specify `pragma experimental ABIEncoderV2;`

N.B. 
- does not support work for contracts complied with solc < 0.6
- does not support multiple structs in function call response
- does not support struct arrays


iikirilov:
> I tried to build the code but this test didn't pass. It said: _type == null_.

Yes - because of the way structs are ordered for code generation. This has since been fixed.

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -330,15 +340,176 @@ private String buildEventDefinitionName(String eventName) {
...
+                                                        + buildTypeName(namedType.getType(), false)
+                                                        + "("
+                                                : "")
+                                        + (namedType.getType().equals("tuple")
+                                                ? namedType.getName()
+                                                : useNativeJavaTypes
+                                                        ? namedType.getName() + ")"
+                                                        : namedType.getName());
+                            } catch (ClassNotFoundException e) {
+                                throw new RuntimeException();
```
josh-richardson, line 490: Maybe add message here

abi/src/main/java/org/web3j/abi/TypeDecoder.java
```diff
@@ -345,6 +445,157 @@ static Utf8String decodeUtf8String(String input, int offset) {
...
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private static <T extends Type> int getDynamicStructDynamicParametersCount(
+            final Class<?>[] cls) {
+        int dynamicParametersCount = 0;
+        for (final Class<?> cl : cls) {
+            if (isDynamic((Class<T>) cl)) {
+                dynamicParametersCount++;
```
josh-richardson, line 560: Might look nicer as a stream/filter.

abi/src/main/java/org/web3j/abi/TypeDecoder.java
```diff
@@ -354,7 +605,10 @@ static BigInteger asBigInteger(Object arg) {
             return Numeric.toBigInt((String) arg);
         } else if (arg instanceof byte[]) {
             return Numeric.toBigInt((byte[]) arg);
-        } else if (arg instanceof Double || arg instanceof Float) {
+        } else if (arg instanceof Double
+                || arg instanceof Float
+                || arg instanceof java.lang.Double
+                || arg instanceof java.lang.Float) {
```
josh-richardson, line 611: Is this meant to be checking both primitive and complex types? If so should the first two not be lower case?

RobertoArtiles:
There seems to be a problem with tuples https://github.com/web3j/web3j/issues/1239

RobertoArtiles:
Was there any reason to omit struct arrays support?

Ferparishuertas:
Is there any plan to support struct arrays support as @RobertoArtiles mentioned?

xaviarias:
At the moment there is no plan in the roadmap to support arrays, but feel free to create an issue with the requirements or a pull-request and we'll happily review it.

esaulpaugh:
esaulpaugh/headlong can do encoding/decoding for struct (tuple) arrays

xaviarias (commented):
> esaulpaugh/headlong can do encoding/decoding for struct (tuple) arrays

I remember this was discussed previously with @iikirilov and there was a reason it was not used, but maybe we can reopen the conversation.

esaulpaugh:
@xaviarias If I recall, the reasons were that adding a new dependency conflicted with the goal of debloating web3j and that abiv2 was a minority use-case. At the time I also suggested that web3j fork headlong and maintain just the features they needed, which was probably more trouble than it was worth, considering that headlong was still in very active development.

I should note that headlong is much more mature now and has essentially entered maintenance mode as of a few weeks ago.

xaviarias:
Sounds good, I guess is something to consider.

@iikirilov @conor10 any thoughts on this?

conor10:
I'd rather we have proper support in Web3j. Including Headlong as a dependency would create a lot of duplicate functionality in terms of how one can interact with smart contract code via Web3j, which could be confusing for your users. If we were to fork Headlong I would presume we'd have to remap a lot of native Headlong types into Web3j types which again would make me question why don't we simply do the work on our side.

Issue #1228, opened 2020-06-20
Add support for privDistributeRawTransaction


magooster:
## Feature description

Add support for Besu API Method [privDistributeRawTransaction](https://besu.hyperledger.org/en/stable/Reference/API-Methods/#priv_distributerawtransaction)

Private transaction processing involves two transactions, the private transaction and the privacy marker transaction. The private transaction and the privacy marker transaction each have their own nonce.

If your private transaction rate requires sending private transactions without waiting for the previous private transaction to be mined, you must use priv_distributeRawTransaction instead of eea_sendRawTransaction. 

When using priv_distributeRawTransaction you create and send the privacy marker transaction yourself rather than eea_sendRawTransaction handling the privacy marker transaction.

PR #1229, opened 2020-06-20
Distribute raw tx

magooster:
### What does this PR do?
Adds support for the privDistributeRawTransaction Besu API Method

Implements #1228

PrivateTransactionManager.java sendAndDistribute method

### Why is it needed?

Private transaction processing involves two transactions, the private transaction and the privacy marker transaction. The private transaction and the privacy marker transaction each have their own nonce.

If your private transaction rate requires sending private transactions without waiting for the previous private transaction to be mined, you must use priv_distributeRawTransaction instead of eea_sendRawTransaction. 

Note: With this PR when using sendAndDistribute method users still need to create and send the privacy marker transaction themselves.


besu/src/test/java/org/web3j/protocol/besu/RequestTest.java
```diff
@@ -12,10 +12,7 @@
  */
 package org.web3j.protocol.besu;
 
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
iikirilov, line 15: please undo the wildcard

besu/src/test/java/org/web3j/protocol/besu/ResponseTest.java
```diff
@@ -23,15 +23,7 @@
...
-import org.web3j.protocol.besu.response.privacy.PrivCreatePrivacyGroup;
-import org.web3j.protocol.besu.response.privacy.PrivFindPrivacyGroup;
-import org.web3j.protocol.besu.response.privacy.PrivGetPrivacyPrecompileAddress;
-import org.web3j.protocol.besu.response.privacy.PrivGetPrivateTransaction;
-import org.web3j.protocol.besu.response.privacy.PrivGetTransactionReceipt;
-import org.web3j.protocol.besu.response.privacy.PrivacyGroup;
-import org.web3j.protocol.besu.response.privacy.PrivateTransactionLegacy;
-import org.web3j.protocol.besu.response.privacy.PrivateTransactionReceipt;
-import org.web3j.protocol.besu.response.privacy.PrivateTransactionWithPrivacyGroup;
+import org.web3j.protocol.besu.response.privacy.*;
```
iikirilov, line 26: wildcard

PR #1231, opened 2020-06-22
Numeric changes

camembertelectrique:
### What does this PR do?
Speed up byte array to hex string conversion

Method toHexString() reimplemented

### Why is it needed?
Two orders of magnitude faster implementation

### DIFF
--- a/utils/src/main/java/org/web3j/utils/Numeric.java
+++ b/utils/src/main/java/org/web3j/utils/Numeric.java
@@ -234,10 +234,10 @@ public final class Numeric {
 
     public static char[] toHexCharArray(byte[] input, int offset, int length, boolean withPrefix) {
         final char[] output = new char[length << 1];
-        for (int i = offset; i < length; i++) {
+        for (int i = offset, j = 0; i < length; i++, j++) {
             final int v = input[i] & 0xFF;
-            output[i * 2] = HEX_CHAR_MAP[v >>> 4];
-            output[i * 2 + 1] = HEX_CHAR_MAP[v & 0x0F];
+            output[j++] = HEX_CHAR_MAP[v >>> 4];
+            output[j] = HEX_CHAR_MAP[v & 0x0F];
         }
         return output;
     }
--- a/utils/src/main/java/org/web3j/utils/Numeric.java
+++ b/utils/src/main/java/org/web3j/utils/Numeric.java
@@ -232,7 +232,7 @@ public final class Numeric {
         return withPrefix ? new StringBuilder(HEX_PREFIX).append(output).toString() : output;
     }
 
-    public static char[] toHexCharArray(byte[] input, int offset, int length, boolean withPrefix) {
+    private static char[] toHexCharArray(byte[] input, int offset, int length, boolean withPrefix) {
         final char[] output = new char[length << 1];
         for (int i = offset, j = 0; i < length; i++, j++) {
             final int v = input[i] & 0xFF;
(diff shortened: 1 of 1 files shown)


conor10:
Thanks for taking the time to submit this.

Whilst I appreciate it is a speedup, is this sort of optimisation really necessary at present in Web3j given the main constraint we have is the EVM? 

It would be helpful to understand what was the underlying motivation that led you to this specific change, as it sacrifices readability for performance.

camembertelectrique:
Conor, we have an app that generates non-EVM-based raw transactions to a precompiled contract that have large payloads compared to an EVM transaction, and we use Numeric.toHexString() to encode them. During testing, we found that this call created a significant slowdown in our app. This patch resulted in a two orders of magnitude reduction in compute time, and solved the performance issue we were having.

utils/src/main/java/org/web3j/utils/Numeric.java
```diff
@@ -227,15 +228,18 @@ private static String toHexStringZeroPadded(BigInteger value, int size, boolean
...
-        }
-        for (int i = offset; i < offset + length; i++) {
-            stringBuilder.append(String.format("%02x", input[i] & 0xFF));
-        }
+        final String output = new String(toHexCharArray(input, offset, length, withPrefix));
+        return withPrefix ? new StringBuilder(HEX_PREFIX).append(output).toString() : output;
+    }
 
-        return stringBuilder.toString();
+    public static char[] toHexCharArray(byte[] input, int offset, int length, boolean withPrefix) {
```
conor10, line 235: why is `withPrefix` still here?

camembertelectrique: It's to preserve the functionality of toHexString and to avoid an interface change.

conor10: but `toHexCharArray` is only defined in this class and could be private

camembertelectrique: Yes, you're right. `toHexCharArray` could be made private to preserve the interface as it is.

conor10: If you make this update I'm happy to merge.

esaulpaugh:
you can also eliminate one array access per byte with a size-256 encoding table:

```java
final int[] ints = new int[] { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f' };
final int leftNibbleMask = 0xF0;
final int rightNibbleMask = 0x0F;
for (int i = 0; i < ENCODE_TABLE.length; i++) {
    int leftChar = ints[(i & leftNibbleMask) >>> Byte.SIZE / 2];
    int rightChar = ints[i & rightNibbleMask];
    ENCODE_TABLE[i] = (leftChar << Byte.SIZE) | rightChar;
}
```

```java
int hexPair = ENCODE_TABLE[input[offset] & 0xFF];
output[j] = (char) (hexPair >>> Byte.SIZE); // left
output[j+1] = (char) (hexPair & 0xFF); // right
```

resulting in a decent speedup https://github.com/esaulpaugh/headlong/blob/master/src/main/java/com/esaulpaugh/headlong/util/FastHex.java

And if you're optimizing for Java 9+ it's faster to encode to bytes and then use `new String(byte[],int,int,int)` because of CompactStrings.

utils/src/main/java/org/web3j/utils/Numeric.java
camembertelectrique: I reviewed and tested the suggestion for performance improvement from @esaulpaugh but found none. I have nothing else to add to this pull, it's good to merge if you approve

conor10: what about the removal of `withPrefix`?

Issue #499, opened 2018-04-14
Constant string too long


samczsun:
Tried to load a really big contract, and the generated contract wouldn't compile:

```
Error:(47, 42) java: constant string too long
```

eepstein:
we need a bit more context.

samczsun:
My apologies.

I can't find the codegen code responsible for this, but I'm referring to the following generated line (not necessarily in this particular contract, but in general):
https://github.com/web3j/web3j/blob/1c4215ff0e9957cfd28e5618b19d7ccdd18cdc1e/integration-tests/src/test/java/org/web3j/generated/Fibonacci.java#L36

The contract I loaded was so large that the binary string was longer than the maximum constant string length.

relonger:
Here is our workaround, run it after class generated.

It replaces one long string to 
```java
    private static final String BINARY =  new StringBuilder()
.append("0x60806040523480156200001157600080fd5b5060036005620000696040805190810160405280600581526020017f61646d696e0000000000000000")
.append("00000000000000000000000000000000000000815250838360016200021b640100000000026401000000009004565b50620000ba3360408051908101")

...

.append("619509565b5090565b905600a165627a7a7230582082b23acccbb08b266e935fbebbf9a062a1873fe9f1dcdd76a8dae8b79f4378180029")
.append("").toString();

```

```bash
#!/usr/bin/env bash
set -e

shopt -s globstar

for file in src/**/*.java; do
    if [ -f $file ]; then
        echo "Processing $file"

        mv "$file" "$file.bk"
        gawk '{ if( match($0, /^(.*?String BINARY = )"([^"]+)"/, m))
            {
                print m[1], "new StringBuilder()"
                n = split (m[2], ary, /.{1,120}/, seps);
                for (i=1; i<=n; i++)
                   printf ".append(\"%s\")%s", seps[i], (i<n ? "\n" : ".toString();\n")
           } else {
                print $0
           }
        }' "$file.bk" > "$file"
        rm "$file.bk"
    fi
done
```

tentenponce:
Is there any proper fix for this?

PR #1241, opened 2020-07-29
split binary string literal longer than 65534 as it is not supported by java complier

amalrajmani:
### What does this PR do?
This PR enhances `SolidityFunctionWrapper.createBinaryDefinition` method. It splits the `BINARY` string by length 65534 into an array and initialises BINARY string with `StringBuilder` that is made up of this array.

function `SolidityFunctionWrapper.createBinaryDefinition` 

### Why is it needed?
The `BINARY` variable(from java) generated by web3j's codegen does not handle when the `BINARY`'s length is more than 65534. This result's in java program generated not able to compile due to the limitation in java ( does not support string literal size more than 65534). It solves issue #499


p-s-dev (commented):
this would be very useful

PR #1254, opened 2020-08-25
Fix for usage of Uint and Int ABI types in EIP712

JamesSmartCell:
### What does this PR do?
Fixes usage of Int and Uint types in EIP712 Structured data

See the added test, ValidStructuredDataWithValues.json. The test includes a structure ```GasData``` which consists of uint256 types. The constructors for the Uint and Int ABI types only accept ```long``` and ```BigInteger```, however the values are encoded as ```String```, which means the constructor encoder at line 329 fails as there's no matching constructor. 

### Why is it needed?
Valid EIP712 structures fail to be encoded, for example this construct:

```
        "GasData": [
            {
                "name": "gasLimit",
                "type": "uint256"
            }
        ],
```
```
"message": {
        ...
        "gasData": {
            "gasLimit": "39507"
        },
```
due to incompatible constructor params. An alternative fix would be to add a ```String``` constructor to the Int and Uint ABI types which attempts to convert a submitted ```String``` into a BigInteger, however the submitted fix is more local and doesn't affect code elsewhere.

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
+++ b/crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
@@ -16,6 +16,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.math.BigInteger;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


JamesSmartCell:
This code is required for correct EIP712 operation. Currently the implementation is broken. Please see the test case.

PR #1260, opened 2020-09-10
add salt to EIP712 domain

davyvanroy:
### What does this PR do?
Adds salt to the EIP712Domain. According to the [EIP](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md), a salt can be added to the domain: 
> an disambiguating salt for the protocol. This can be used as a domain separator of last resort

I double checked the results with what I got back from an [EIP712 demo ](https://weijiekoh.github.io/eip712-signing-demo/index.html)together with metamask. I also verified it in remix.

4 changes:
- Added the salt in StructuredData - EIP712Domain
- Made sure the salt is not present when it is null (StructuredDataEncoder)
- Added a valid JSON test file with salt
- Created a new test to validate the result

### Why is it needed?
If you use a valid EIP712 json with a salt, you get a deserialisation exception (unknown field). 
Since this is part of the EIP it should be implemented.

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/StructuredData.java
+++ b/crypto/src/main/java/org/web3j/crypto/StructuredData.java
@@ -48,17 +48,20 @@ public class StructuredData {
         private final String version;
         private final Uint256 chainId;
         private final Address verifyingContract;
+        private final String salt;
 
         @JsonCreator
         public EIP712Domain(
                 @JsonProperty(value = "name") String name,
                 @JsonProperty(value = "version") String version,
                 @JsonProperty(value = "chainId") Uint256 chainId,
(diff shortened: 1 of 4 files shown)


thor-ragnarok:
LGTM - although someone needs to decide something about backwards compatibility something

davyvanroy:
> LGTM - although someone needs to decide something about backwards compatibility something

I checked, inside @JsonProperty the required is by default false:

>     boolean required() default false;

All tests inside "StructuredDataTest.java" still pass, which still has examples without salt and they are still green 👍  
A test was added (and example json file) that tests the structured data with salt. I did not touch the ones without salt.

conor10:

https://community.web3labs.com/t/web3j-v4-8-0-released/112/1

PR #1265, opened 2020-09-15
Fixes folder path containing space

SweeXordious:
### What does this PR do?
Fixes folder path containing spaces in CompilerClassLoader
CompilerClassLoader class

### Why is it needed?
To compile classes containing spaces in their path in CompilerClassLoader


AlexandrouR:
LGTM

PR #1273, opened 2020-09-30
build(dependencies) upgrade dependencies

mmaryo:
an upgrade to dependencies


josh-richardson:
Please re-open this with a coherent explanation as to why you think it would be beneficial & follow the project template.

PR #1284, opened 2020-10-27
Converted Readme to MD format

AlexandrouR:
Updated README.md

### What does this PR do?
Converted README.rts to README.MD

Updated Readme according to specifications.

Readme file
### Why is it needed?
*required*


README.MD
```diff
@@ -0,0 +1,157 @@
+Web3j: Web3 Java Ethereum Ðapp API
+==================================
+
+[![Documentation Status](https://img.shields.io/travis/web3j/web3j-docs?label=docs)](https://docs.web3j.io/)
+
+[![Build Status](https://travis-ci.org/web3j/web3j.svg?branch=master)](https://travis-ci.org/web3j/web3j)
+
+[![codecov](https://codecov.io/gh/web3j/web3j/branch/master/graph/badge.svg)](https://codecov.io/gh/web3j/web3j)
+
+[![Join the chat at https://gitter.im/web3j/web3j](https://img.shields.io/discourse/users?server=https%3A%2F%2Fcommunity.web3labs.com)](https://community.web3labs.com)
```
SweeXordious, line 10: I think it would be better if you don't skip a line after each icon... they would look better horizontally

README.MD
```diff
@@ -0,0 +1,157 @@
...
+-   [Jnr-unixsocket](https://github.com/jnr/jnr-unixsocket) for \*nix
+    IPC (not available on Android)
+
+It also uses [JavaPoet](https://github.com/square/javapoet) for
+generating smart contract wrappers.
+
+QuickStart
+---------
+- Use Epirus CLI (recommended for full project creation)
+    - Install Epirus CLI (include Windows installation instructions too)
```
SweeXordious, line 79: include windows instructions

README.MD
```diff
@@ -0,0 +1,157 @@
...
+    - Run ```$ epirus new```
```
SweeXordious, line 80: Maybe would look better if you do something like: 

The simplest way to start your journey with Web3j is to create a project.
We provide this functionality using the Epirus-cli (with link). This latter can be installed as follows:
```
comand to install epirus CLI
```
For Windows, in Powershell:
```
these blocks of instructions better be in a separate line to be more clear where the commands are
```
Then, say that you can create a project using:
```
$ epirus new
```
Then, Run app etc...
I think it would be better if we use simple sentences instead of -

README.MD
```diff
@@ -0,0 +1,157 @@
...
+    - Run app:
+        - Create [Epirus account](https://docs.epirus.io/sdk/cli/#account-creation)
+        - Provide parameters to run project locally
+        
+- Use [Maven](https://github.com/web3j/web3j-maven-plugin) / [Gradle](https://github.com/web3j/web3j-gradle-plugin) plugin
+
+Documentation
+-------------
+Please head to the [Web3j Documentation](https://docs.web3j.io) for further instructions on using Web3j.
```
SweeXordious, line 89: Why not put this sentence like this:

### Please head to the [Web3j Documentation](https://docs.web3j.io) for further instructions on using Web3j.

without specifying that `Documentation` title.
Titles in md files can be done using hashtags... one hashtag is biggest title, two hashtags is smaller one etc. I think it goes down to 6 levels

README.MD
```diff
@@ -0,0 +1,157 @@
...
+-----
+
+Java:
+
+``` 
+<dependency>
+  <groupId>org.web3j</groupId>
+  <artifactId>core</artifactId>
+  <version>4.7.0</version>
+</dependency>
```
SweeXordious, line 101: why not use latest versions ?

README.MD
```diff
@@ -0,0 +1,157 @@
...
+```
+
+Android:
+
+``` 
+<dependency>
+  <groupId>org.web3j</groupId>
+  <artifactId>core</artifactId>
+  <version>4.6.0-android</version>
+</dependency>
```
SweeXordious, line 111: version too. Not sure about the android releases tho...

README.MD
```diff
@@ -0,0 +1,157 @@
...
+
+``` 
+compile ('org.web3j:core:4.7.0')
+```
+
+Android:
+
+``` 
+compile ('org.web3j:core:4.6.0-android')
+```
```
SweeXordious, line 127: maybe bump the version and also use the new `implementation 'org.web3j:core:4.8.1'` instead of the old `compile`

SweeXordious (changes requested):

Issue #1238, opened 2020-07-17
Facing java.security.NoSuchAlgorithmException: no such algorithm: ECDSA for provider BC for wallet creation and minifyEnabled is true


W3-Mimo:
# Facing java.security.NoSuchAlgorithmException: no such algorithm: ECDSA for provider BC for wallet creation and minifyEnabled is true.

## _Issue_description_
My application has a wallet creation feature So that I am using the following code

`WalletUtils.generateLightNewWalletFile("<Password>", destinationFile);`

This line working fine in debug mode but when I am preparing release build with `minifyEnabled ` is true then I am facing this exception
`java.security.NoSuchAlgorithmException: no such algorithm: ECDSA for provider BC`

I am using `setupBouncyCastle()` in Application class

```
private void setupBouncyCastle() {
        final Provider provider = Security.getProvider(BouncyCastleProvider.PROVIDER_NAME);
        if (provider == null) {
            // Web3j will set up the provider lazily when it's first used.
            return;
        }
        if (provider.getClass().equals(BouncyCastleProvider.class)) {
            // BC with same package name, shouldn't happen in real life.
            return;
        }
        // Android registers its own BC provider. As it might be outdated and might not include
        // all needed ciphers, we substitute it with a known BC bundled in the app.
        // Android's BC has its package rewritten to "com.android.org.bouncycastle" and because
        // of that it's possible to have another BC implementation loaded in VM.
        Security.removeProvider(BouncyCastleProvider.PROVIDER_NAME);
        Security.insertProviderAt(new BouncyCastleProvider(), 1);
    }
```

I am using **`implementation 'org.web3j:core:4.6.0'`** version of web3J

And I am adding proguard rule in

```
-dontwarn java8.util.**
-dontwarn jnr.posix.**
-dontwarn com.kenai.**

#-keep class org.bouncycastle.**
-dontwarn org.bouncycastle.jce.provider.X509LDAPCertStoreSpi
-dontwarn org.bouncycastle.x509.util.LDAPStoreHelper

-keepclassmembers class org.web3j.protocol.** { *; }
-keepclassmembers class org.web3j.crypto.* { *; }

-keep class * extends org.web3j.abi.TypeReference
-keep class * extends org.web3j.abi.datatypes.Type

#-dontwarn java.lang.SafeVarargs
-dontwarn org.slf4j.**
```

And I am using both proguard file (those are given below) for resolving this issue but it didn't

1.
```
minifyEnabled true
[...]

Issue #915, opened 2019-04-13
No such algorithm: ECDSA for provider BC


pohonduit:
Hi, i'm trying to create/generate wallet on Android P, but found this issue,
any idea what should i do?

`Caused by: java.security.NoSuchAlgorithmException: no such algorithm: ECDSA for provider BC
        at sun.security.jca.GetInstance.getService(GetInstance.java:87)
        at sun.security.jca.GetInstance.getInstance(GetInstance.java:206)
        at java.security.KeyPairGenerator.getInstance(KeyPairGenerator.java:304)
        at org.web3j.crypto.Keys.createSecp256k1KeyPair(Keys.java:58)
        at org.web3j.crypto.Keys.createEcKeyPair(Keys.java:75)
        at org.web3j.crypto.Keys.createEcKeyPair(Keys.java:70)
        at org.web3j.crypto.WalletUtils.generateNewWalletFile(WalletUtils.java:61)
        at org.web3j.crypto.WalletUtils.generateFullNewWalletFile(WalletUtils.java:40)`

Dawnton:
Try this : Add the code below in your custom Application

` @Override
    public void onCreate() {
        super.onCreate();
        setupBouncyCastle();
    }

    private void setupBouncyCastle() {
        final Provider provider = Security.getProvider(BouncyCastleProvider.PROVIDER_NAME);
        if (provider == null) {
            // Web3j will set up the provider lazily when it's first used.
            return;
        }
        if (provider.getClass().equals(BouncyCastleProvider.class)) {
            // BC with same package name, shouldn't happen in real life.
            return;
        }
        // Android registers its own BC provider. As it might be outdated and might not include
        // all needed ciphers, we substitute it with a known BC bundled in the app.
        // Android's BC has its package rewritten to "com.android.org.bouncycastle" and because
        // of that it's possible to have another BC implementation loaded in VM.
        Security.removeProvider(BouncyCastleProvider.PROVIDER_NAME);
        Security.insertProviderAt(new BouncyCastleProvider(), 1);
    }`

iikirilov:
@pohonduit are you using web3j-android?

kartik-karz:
So, I tried using that and after using the setup function the application is crashing with 
```
Caused by: java.lang.OutOfMemoryError: Failed to allocate a 1040 byte allocation with 1000 free bytes and 1000B until OOM, max allowed footprint 268435456, growth limit 268435456
[...]

Issue #828, opened 2018-12-27
ECDSA implementation not found in 4.1.0-android


WarSame:
I am getting an exception on Android when trying to generate a new light wallet using 4.1.0-android. This is code that previously used to work:

`this.fileName = WalletUtils.generateLightNewWalletFile(longPassword, getWalletDir(context));`

longPassword is a String, getWalletDir returns a directory to store all of the wallet files in.

```
W/System.err: java.security.NoSuchAlgorithmException: KeyPairGenerator ECDSA implementation not found
        at org.apache.harmony.security.fortress.Engine.notFound(Engine.java:190)
W/System.err:     at org.apache.harmony.security.fortress.Engine.getInstance(Engine.java:183)
W/System.err:     at java.security.KeyPairGenerator.getInstance(KeyPairGenerator.java:149)
W/System.err:     at java.security.KeyPairGenerator.getInstance(KeyPairGenerator.java:122)
        at org.web3j.crypto.Keys.createSecp256k1KeyPair(Keys.java:58)
        at org.web3j.crypto.Keys.createEcKeyPair(Keys.java:75)
        at org.web3j.crypto.Keys.createEcKeyPair(Keys.java:70)
        at org.web3j.crypto.WalletUtils.generateNewWalletFile(WalletUtils.java:61)
W/System.err:     at org.web3j.crypto.WalletUtils.generateLightNewWalletFile(WalletUtils.java:47)
        at com.example.graeme.beamitup.wallet.Wallet.<init>(Wallet.java:43)
W/System.err:     at com.example.graeme.beamitup.wallet.Wallet.<init>(Wallet.java:22)
        at com.example.graeme.beamitup.wallet.Wallet$WalletBuilder.build(Wallet.java:161)
        at com.example.graeme.beamitup.wallet.GenerateWalletTask.handleWalletCreation(GenerateWalletTask.java:74)
        at com.example.graeme.beamitup.wallet.GenerateWalletTask.doInBackground(GenerateWalletTask.java:51)
        at com.example.graeme.beamitup.wallet.GenerateWalletTask.doInBackground(GenerateWalletTask.java:18)
W/System.err:     at android.os.AsyncTask$2.call(AsyncTask.java:295)
        at java.util.concurrent.FutureTask.run(FutureTask.java:237)
        at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:234)
[...]

PR #1289, opened 2020-11-03
Generate secp256k1 keypair on Android without modifying system security providers

ydanneg:
### What does this PR do?
This PR makes available to generate secp256k1 keypair on Android devices without the need of removing native android BC provider and inserting external BC provider, that can break many functionality of Android native components.

There is only one place in web3j that modifies system properties by injecting BC provider if it not registered yet.
crypto/src/main/java/org/web3j/crypto/Keys.java

### Why is it needed?
Android has always registered native BC provider and it's implementation doesn't support ECDSA key generation: createSecp256k1KeyPair methods will always throw NoSuchAlgorithmException exception.
This PR removes unnecessary registration of BC provider in java security (aka changing System settings).
Instead BouncyCastle provider instance is used to get KeyPairGenerator instance from included external BouncyCastle lib.
Before this change Android developers have to remove existing native BC provider and insert real one that support ECDSA key generation. Now it's not required at all.

I've tested it on Android 11.
Before change it throws an exception.
After change it successfully generates the keypair.

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/Keys.java
+++ b/crypto/src/main/java/org/web3j/crypto/Keys.java
@@ -20,7 +20,6 @@ import java.security.NoSuchAlgorithmException;
 import java.security.NoSuchProviderException;
 import java.security.Provider;
 import java.security.SecureRandom;
-import java.security.Security;
 import java.security.spec.ECGenParameterSpec;
 import java.util.Arrays;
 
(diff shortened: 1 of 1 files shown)


ydanneg:
is this project alive? No Android users here?

aeharvlee:
@ydanneg 
Hi ydanneg. I'm an android user and I'm still getting same errors like
* #1238
* #915
* #828 
 
I think your PR will fix these issues 
but why this PR has not been reviewed by maintainers of web3 j? 
🤔

PR #1292, opened 2020-11-10
Add blockHash filter to EthFilter due to EIP-234

JhChoy:
### What does this PR do?
1. Add blockHash filter to EthFilter due to [EIP-234](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-234.md).
2. Fix test bug that actual and expected value were switched.

1. EthFilter => RequestTest
2. RequestTester

### Why is it needed?
1. Support to query logs with blockHash which is described at `eth_getLogs` part of [ethereum json wiki](https://eth.wiki/json-rpc/API)
2. Test logs were confused.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/core/methods/request/EthFilter.java
+++ b/core/src/main/java/org/web3j/protocol/core/methods/request/EthFilter.java
@@ -12,7 +12,7 @@
  */
 package org.web3j.protocol.core.methods.request;
 
-import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 import org.web3j.protocol.core.DefaultBlockParameter;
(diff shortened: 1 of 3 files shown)


junbeomlee:
LGTM!

core/src/main/java/org/web3j/protocol/core/methods/request/EthFilter.java
```diff
@@ -40,7 +41,17 @@ public EthFilter(
 
     public EthFilter(
             DefaultBlockParameter fromBlock, DefaultBlockParameter toBlock, String address) {
-        this(fromBlock, toBlock, Arrays.asList(address));
+        this(fromBlock, toBlock, Collections.singletonList(address));
+    }
+
+    public EthFilter(String blockHash) {
+        super();
```
xaviarias, line 48: I believe this `super();` call is redundant.

xaviarias (approved):
Great, thanks!

PR #1299, opened 2020-11-27
Unable to parse input ABI file fix

AlexandrouR:
### What does this PR do?
Fixes https://github.com/web3j/solidity-gradle-plugin/issues/32
*required*

### Why is it needed?
*required*


SweeXordious (approved):

xaviarias (commented):
Maybe we should print a line saying "Ignoring empty ABI file: <filename>.abi"

xaviarias (approved):

PR #1314, opened 2020-12-21
Restore ClassProvider and CompilerClassLoader.

AlexandrouR:
### What does this PR do?
Restores ClassProvider and CompilerClassLoader
Added files
### Why is it needed?
They are used in unit test generation in web3j-cli


SweeXordious (approved):
LGTM

codegen/src/main/java/org/web3j/codegen/unit/gen/ClassProvider.java
```diff
@@ -0,0 +1,74 @@
...
+                                throw new RuntimeException(e);
+                            }
+                        })
+                .collect(Collectors.toList());
+    }
+
+    private List<String> getFormattedClassPath() throws IOException {
+        Stream<Path> walk = Files.walk(Paths.get(pathToJavaFiles.toURI()));
+        return getClassPathFromURL(
+                walk.filter(Files::isRegularFile).map(Path::toString).collect(Collectors.toList()));
```
xaviarias, line 63: This code is not very clear. You are creating a stream, collection into a list, then calling `getClassPathFromURL` that maps the resulting list into anther one.

When using streams, they should be executed in one expression like:
```
list.stream()
    .map(::mapFunction1)
    ...
    .map(::mapFunctionN)
    .collect(...)
```
Splitting into smaller functions is useful when it actually provides some advantage. In this case the code becomes less readable and less efficient, so there is no point in separating into functions just for the sake of it.

It looks more tidy but it actually is more confusing.

codegen/src/main/java/org/web3j/codegen/unit/gen/ClassProvider.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private List<String> getClassPathFromURL(final List<String> listOfUrl) throws IOException {
+        int length = pathToJavaFiles.getCanonicalPath().length();
+        List<String> formattedClassPath = new ArrayList<>();
```
xaviarias, line 68: This is a code smell in the lines 68 to 72 that we already discussed in the past:

```
List listY = new ArrayList()
for (X elem : listX) {
    listY.add(elem)
}
return listY;
```

This is a mapping loop that should be done in a stream.

xaviarias (changes requested):
This was deleted because unused.

Why it is added again if is not used anywhere?

Please provide a reason why we are going back to code that was refactored and cleared up.

AlexandrouR:
@xaviarias  This is used in epirus-cli and after the cli refactoring will be used in web3j-cli. The reason why the effects of this change was not noticed by the cli is because it is using web3j 4.7.0

xaviarias:
> @xaviarias This is used in epirus-cli and after the cli refactoring will be used in web3j-cli. The reason why the effects of this change was not noticed by the cli is because it is using web3j 4.7.0

Ok still the code needs a good refactor because ClassProvider is not a reusable component and what it does can be much more simplified and clean.

I will open a PR after being merged.

PR #1321, opened 2020-12-31
Structs arrays support

SweeXordious:
### What does this PR do?
Adds support for:
- Arrays of structs
- Multiple structs returns

Abi encoder/decoder and check the encoding formal specification

### Why is it needed?
More and more users are using structs in their Solidity smart contracts


thor-ragnarok:
test coverage looks good - needs a spotless apply

SweeXordious:
Still needs more encoding tests with different functions and also codegen module update to generate arrays of structs.

SweeXordious:
[Limitation](https://github.com/web3j/web3j/issues/1329)

abi/src/main/java/org/web3j/abi/DefaultFunctionEncoder.java
```diff
@@ -64,10 +68,21 @@ private static String encodeParameters(
...
+    @SuppressWarnings("unchecked")
     private static int getLength(final List<Type> parameters) {
         int count = 0;
         for (final Type type : parameters) {
-            if (type instanceof StaticArray) {
+            if (type instanceof StaticArray
+                    && StaticStruct.class.isAssignableFrom(
+                            ((StaticArray) type).getComponentType())) {
+                count +=
+                        staticStructCanonicalFieldsCount(((StaticArray) type).getComponentType())
```
xaviarias, line 79: Is a good idea to make utility methods for general use. 

In this case having a count is very specific, the method `staticStructCanonicalFieldsCount` should return the list of fields. 

Also the name is not clear, the only difference with `staticStructsNestedFieldsFlatList` is that the fields are public.  

So maybe call it something like `staticStructsNestedPublicFieldsFlatList`? To make clear that is the same method - just a suggestion.

abi/src/main/java/org/web3j/abi/TypeDecoder.java
```diff
@@ -640,24 +651,55 @@ static List arrayToList(Object array) {
 
         try {
             Class<T> cls = Utils.getParameterizedTypeFromArray(typeReference);
-            if (Array.class.isAssignableFrom(cls)) {
-                throw new UnsupportedOperationException(
-                        "Arrays of arrays are not currently supported for external functions, see"
-                                + "http://solidity.readthedocs.io/en/develop/types.html#members");
-            } else {
+            if (StructType.class.isAssignableFrom(cls)) {
                 List<T> elements = new ArrayList<>(length);
```
xaviarias, line 655: Could this if-else branching be splitted into private functions? It makes the function too long to read.

abi/src/main/java/org/web3j/abi/Utils.java
```diff
@@ -152,4 +161,43 @@ static String getSimpleTypeName(Class<?> type) {
         }
         return result;
     }
+
+    /**
+     * Returns number of canonical fields in a static struct. Example: struct Baz { Struct Bar { int
+     * a, int b }, int c } will return three fields count.
+     *
+     * @param classType
```
xaviarias, line 169: View comment above.

Also, please remove all empty Javadoc or even better, fill it if possible. 

Eg `@param classType`, `@return` are meaningless.

xaviarias (approved):

874341642:
> ### What does this PR do?
> Adds support for:
------------------------------------------------------
abi
----------------------------------------------------------------
"outputs": [{
		"components": [ {
			"internalType": "uint256[]",
			"name": "blocks",
			"type": "uint256[]"
		}],
		"internalType": "struct Mining.PlayerWork",
		"name": "",
		"type": "tuple"
	}]
----------------------------------------------------------------
java
----------------------------------------------------------------
  public PlayerWork(DynamicArray<Uint256> blocks) {
           super(blocks); 
       }
----------------------------------------------------------------
Error info
----------------------------------------------------------------
java.lang.UnsupportedOperationException: Array types must be wrapped in a TypeReference
----------------------------------------------------------------
Bug?Bug?Bug?Bug?Bug?Bug?Bug?Bug?

SweeXordious:
@874341642 Please open a new issue with this case, and provide more information on solidity version, compiler used etc.

874341642:
The tuple contains arrays, what do I need to do to solve this problem?

> @874341642 Please open a new issue with this case, and provide more information on solidity version, compiler used etc. I personally don't think I will have time to pick this up in upcoming months. But, if somebody else did, I would be happy to provide context and review.

The tuple contains arrays, what do I need to do to solve this problem?

874341642:
> @874341642 Please open a new issue with this case, and provide more information on solidity version, compiler used etc. I personally don't think I will have time to pick this up in upcoming months. But, if somebody else did, I would be happy to provide context and review.

Please help me

tonghoang196:
when is this included in the new version?

PR #1325, opened 2020-12-31
Fix leak in HttpService

kare:
### What does this PR do?
- Fixes a `okhttp3.Response` leak in `HttpService`
- Fixes deprecated `okhttp3` API calls

See commit 3b5d027

### Why is it needed?
Response objects should be closed.


AlexandrouR:
LGTM

PR #1341, opened 2021-01-14
Update Bouncy Castle 1.68

### LINKED ISSUE #1336: Please update Bouncy Castle
Please update Bouncy Castle
- https://www.bouncycastle.org/
- https://www.bouncycastle.org/releasenotes.html
- http://www.bouncycastle.org/latest_releases.html
- https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=bouncy%20castle
- https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=bouncycastle
- https://www.cvedetails.com/vulnerability-list/vendor_id-7637/Bouncycastle.html

Here:
- https://github.com/web3j/web3j/blob/master/build.gradle#L14

kare:
### What does this PR do?
Fixes #1336

By reading the attached Bouncy Castle release notes.

### Why is it needed?
Security related libraries should be kept up to date.

## Bouncy Castle Release: 1.68
Date:      2020, December 21st.

### Defects Fixed

- Some BigIntegers utility methods would fail for BigInteger.ZERO. This has been fixed.
- PGPUtil.isKeyRing() was not detecting secret sub-keys in its input. This has been fixed.
- The ASN.1 class, ArchiveTimeStamp was insisting on a value for the optional reducedHashTree field. This has been fixed.
- BCJSSE: Lock against multiple writers - a possible synchronization issue has been removed.

### Additional Features and Functionality

- BCJSSE: Added support for system property com.sun.net.ssl.requireCloseNotify. Note that we are using a default value of 'true'.
- BCJSSE: 'TLSv1.3' is now a supported protocol for both client and server. For this release it is only enabled by default for the 'TLSv1.3' SSLContext, but can be explicitly enabled using 'setEnabledProtocols' on an SSLSocket or SSLEngine, or via SSLParameters.
- BCJSSE: Session resumption is now also supported for servers in TLS 1.2 and earlier. For this release it is disabled by default, and can be enabled by setting the boolean system property org.bouncycastle.jsse.server.enableSessionResumption to 'true'.
- The provider RSA-PSS signature names that follow the JCA naming convention.
- FIPS mode for the BCJSSE now enforces namedCurves for any presented certificates.
- PGPSignatureSubpacketGenerator now supports editing of a pre-existing sub-packet list.

## Bouncy Castle Release: 1.67
Date:      2020, November 1st.

### Defects Fixed

- BCJSSE: SunJSSE compatibility fix - override of getChannel() removed and 'urgent data' behaviour should now conform to what the SunJSSE expects.
- Nested BER data could sometimes cause issues in octet strings. This has been fixed.
- Certificates/CRLs with short signatures could cause an exception in toString() in the BC X509 Certificate implmentation. This has been fixed.
- In line with latest changes in the JVM, SignatureSpis which don't require parameters now return null on engineGetParameters().
- The RSA KeyFactory now always preferentially produces RSAPrivateCrtKey where it can on requests for a KeySpec based on an RSAPrivateKey.
- CMSTypedStream$FullReaderStream now handles zero length reads correctly.
- Unecessary padding was added on KMAC when the key length was block aligned. This has been fixed.
- Zero length data would cause an unexpected exception from RFC5649WrapEngine. This has been fixed.
- OpenBSDBcrypt was failing to handle some valid prefixes. This has been fixed.

### Additional Features and Functionality

- Performance of Argon2 has been improved.
- Performance of Noekeon has been improved.
- A setSessionKeyObfuscation() method has been added to PublicKeyKeyEncryptionMethodGenerator to allow turning off of session key obfuscation (default is on, method primarily to get around early version GPG issues with AES-128 keys).
- Implemented 'safegcd' constant-time modular inversion (as well as a variable-time variant). It has replaced Fermat inversion in all our EC code, and BigInteger.modInverse in several other places, particularly signers. This improves side-channel protection, and also gives a significant performance boost.
- Performance of custom binary ECC curves and Edwards Curves has been improved.
- BCJSSE: New boolean system property 'org.bouncycastle.jsse.keyManager.checkEKU' allows to disable ExtendedKeyUsage restrictions when selecting credentials (although the peer may still complain).
- Initial support has been added for "Composite Keys and Signatures For Use In Internet PKI" using the test OID. Please note there will be further refinements to this as the draft is standardised.
- The BC EdDSA signature API now supports keys implementing all methods on the EdECKey and XECKey interfaces directly.
- Work has begun on classes to support the ETSI TS 103 097, Intelligent Transport Systems (ITS) in the bcpkix package.
- Further optimization work has been done on GCM.
- A NewHope based processor, similar to the one for Key Agreement has been added for trying to "quantum hard" KEM algorithms.
- PGP clear signed signatures now support SHA-224.
- Treating absent vs NULL as equivalent can now be configured by a system property. By default this is not enabled.
- Mode name checks in Cipher strings should now make sure an improper mode name always results in a NoSuchAlgorithmException.
- In line with changes in OpenSSL, the OpenSSLPBKDF now uses UTF8 encoding.

## Bouncy Castle Release: 1.66
Date:      2020, July 4th.

### Defects Fixed

- EdDSA verifiers now reset correctly after rejecting overly long signatures.
- BCJSSE: SSLSession.getPeerCertificateChain could throw NullPointerException. This has been fixed.
- qTESLA-I verifier would reject some valid signatures. This has been fixed.
- qTESLA verifiers now reject overly long signatures.
- PGP regression caused failure to preserve existing version header when headers were reset. This has now been fixed.
- PKIXNameConstraintValidator had a bad cast preventing use of multiple OtherName constraints. This has been fixed.
- Serialisation of the non-CRT RSA Private Key could cause a NullPointerException. This has been fixed.
- An extra 4 bytes was included in the start of HSS public key encodings. This has been fixed.
- CMS with Ed448 using a direct signature was using id-shake256-len rather than id-shake256. This has been fixed.
- Use of GCMParameterSpec could cause an AccessControlException under some circumstances. This has been fixed.
- DTLS: Fixed high-latency HelloVerifyRequest handshakes.
- An encoding bug for rightEncoded() in KMAC has been fixed.
- For a few values the cSHAKE implementation would add unnecessary pad bytes where the N and S strings produced encoded data that was block aligned. This has been fixed.
- There were a few circumstances where Argon2BytesGenerator might hit an unexpected null. These have been removed.

### Additional Features and Functionality

- The qTESLA signature algorithm has been updated to v2.8 (20191108).
- BCJSSE: Client-side OCSP stapling now supports status_request_v2 extension.
- Support has been added for PKIXRevocationChecker for users of Java 8 and later.
- Support has been added for "ocsp.enable", "ocsp.responderURL" for users of Java 8 and later.
- Support has been added for "org.bouncycastle.x509.enableCRLDP" to the PKIX validator.
- BCJSSE: Now supports system property 'jsse.enableFFDHE'
- BCJSSE: Now supports system properties 'jdk.tls.client.SignatureSchemes' and 'jdk.tls.server.SignatureSchemes'.
- Multi-release support has been added for Java 11 XECKeys.
- Multi-release support has been added for Java 15 EdECKeys.
- The MiscPEMGenerator will now output general PrivateKeyInfo structures.
- A new property "org.bouncycastle.pkcs8.v1_info_only" has been added to make the provider only produce version 1 PKCS8 PrivateKeyInfo structures.
- The PKIX CertPathBuilder will now take the target certificate from the target constraints if a specific certificate is given to the selector.
- BCJSSE: A range of ARIA and CAMELLIA cipher suites added to supported list.
- BCJSSE: Now supports the PSS signature schemes from RFC 8446 (TLS 1.2 onwards).
- Performance of the Base64 encoder has been improved.
- The PGPPublicKey class will now include direct key sigantures when checking for key expiry times.

### Notes

The qTESLA update breaks compatibility with previous versions. Private keys now include a hash of the public key at the end, and signatures are no longer interoperable with previous versions.


kare:
Build for 1.67 was auto cancelled by Travis. Perhaps auto cancellation can be turned off from Travis settings?

```
Done: Job Cancelled
This job was cancelled because the "Auto Cancellation" feature is currently enabled, and a more recent build (#2474) for pull request #1341 came in while this job was waiting to be processed.
```

AlexandrouR:
@SweeXordious  Looks good to me what do you think ?

SweeXordious (approved):

PR #1344, opened 2021-01-14
fixes closed connection before getting response from Geth

SweeXordious:
### What does this PR do?
Fixes`HttpService` closed connection before the Json-RPC request being completed.

`HttpService` class

### Why is it needed?
Fixes a bug introduced in the previous release

### DIFF
--- a/core/src/main/java/org/web3j/protocol/http/HttpService.java
+++ b/core/src/main/java/org/web3j/protocol/http/HttpService.java
@@ -182,8 +182,6 @@ public class HttpService extends Service {
     }
 
     private InputStream buildInputStream(ResponseBody responseBody) throws IOException {
-        InputStream inputStream = new ByteArrayInputStream(responseBody.bytes());
-
         if (includeRawResponse) {
             // we have to buffer the entire input payload, so that after processing
             // it can be re-read and used to populate the rawResponse field.
@@ -199,6 +197,8 @@ public class HttpService extends Service {
             }
 
             int bufferSize = (int) size;
+            InputStream inputStream = responseBody.byteStream();
+
             BufferedInputStream bufferedinputStream =
                     new BufferedInputStream(inputStream, bufferSize);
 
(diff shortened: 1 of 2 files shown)


integration-tests/src/test/java/org/web3j/protocol/geth/GethIT.java
```diff
@@ -12,27 +12,33 @@
...
 
     private Geth web3j;
+    private String url = "http://localhost:8545";
 
     public GethIT() {}
 
     @BeforeEach
     public void setUp() {
-        this.web3j = Geth.build(new HttpService("http://localhost:8545"));
+        this.web3j = Geth.build(new HttpService(url));
```
conor10, line 41: Doesn't the empty constructor for HttpService point to this URL?

core/src/main/java/org/web3j/protocol/http/HttpService.java
```diff
@@ -157,22 +157,20 @@ protected InputStream performIO(String request) throws IOException {
         okhttp3.Request httpRequest =
                 new okhttp3.Request.Builder().url(url).headers(headers).post(requestBody).build();
 
-        try (okhttp3.Response response = httpClient.newCall(httpRequest).execute()) {
```
conor10, line 160: how come the try with resources has been removed here?

SweeXordious: I think because the try with resource closes the resource after exiting the scope. However, we return an `InputStream` to the `responseBody` of the `response` in the try. This latter gets closed after exiting the scope then the `InputStream` has nothing to read. Not sure anyway if this is it.

fcorneli: I guess `okhttp3.Response` should really be closed. Only solution I see is to return a `ByteArrayInputStream` to "disconnect" the two streams completely.

SweeXordious: Something like this ? (I just added a change)

xaviarias (approved):

PR #1380, opened 2021-03-04
Fix #1379 infinity recursion for SolidityFunctionWrapper

### LINKED ISSUE #1379: SolidityFunctionWrapper go into recursion with generateSendTxForCalls=true
# SolidityFunctionWrapper recursive call
SolidityFunctionWrapper go into recursion with generateSendTxForCalls=true

## Steps To Reproduce
```
 new SolidityFunctionWrapper(
                true,
                false,
                true,
                Address.DEFAULT_LENGTH / Byte.SIZE)
                .generateJavaFiles(
                    Contract.class,
                    "Test",
                    Contract.BIN_NOT_PROVIDED,
                    loadContractDefinition(anyAbiFile),
                    destinationDir.toString(),
                    "test.pckg",
                    null);

 private static List<AbiDefinition> loadContractDefinition(File absFile) throws IOException {
        ObjectMapper objectMapper = ObjectMapperFactory.getObjectMapper();
        AbiDefinition[] abiDefinition = objectMapper.readValue(absFile, AbiDefinition[].class);
        return Arrays.asList(abiDefinition);
    }                    
```

### Expected behavior
Generate Java Wrapper for the contract

### Actual behavior
Recursive call without any condition ` results.addAll(buildFunctions(sendFuncDefinition));`

```
Generating pro.belbix.ethparser.web3.abi.contracts.harvest.WrappedVaultProxy ... Exception in thread "main" java.lang.StackOverflowError
	at java.base/java.util.ArrayList.<init>(ArrayList.java:179)
	at com.squareup.javapoet.Util.immutableList(Util.java:77)
	at com.squareup.javapoet.TypeName.<init>(TypeName.java:103)
	at com.squareup.javapoet.TypeName.<init>(TypeName.java:108)
	at com.squareup.javapoet.ClassName.<init>(ClassName.java:47)
	at com.squareup.javapoet.ClassName.<init>(ClassName.java:43)
	at com.squareup.javapoet.ClassName.get(ClassName.java:135)
	at com.squareup.javapoet.TypeName.get(TypeName.java:334)
	at com.squareup.javapoet.TypeName.get(TypeName.java:318)
	at com.squareup.javapoet.CodeBlock$Builder.argToType(CodeBlock.java:224)
	at com.squareup.javapoet.CodeBlock$Builder.add(CodeBlock.java:176)
	at com.squareup.javapoet.CodeBlock$Builder.addStatement(CodeBlock.java:267)
	at com.squareup.javapoet.MethodSpec$Builder.addStatement(MethodSpec.java:465)
	at org.web3j.codegen.SolidityFunctionWrapper.buildConstantFunction(SolidityFunctionWrapper.java:1346)
	at org.web3j.codegen.SolidityFunctionWrapper.buildFunctions(SolidityFunctionWrapper.java:1290)
```

## Environment
- Web3j version - 4.8.4
- Java 11
- Operating System Win10

with `sendFuncDefinition.setConstant(false);` need to also change StateMutability

belbix:
### What does this PR do?
Fix infinity recursion for SolidityFunctionWrapper

https://github.com/web3j/web3j/issues/1379

### Why is it needed?
able to generate wrappers with call_ functions

### DIFF
--- a/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
+++ b/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
@@ -1299,6 +1299,7 @@ public class SolidityFunctionWrapper extends Generator {
             if (generateSendTxForCalls) {
                 AbiDefinition sendFuncDefinition = new AbiDefinition(functionDefinition);
                 sendFuncDefinition.setConstant(false);
+                sendFuncDefinition.setStateMutability("nonpayable");
                 results.addAll(buildFunctions(sendFuncDefinition));
             }
         }
--- a/codegen/src/test/java/org/web3j/codegen/SolidityFunctionWrapperTest.java
+++ b/codegen/src/test/java/org/web3j/codegen/SolidityFunctionWrapperTest.java
@@ -12,20 +12,31 @@
  */
 package org.web3j.codegen;
 
-import java.math.BigInteger;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
+import static org.junit.jupiter.api.Assertions.assertAll;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.web3j.codegen.SolidityFunctionWrapper.buildTypeName;
+import static org.web3j.codegen.SolidityFunctionWrapper.createValidParamName;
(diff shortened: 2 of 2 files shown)


SweeXordious:
@belbix I am checking the logic behind your PR through the test.
You are passing a function similar to the following:
```
function test(uint8 param) public pure  {
   ...
}
```
and you want a wrapper function to be generated for it.

In this case, calling this function is meaningless. Because, it is `pure`, which means, it can only use the `parameters` provided to it, and has no `outputs`. Also, it is unable to make any change on the `blockchain` or even `emit events`.

I don't see a point in having a function to make this call ?
Correct me please if I am missing the point.

belbix:
ah, it is a synthetic function
the main problem was an infinity recursion
this test cover only it
if it is necessary I can use a more real input data

SweeXordious:
I agree on that. It is just I am not sure if we will hit that edge case when generating the wrappers

belbix:
> I agree on that. It is just I am not sure if we will hit that edge case when generating the wrappers. But yes, we shouldn't have the infinite recursion.

no problem, I changed a bit ABI
now it looks like a more common test case

PR #1381, opened 2021-03-07
generate struct java class even when they are used only in arrays

BJMg:
### What does this PR do?
It hopefully fixes a problem when sol source file contains struct only in array and not as individual type.

### Why is it needed?
To cover even more cases for java generated code.


SweeXordious (changes requested):
LGTM.
Sorry for the late reply and thanks a lot for spotting this and fixing it.

What I would suggest for the PR to be ready to merge is to add a test to `SolidityFunctionWrapperTest` containing the wanted generated code along with the struct generated. And then, I will gladly merge it.

SweeXordious:
@BJMg  Can you please leave this test in `SolidityFunctionWrapperGeneratorTest.java ` as it is important to have:
```
 @Test
    public void testStructOnlyInArray() throws Exception {
        testCodeGeneration("onlyinarraystruct", "OnlyInArrayStruct", JAVA_TYPES_ARG, false);
    }
```

and also keep the new one ?

SweeXordious:
Perfect, thanks a lot

SweeXordious (approved):
LGTM

xaviarias (approved):

PR #1383, opened 2021-03-14
WIP fixes nested structs with different parameters decoding 

SweeXordious:
### What does this PR do?
Fixes nested structs decoding

Added test

### Why is it needed?
Fix


SweeXordious:
The PR also contains some cosmetics to reduce the `warnings` when building.

PR #1386, opened 2021-03-17
Alignment with Besu OnChainPrivacyGroups signature modified from 20.10.…

albert0-hernandez:
### What does this PR do?
This PR tries to align the Besu OnChainPrivacyGroups with the last version of Besu.
Since the versión of besu 20.10.0, some changes were introduced in the Besu OnchainPermissioning Smartcontracts and the actual version of Web3j is unable to create an OnChainPrivacyGroups.

Reviewing the last versión of SC OnChainPrivacyGroups implementation and verify that the signature mismatch since v20.10.0

### Why is it needed?
- Fix the OnChainPrivacyGroups creation process
- Fix the hardcoded chainId for the rawTransaction
- Fix the ZeroGasFactory used to generate transaction
- Parametrize the hardcoded transaction such us Restriction and Address of the EeaRawTransaction

### DIFF
--- a/besu/src/main/java/org/web3j/protocol/besu/JsonRpc2_0Besu.java
+++ b/besu/src/main/java/org/web3j/protocol/besu/JsonRpc2_0Besu.java
@@ -64,6 +64,7 @@ public class JsonRpc2_0Besu extends JsonRpc2_0Eea implements Besu {
 
     private final JsonRpc2_0BesuRx besuRx;
     private final long blockTime;
+    private final OnChainPrivacyTransactionBuilder onChainPrivacyTransactionBuilder;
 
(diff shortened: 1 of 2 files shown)


lucassaldanha:
The changes seems to be correct. @albert0-hernandez Has this been tested using Besu 20.10.0?

albert0-hernandez:
> The changes seem to be correct. @albert0-hernandez Has this been tested using Besu 20.10.0?

The changes were tested on Besu 21.1.0 which maintains the addParticipants signature since 20.10.0 as you can check on the history.

mohamedelshami (approved):
The changes looks good.

SweeXordious (approved):

Issue #953, opened 2019-07-09
Request logging vs provided OkHttpClient 


hmijail:
## Feature description
The web3j request logging is only enabled when web3j creates its own OkHttpClient.
This client is private, so there is no way to customize it afterwards (like eg. setting its timeout value). 
So for example it is impossible to have at the same time a custom OkHttpClient + request logging.   
I can imagine 2 ways to improve on this:
- Give access to the internal client once created
- Add logging to an external client

vincentk:
To a degree, a similar issue was raised (and addressed) in https://github.com/web3j/web3j-spring-boot-starter/pull/4, though I would argue that indeed it would be nice to allow library users to customize the `OKHttpClient` that is being used.

SweeXordious:
Closing. https://github.com/web3j/web3j/pull/1387 merged

PR #1387, opened 2021-03-18
Expose `OkHttpClient.Builder` to `HttpService` users

vincentk:
Expose the pre-configured client builder to library users so that they can customize configurations such as
* logging,
* time-outs.

By making the `Builder` public, library users can customize the `OkHttpClient` configuration before constructing an instance of `HttpService` while re-using the present default configuration.

For motivation of this PR, see e.g. 
* https://github.com/web3j/web3j/issues/953
* https://github.com/web3j/web3j-spring-boot-starter/pull/4

### DIFF
--- a/core/src/main/java/org/web3j/protocol/http/HttpService.java
+++ b/core/src/main/java/org/web3j/protocol/http/HttpService.java
@@ -134,11 +134,15 @@ public class HttpService extends Service {
         this(DEFAULT_URL);
     }
 
-    private static OkHttpClient createOkHttpClient() {
+    public static OkHttpClient.Builder getOkHttpClientBuilder() {
         final OkHttpClient.Builder builder =
                 new OkHttpClient.Builder().connectionSpecs(CONNECTION_SPEC_LIST);
         configureLogging(builder);
-        return builder.build();
+        return builder;
+    }
+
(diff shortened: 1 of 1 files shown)


SweeXordious (approved):
LGTM

mohamedelshami:
Hi @vincentk we have not revisited the connectivity part in a while, but  are considering externalising additional configuration rather than exposing underlying HttpService implementation. 

For logging we can use possibly use Http Interceptor?

vincentk:
The use case that triggered this PR was along the lines of https://github.com/web3j/web3j-spring-boot-starter/pull/4 . I.e. time-out configuration, as opposed to e.g. logging.

This PR aimed to be a minimal change that facilitates this. Feel free to close this PR if you think it is not aligned with project goals. Personally, I feel this project is fairly tightly coupled to `OkHttp` already and exposing configuration options would do more good than harm.

mohamedelshami (approved):
I looked at this again recently for another issue, since we are also allowing the passing OkHttpClient as a constructor argument in the HttpService. This impact of this change is minimal and consistent with exposed public API.

mohamedelshami:
@vincentk I just had a chance to look at this again, I approved the change.

PR #1395, opened 2021-04-01
Fix PrivateTransactionManager sendCall()

lucassaldanha:
### What does this PR do?
Updates PrivateTransactionManager sendCall() to use `priv_call` instead of sending a tx to read values.

I have updated the `sendCall()` method to match the implementation on `ClientTransactionManager` (the public counterpart). `priv_call` behaves exactly like `eth_call`, however it expects the `privacyGroupId` as an extra parameter.

Given that Web3j already supports `priv_call`, this change was really simple and should be easy to review.

### Why is it needed?
It's been a while since Besu added support for reading information from private contracts using `priv_call` (https://besu.hyperledger.org/en/stable/Reference/API-Methods/#priv_call). Using this method makes more sense then sending a transaction to read information from a contract.

### DIFF
--- a/besu/src/main/java/org/web3j/tx/PrivateTransactionManager.java
+++ b/besu/src/main/java/org/web3j/tx/PrivateTransactionManager.java
@@ -22,15 +22,15 @@ import org.slf4j.LoggerFactory;
 import org.web3j.crypto.Credentials;
 import org.web3j.protocol.besu.Besu;
 import org.web3j.protocol.besu.response.privacy.PrivateEnclaveKey;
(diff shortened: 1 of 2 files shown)


lucassaldanha:
@SweeXordious I'd appreciate it if you could take a look at this PR.

SweeXordious:
@lucassaldanha Sorry for late reply, I will take a look on it ASAP.

SweeXordious (approved):
LGTM

mohamedelshami (approved):
Thanks for adding the test

PR #1428, opened 2021-06-03
Add ability to specify restriction for private transactions

### LINKED ISSUE #1424: Add support for UNRESTRICTED privacy in PrivateTransactionManager
## Feature description
Currently, the transaction restriction type is hardcoded to `RESTRICTED` in https://github.com/web3j/web3j/blob/master/besu/src/main/java/org/web3j/tx/PrivateTransactionManager.java#L140

Efforts should be made to remove the branching logic in the abstract class and rely on Open Closed Principles.

antonydenyer:
### What does this PR do?
Add ability to specify restriction for private transactions Fixes #1424

### Breaking Changes
I've made breaking changes to the constructor and switched to a builder pattern for optional arguments. These are easy to remedy when the user upgrades their version.


antonydenyer:
Is anyone free to take a look at this? :pray:

mohamedelshami:
Hi @antonydenyer, thanks for the PR. I have a couple of questions/points:

1. What's the requirement for restrictions? is it an optional parameter for Besu?
2. I'm more inclined to keep the private transaction constructor consistent with the core transaction types, my approach would be to add an overloaded constructor to pass restrictions parameter rather than introducing the builder pattern.
3. It looks like BesuPrivateTransactionManager and LegacyPrivateTransactionManager are not heavily used, but I would suggest marking them as deprecated rather than delete them right away? they can then be deleted later after a major release.

antonydenyer:
>| What's the requirement for restrictions? is it an optional parameter for Besu?
It's a required parameter for `eea_sendRawTransaction` https://entethalliance.github.io/client-spec/spec.html#dfn-restriction

>| I'm more inclined to keep the private transaction constructor consistent with the core transaction types, my approach would be to add an overloaded constructor to pass restrictions parameter rather than introducing the builder pattern.

Okay, removed 

>| It looks like BesuPrivateTransactionManager and LegacyPrivateTransactionManager are not heavily used, but I would suggest marking them as deprecated rather than delete them right away? they can then be deleted later after a major release.

Sure, marketd as deprecated


https://github.com/web3j/web3j/pull/1428/commits/e48eba412e9c4e9a291bb378eba55943f8c3f5f9

mohamedelshami:
Thank you! I've done the merge.

PR #1431, opened 2021-06-06
Minor fix to adapt to String type replyId

HentaiOusama:
Long ago, I was working using this library to work with matic chain (Polygon). Apparently, It seems that in the JSON, they send a long value in string format. So it kept throwing error in the code. I use this patched version in my code, but i thought i should make this change on original code and propose a pull request so that others don't have to go through the same issue.

### What does this PR do?
In this patch, if the id in JSON is not a long value, but still a long value in text format, then we check for that and if possible, we try to return that value.

Very very small change has been made. Just look at the changes made at line 435 and few lines below it.

### Why is it needed?
I believe it is required so that we can avoid errors when working with different types of chains. At least it is needed for matic when using maticVigil endpoints for receiving data.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
+++ b/core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
@@ -436,12 +436,16 @@ public class WebSocketService implements Web3jService {
             if (idField.isTextual()) {
                 try {
                     return Long.parseLong(idField.asText());
-                } catch (Exception e) {
-                    throw new IOException(String.format("Textual Found 'id' that cannot be casted to long. Input : '%s'", idField.asText()));
+                } catch (NumberFormatException e) {
+                    throw new IOException(
+                            String.format(
+                                    "Found Textual 'id' that cannot be casted to long. Input : '%s'",
+                                    idField.asText()));
                 }
             } else {
                 throw new IOException(
-                        String.format("'id' expected to be long, but it is: '%s'", idField.asText()));
+                        String.format(
+                                "'id' expected to be long, but it is: '%s'", idField.asText()));
             }
         }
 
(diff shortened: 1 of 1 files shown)


mohamedelshami (approved):

SweeXordious:
@HentaiOusama LGTM. Can you please `spotlessApply` so we can merge :D

core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
```diff
@@ -433,8 +433,16 @@ private long getReplyId(JsonNode replyJson) throws IOException {
         }
 
         if (!idField.isIntegralNumber()) {
-            throw new IOException(
-                    String.format("'id' expected to be long, but it is: '%s'", idField.asText()));
+            if (idField.isTextual()) {
+                try {
+                    return Long.parseLong(idField.asText());
+                } catch (Exception e) {
```
xaviarias, line 439: Shouldn't it make more sense to catch only `NumberFormatException`s?

xaviarias (approved):
LGTM

core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
HentaiOusama: I agree. I will update the Exception part. 

Also, I don't know how to `spotlessApply`, (I think it is about format of code, like spaces, structure, etc). I will search about it and see if i can do it. I have pulled the code and opened in intellij. Will see if reformat code does the job.

SweeXordious:
@HentaiOusama `./gradlew spotlessApply` 😄

HentaiOusama:
> @HentaiOusama `./gradlew spotlessApply` 😄

Yeah, I just figured it out from google as well. Applying it right now.

HentaiOusama:
I have performed `spotlessApply`. Can be mergerd now.

PR #1456, opened 2021-07-20
Signer Abstraction

albert0-hernandez:
### What does this PR do?
With this PR a new Signer component is created in order to abstract the signing process of public and private transactions.
Abstracting the signing as an independent process brings benefits for the security because the Transactions Managers do not have to store and handle the Private Key in the credentials objects and opens the possibility to integrate external signer components without changing the behaviour of the library. 

The reviewer should start validating the new interface Signer that should model the actual signing process. With this interface and the Default implementation (DefaultSigner) with stores but not expose the private key of the Credentails object all the logic is modelled.
The rest of the changes of the PR is creating new contructors for the Tx Managers to accept a signer instead of a Credentails.
The approach for the actual changes inside these Tx Managers was substitute the credentials object for the Default Signer.

### Why is it needed?
Moving to a wide and flexible signing solutions should be appreciate to enable the web3j more suitable and secure for both public and private solutions.

### DIFF
--- a/besu/src/main/java/org/web3j/protocol/besu/Besu.java
+++ b/besu/src/main/java/org/web3j/protocol/besu/Besu.java
@@ -18,6 +18,7 @@ import java.util.Map;
 import java.util.concurrent.ScheduledExecutorService;
 
 import org.web3j.crypto.Credentials;
+import org.web3j.crypto.signer.Signer;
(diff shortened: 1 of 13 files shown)


albert0-hernandez:
Please, let me know if you need more details on this PR.

PR #1458, opened 2021-07-21
Use string builder when generating binary param

antonydenyer:
### What does this PR do?
When generating sources with large contracts the binary can go over the 64k string limit.

Switched to use a string builder instead as per https://github.com/square/javapoet/issues/776


ricardolyn (approved):

ricardolyn:
@conor10 it will be good to merge and release this work as is currently a blocker for https://github.com/ConsenSys/quorum-acceptance-tests/pull/123. thank you!

conor10:
Sure, I'll get one of the team on it.

mohamedelshami:
@antonydenyer @ricardolyn I have done the merge let me know if there is anything that needs to be done here. 

As a side note, I was curious about the contract size limit in Quorum, it appears to allow 64kb as opposed to Etheruem EIP 170 (24kb)?

https://ethereum.stackexchange.com/questions/56000/quorum-oversized-data-during-contract-creation-contract-code-size-limit?rq=1

ricardolyn:
@mohamedelshami yes, GoQuorum allows to configure the contract dize limit by using [this](https://github.com/ConsenSys/quorum/blob/master/params/config.go#L343) and [this](https://github.com/ConsenSys/quorum/blob/master/params/config.go#L350) configuration on the genesis file.  

those values can be configured between 24kb and 128kb. More information on https://docs.goquorum.consensys.net/en/stable/HowTo/Configure/GenesisOptions/#contract-code-size

ricardolyn:
@conor10 @mohamedelshami 

I updated the version on the plugin: https://github.com/web3j/web3j-maven-plugin/pull/74

Issue #254, opened 2017-11-22
Decode Transaction.input data from Contract


sirhill:
My use case is the following:
I would like to have a daemon monitoring what is happening to my smart contract and how users are interacting with it.
I would like to set my

Similarly to what ConsenSys [https://github.com/ConsenSys/abi-decoder](abi-decoder) does in Javascript, it would be very practical to have a way to decode an input.

If there is interest in the feature and it makes sense (project-wise), I am available for the pull request.

Best,
Cyril

conor10:
web3j's smart contract wrappers enable you to do this already - are you familiar with how they work? See https://docs.web3j.io/smart_contracts.html#invoking-transactions-and-events.

sirhill:
Thanks a lot @conor10 for your reply.
If I understand your proposal, you are suggesting that I process the smart contract's events instead ?

Events cost a bit of gas and in my case all the information should already be available in the input data.
So there is no advantage to pay for some events.
I've dig a little bit on web3j's api but I was unable to find a method to decode it.
Hence my proposal to extends web3j for it.
What do you think ?

flefevre:
Hey, I do agree, it will make sense to have a function to decode directly the String of 

`TransactionResult<org.web3j.protocol.core.methods.response.Transaction> tr = block.getTransactions().get(0);
	tr.get().getInput();`

If we could retrieve it, as ascii or as json or eventually directly as the event.
But perhaps it is a miss understanding from me on how to go from block>transaction to TransactionReceipt ?

Thanks for your input

fcorneli:
I'm also looking for a way to decode the Input of a transaction.

iikirilov:
This is not an issue with Web3j. Web3j also support a number of decoding and encoding methods. If you have specific feature requests or issues please open a new thread with detail. Closing thread.

Please ask usage question on [gitter](https://gitter.im/web3j/web3j?utm_source=share-link&utm_medium=link&utm_campaign=share-link).

codebazz:
transaction.input=>input parameters
Any update ?

shanbub:
> transaction.input=>input parameters
> Any update ?

Just encountered this:
https://github.com/prettymuchbryce/abidecoder
But having trouble using it from my java project at the moment.

iikirilov:
Hey, after some thinking this may be possible but is non-trivial. It is possible to extend the contract wrapper generation to include the primitives you need to decode input of transactions to known contracts.

[...]

PR #1481, opened 2021-08-15
Codegen an abi's input TypeReference

djma:
### What does this PR do?
This modifies the smart contract abi codegen to include static variables that can be used to decode a transaction's `getInput()`.

Here's a sample usage:

```java
    Subscription subTxn = (Subscription) web3j
      .transactionFlowable()
      .subscribe(Web3App::acceptNewTxn);
  }

  private static void acceptNewTxn(Transaction tx) throws IOException {
    if (
      "0x7a250d5630b4cf539739df2c5dacb4c659f2488d".equals(tx.getTo()) && // UniswapV2Router02
      tx.getInput().startsWith("0x8803dbee") // swapTokensForExactTokens
    ) {
      System.out.println(tx.getTransactionIndex() + ":" + tx.getInput());

      var inTypes = UniswapV2Router02.FUNC_SWAPTOKENSFOREXACTTOKENS_INPUT;

      // Note that I'm reusing the static function in FunctionReturnDecoder, but it would technically be called FunctionInputDecoder or something.
      List<Type> decodedInput = FunctionReturnDecoder.decode(
        tx.getInput().substring(10),
        inTypes
      );

      for (Type type : decodedInput) {
        System.out.println(type.getTypeAsString() + ": " + type.getValue());
      }
    }
  }
```

Here is a sample addition in generated code:

```java
  public static final List<TypeReference<Type>> FUNC_SWAPTOKENSFOREXACTTOKENS_INPUT = Utils.convert(
            Arrays.<TypeReference<?>>asList(new TypeReference<Uint256>() {},
            new TypeReference<Uint256>() {},
            new TypeReference<DynamicArray<Address>>() {},
            new TypeReference<Address>() {},
            new TypeReference<Uint256>() {}));

```

If was generated by running `SolidityFunctionWrapperGenerator.java:main()` with `-a UniswapV2Router02.abi -o <path_to_dir> -p=org.path.to.generated.contracts`.

Since this is my first contribution, this PR is just a proof of concept to better understand the codebase. I plan on cleaning it up if we decide to move forward with it.

Please look at the added code in `SolidityFunctionWrapper.java:712`.

There are some whitespace changes, but I'm happy to undo them if somebody points me to the right autoformatter to use.

### Why is it needed?
I'm trying to subscribe to a transaction stream and get which function was called and with what parameters.

I ran into this [issue](https://github.com/web3j/web3j/issues/254) which matched my problem but it doesn't look like any progress was made despite the last comment by @iikirilov. (I think the linked PR isn't related.)

### DIFF
--- a/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
+++ b/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
@@ -49,6 +49,7 @@ import org.slf4j.LoggerFactory;
 import org.web3j.abi.EventEncoder;
 import org.web3j.abi.FunctionEncoder;
 import org.web3j.abi.TypeReference;
+import org.web3j.abi.Utils;
 import org.web3j.abi.datatypes.Address;
 import org.web3j.abi.datatypes.DynamicArray;
 import org.web3j.abi.datatypes.DynamicStruct;
@@ -84,7 +85,9 @@ import org.web3j.utils.Collection;
 import org.web3j.utils.Strings;
 import org.web3j.utils.Version;
 
-/** Generate Java Classes based on generated Solidity bin and abi files. */
+/**
+ * Generate Java Classes based on generated Solidity bin and abi files.
+ */
 public class SolidityFunctionWrapper extends Generator {
 
     private static final String BINARY = "BINARY";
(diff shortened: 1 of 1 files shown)


Conanza00 (approved):

Conanza00 (approved):

Conanza00 (approved):

SweeXordious (commented):
Sorry for the late reply.
Please if still needed, add a codegen test for this generation so that we know exactly what the output will be and whether it makes sense to generate or just add a generic enough method in the library to handle this.
Thank you :D

PR #1491, opened 2021-08-27
Add support to generate both `call` and `send` functions in wrappers

alexdupre:
### What does this PR do?
- replace the broken generateSendTxForCalls flag with generateBothCallAndSend
- keep backward compatibility
- add -B / --generateBoth flag to CLI
- add tests
- fix an ordering bug in construstor arguments

What does it mean?! :-)

### Why is it needed?
Because it fixes bugs and add the ability to call methods without sending transactions.


AlexandrouR:
Looks good

PR #1494, opened 2021-08-30
Fix ClassCastException when using native types (default) and arrays.

alexdupre:
### What does this PR do?
`getValue` on `Array` types doesn't automatically return a List of
native types, but a List of solidity types, that generates a
ClassCastException at runtime. This patch fixes it, by adding a new
function in `Array` class and using it in the generated code.

### Why is it needed?
To avoid ClassCastException

### DIFF
--- a/abi/src/main/java/org/web3j/abi/datatypes/Array.java
+++ b/abi/src/main/java/org/web3j/abi/datatypes/Array.java
@@ -14,6 +14,7 @@ package org.web3j.abi.datatypes;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


alexdupre:
The only activity to be done is to merge this pull request :-)

AlexandrouR:
LGTM @SweeXordious  Please have a look and if you agree merge

SweeXordious (approved):
LGTM. Please add a test for the codegen that reflects these changes, then I will gladly merge.

alexdupre:
@SweeXordious can this be finally merged, please?

PR #1507, opened 2021-09-14
Fix reinstallFilter interrupting its own thread, causing InterruptedI…

wxker95:
…OException when installing new filters

### What does this PR do?
Fixes the issue where an InterruptedIOException is thrown when reinstalling filters after a network issue or the geth node restarting. This is caused by a ScheduledFuture execution being cancelled with interruption enabled from within its execution. The fix disables the interruption.

Only one line is changed: line 158 of Filter.java

### Why is it needed?
Reinstallation of filters after a network issue or geth node restart is currently impossible other than restarting the app using the web3j module.


AlexandrouR:
LGTM

PR #1511, opened 2021-09-20
Add support for decoding EIP1559 transaction hex string.

yuvanesh-srib:
### What does this PR do?
*TransactionDecoder currently supports only decoding of legacy transaction hex string to a RawTransaction instance. This PR add implementation required to support the decoding of EIP1559 transaction hex string.*

*Reviewers can start looking at the TransactionDecoder class followed by the unit tests.*

### Why is it needed?
*It is needed to decode eip1559 transaction hex string to its corresponding RawTransaction.*

### DIFF
--- a/crypto/src/test/java/org/web3j/crypto/TransactionDecoderTest.java
+++ b/crypto/src/test/java/org/web3j/crypto/TransactionDecoderTest.java
@@ -12,16 +12,14 @@
  */
 package org.web3j.crypto;
 
-import java.math.BigInteger;
-
 import org.junit.jupiter.api.Test;
-
+import org.web3j.crypto.transaction.type.Transaction1559;
 import org.web3j.utils.Numeric;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import java.math.BigInteger;
+import java.security.SignatureException;
+
+import static org.junit.jupiter.api.Assertions.*;
 
 public class TransactionDecoderTest {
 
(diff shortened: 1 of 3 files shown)


mahima-manik:
Please add check on the lengths of rlpList and values array. It can cause index out of bound exceptions.

yuvanesh-srib:
> Please add check on the lengths of rlpList and values array. It can cause index out of bound exceptions.

Thanks for the review.

The original TransactionDecoder.decode() function throws these types of RuntimeExceptions like index out of bound, not a rlp string etc., incase of an invalid transaction hex string is passed to it. Hence to be in similar lines I have avoided these length checks.

wdyt ?

yuvanesh-srib:
@SweeXordious @AlexandrouR 

Could you please review this PR ?
Thanks.

SweeXordious (approved):
LGTM

xaviarias (approved):

crypto/src/test/java/org/web3j/crypto/TransactionDecoderTest.java
```diff
@@ -13,15 +13,14 @@
...
 import org.junit.jupiter.api.Test;
 
+import org.web3j.crypto.transaction.type.Transaction1559;
 import org.web3j.utils.Numeric;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
```
xaviarias, line 23: Wildcard imports are not used in the Web3j code style.

xaviarias (changes requested):
LGTM, just please remove wildcard imports.

mahima-manik:
As the serialized data is expected to be well formed, the checks for runtime exceptions are skipped.

Approving PR for now with suggestion:

To review the right place to add these checks and eliminate any runtime exceptions. Or add try-catch block in the decoder to show current message to end user.

crypto/src/test/java/org/web3j/crypto/TransactionDecoderTest.java
yuvanesh-srib: Done.

xaviarias (approved):

PR #1555, opened 2021-11-29
generate call and send for transaction functions

kalinbas:
### What does this PR do?
It allows the java code generator to generate both call and send functions for functions which do transactions (similar to staticcall in ethers.js). Before this change it didn't work because the constructor was protected so the CLI didn't allow the "generateBothCallAndSend" parameter. Also it generated the call_ and send_ functions for "view" and "pure" functions and not for transacting functions.

### Why is it needed?
Because at the moment "generateBothCallAndSend" parameter doesn't work at all.

### DIFF
--- a/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
+++ b/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
@@ -1333,7 +1333,7 @@ public class SolidityFunctionWrapper extends Generator {
 
         if (generateBothCallAndSend) {
             final String funcNamePrefix;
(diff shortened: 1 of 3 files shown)


alexdupre:
A few comments:

1. the CLI *does* allow to `generateBothCallAndSend`, it's via API that is not currently possible. I have a branch with a new public constructor to allow it, I'll create a PR soon
2. the code does exactly what it's expected to do, it generates both `call_` and `send_` functions for *all* methods: payable, non-payable, view, pure

alexdupre:
https://github.com/web3j/web3j/pull/1666

PR #1580, opened 2021-12-17
Don't pad filter id with leading zero

### LINKED ISSUE #1392: eth_getFilterLogs not working on hardhat due to hex prefix
# eth_getFilterLogs not working on hardhat due to hex prefix

## Steps To Reproduce
Run a hardhat server with a smart contract, execute a event flowable call against it.

### Expected behavior
It should work.

### Actual behavior
The filter call fails.
The sequence of calls to the endpoint is:
```
call:
{"jsonrpc":"2.0","method":"eth_newFilter","params":[{"topics":["0x22[....]"],"fromBlock":"0x1","toBlock":"0x174876e800","address":["0x5F[...]"]}],"id":1}
response: 
{"jsonrpc":"2.0","id":1,"result":"0x5"}

next call web3j -> api:
{"jsonrpc":"2.0","method":"eth_getFilterLogs","params":["0x05"],"id":2}
response:
{"jsonrpc":"2.0","id":2,"error":{"code":-32602,"message":"Errors encountered in param 0: Invalid value \"0x05\" supplied to : QUANTITY"}}
```

I checked with curl, and when the call would be `{"jsonrpc":"2.0","method":"eth_getFilterLogs","params":["0x5"],"id":2}`, i.e., without the leading zero in the hex code of the params, then it works.

The [API spec](https://ethereum.org/nl/developers/docs/apis/json-rpc/#hex-value-encoding) says "WRONG: 0x0400 (no leading zeroes allowed)", but web3j is using `Numeric::toHexStringWithPrefixSafe` which does add leading zeroes.
My particular problem comes from `JsonRpc2_0Web3j.java` line 524 where the `eth_getFilterLogs` request is built (and it uses that same method for a couple of other calls as well, which should probably also be changed?).

## Environment
- Web3j version 4.8.4
- JVM on Linux

jimmyneutront:
### What does this PR do?
This PR causes ethUninstallFilter, ethGetFilterChanges, and ethGetFilterLogs to NOT pad a filter id's hexadecimal string representation with a zero when the hexadecimal string is less than two characters long. This is accomplished by using `Numeric.toHexStringWithPrefix` instead of `Numeric.toHexStringWithPrefixSafe`.

The reviewer should start by examining the changes in `JsonRpc2_0Web3j.java`.

### Why is it needed?
Resolves #1392 . As specified in Issue #1392 , Web3j currently cannot get filter changes or logs or uninstall a filter on a Hardhat network if that filter's hexadecimal id is less than "0x10". This PR will fix that issue by bringing Web3j into compliance with the [official Ethereum JSON-RPC specification](https://ethereum.org/en/developers/docs/apis/json-rpc/) for [eth_UninstallFilter, eth_getFilterChanges and eth_GetFilterLogs](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/eth1.0-apis/assembled-spec/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=true&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false), which explicitly specifies that filter ids in requests should NOT contain a leading "0".

### DIFF
--- a/core/src/main/java/org/web3j/protocol/core/JsonRpc2_0Web3j.java
+++ b/core/src/main/java/org/web3j/protocol/core/JsonRpc2_0Web3j.java
@@ -504,7 +504,7 @@ public class JsonRpc2_0Web3j implements Web3j {
     public Request<?, EthUninstallFilter> ethUninstallFilter(BigInteger filterId) {
         return new Request<>(
                 "eth_uninstallFilter",
-                Arrays.asList(Numeric.toHexStringWithPrefixSafe(filterId)),
+                Arrays.asList(Numeric.toHexStringWithPrefix(filterId)),
                 web3jService,
                 EthUninstallFilter.class);
     }
@@ -513,7 +513,7 @@ public class JsonRpc2_0Web3j implements Web3j {
     public Request<?, EthLog> ethGetFilterChanges(BigInteger filterId) {
         return new Request<>(
                 "eth_getFilterChanges",
-                Arrays.asList(Numeric.toHexStringWithPrefixSafe(filterId)),
+                Arrays.asList(Numeric.toHexStringWithPrefix(filterId)),
                 web3jService,
                 EthLog.class);
     }
@@ -522,7 +522,7 @@ public class JsonRpc2_0Web3j implements Web3j {
     public Request<?, EthLog> ethGetFilterLogs(BigInteger filterId) {
         return new Request<>(
                 "eth_getFilterLogs",
-                Arrays.asList(Numeric.toHexStringWithPrefixSafe(filterId)),
+                Arrays.asList(Numeric.toHexStringWithPrefix(filterId)),
                 web3jService,
                 EthLog.class);
     }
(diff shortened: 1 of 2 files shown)


SweeXordious (commented):
Looks good, however, I am wondering:
1- Do we need to make these changes also for the whisper ?
2- Do we want to support both formats ? Padded and not ? If so, we should add a test for this, instead of replacing it...

jimmyneutront:
@SweeXordious 1: I'm not familiar with Ethereum Whisper, but I can read up on it if necessary. However, this PR is only intended to bring Web3J into compliance with the official JSON-RPC spec, so I don't believe that any changes are necessary for Whisper. 2: I'm willing to add another commit to support both formats. However, since the official spec explicitly states that leading zeros should not be used, I'm not sure that it is necessary. Either way, please let me know what you want; I'll happily make whatever changes are necessary ASAP.

PR #1587, opened 2021-12-24
Fix recId calculation

mbaxter:
### What does this PR do?
* Fix `Sign.getRecId()` calculation
  * Fix calculation for replay-protected signatures
  * Reject all invalid `v` params
  * Add units tests
* Consolidate signing constants

Start at `Sign.getRecId()`

### Why is it needed?
The `v` parameter for [EIP-155](https://eips.ethereum.org/EIPS/eip-155) compatible replay protected transactions is calculated with `v = recId + CHAIN_ID * 2 + 35` where `recId` is in `{0,1}`.  Which means in order to recover the `recId` we need to run: `v - CHAIN_ID * 2  - 35` rather than `v - CHAIN_ID * 2  + 35`.


mbaxter:
@conor10 - anyone available to review this?

andrii-kl:
Hi @mbaxter 
You did a good job, thank you for a help and sorry for a delay. All team work hardly on new features.
You have found a really good point to improve but there are quite a lot of transaction signing logic which use this old approach too. 
For example method `signedMessageHashToKey()` 
`int recId = header - 27;`
So to avoid mess in the code, need to fix it everywhere.

andrii-kl:
We also refactored this code to make it more reusable. So you should poll these changes, and resolve a conflict. If you need any help, just let me know.

andrii-kl (approved):

mbaxter:
@andrii-kl - thanks for getting this in!

PR #1597, opened 2022-01-08
Fix wrapper generator to create Struct java native constructor 

andrii-kl:
Fix wrapper generator to create Struct java native constructor, Add unit test

### What does this PR do?
Fix a contract wrapper generation issue with functionality "Struct"(solidity ^0.5.0). Generated class does not compiled.
Java native constructor generates in not right way, if object(struct) has an Array field.

### Why is it needed?
 To use Array fields in Objects (struct) for solidity contracts. Java native constructor generates in right way.
 It's a first part of the fix, there is a second case with second object constructor. Second case looks harder.

### Other
ArraysInStruct.so made with not right extension to avoid automatic contract wrapper generation. Right now wrapper contract won't compile because of second bug with second entity constructor.


AlexandrouR (approved):
Looks good

PR #1602, opened 2022-01-17
Feature/364 integretion with hsm

andrii-kl:
### What does this PR do?
Add functionality to work with HSM (hardware security module). Reference implementation of working with HSM over HTTP 

### Why is it needed?
To sign transactions with hard wallet / HSM (hardware security module)


AlexandrouR (approved):
LGTM

besu/src/main/java/org/web3j/tx/PrivateTransactionManager.java
```diff
@@ -200,14 +201,7 @@ public EthGetCode getCode(
...
 
-        final byte[] signedMessage;
-
-        if (chainId > ChainIdLong.NONE) {
-            signedMessage =
-                    PrivateTransactionEncoder.signMessage(rawTransaction, chainId, credentials);
-        } else {
-            signedMessage = PrivateTransactionEncoder.signMessage(rawTransaction, credentials);
-        }
+        final byte[] signedMessage = txSignService.sign(rawTransaction, chainId);
```
freemanzMrojo, line 204: Hi @andrii-kl , I think this change is not ok since it gets rid of the `PrivateTransactionEncoder` to replace it with the regular `TransactionEncoder` within `TxSignServiceImpl` .

I have run some tests in Besu using `4.9.3` and indeed the signature is different.

Let me know if I am missing something, otherwise probably it would be good to open an issue for the fix.

Thanks in advance, Miguel.

antonydenyer: just spotted this also!

freemanzMrojo: Thanks for creating the issue @antonydenyer ! If @andrii-kl is not available I could take the implementation of the fix.

nissimr:
Is there any documentation about how to use this with a hardware wallet?

PR #1608, opened 2022-01-19
366 add chainid field, fix chainid calculation

andrii-kl:
### What does this PR do?
Fix chainId calculation

### Why is it needed?
Current Geth implementation has a new field "ChainId"
Web3j calculates chainId incorrectly

### Other

Links on Geth:
ChainId field - https://github.com/ethereum/go-ethereum/blob/c503f98f6d5e80e079c1d8a3601d188af2a899da/core/types/transaction_marshalling.go#L46
ChainId calculation - https://github.com/ethereum/go-ethereum/blob/4e599ee469223ccc452f40f36c85bac3bea9c16b/core/types/transaction_signing.go#L510


AlexandrouR (approved):
LGTM

Issue #1475, opened 2021-08-07
tightly pack arguments for Hash.sha3()


MGCR71990:
Hi
I want to hash a concatenation of the types (bytes ,address, unit8, unit256) both off-chain (Java using web3j) and on-chain (Solidity), and I want both hashing results to be the same.

so, I have the following function in solidity:

`
function calculateHash (address _myAddress, unit8 amount, unit256 nonce) internal pure returns (bytes32) 
{
        return keccak256("\x19Ethereum Signed Message:\n32", _myAddress, amount, nonce);
}
`
how I can get the same hashing results of the above function in java using Hash.sha3()?

Thanks,
Mohamad

PR #1628, opened 2022-02-15
FIX-1475 | add encodePacked

gtebrean:
### What does this PR do?
This PR adds the abi.encodePacked functionality

Should start from the abi module in FunctionEncoder

### Why is it needed?
It is needed to provide this abi.encodePacked in order to can encode values and to get the right keccak256 on it as in EVM
https://github.com/web3j/web3j/issues/1475
https://trello.com/c/I4qyUzsJ/362-adding-support-for-sha3-with-multiple-arguments-in-web3j


abi/src/main/java/org/web3j/abi/TypeEncoder.java
```diff
@@ -326,4 +384,39 @@ private static String encodeDynamicStructValues(final DynamicStruct value) {
...
+    /**
+     * Checks if the received array doesn't contain any element that can make the array unsupported
+     * for abi.encodePacked
+     *
+     * @param value Array to which the abi.encodePacked should be applied
+     * @param <T> Types of elements from the array
+     * @return if the encodePacked is supported for the given array
+     */
+    private static <T extends Type> boolean isSupportingEncodedPacked(Array<T> value) {
+        if (Utf8String.class.isAssignableFrom(value.getComponentType())
```
andrii-kl, line 397: It looks like you make encode/decode Utf8String below. (DefaultFunctionEncoderTest.java: 92)

gtebrean: indeed but here I check only for arrays if supports encodedPacked, because for example now encodedPacked is not supported for arrays of Utf8Strings, or other nested arrays

andrii-kl (approved):

PR #1645, opened 2022-02-25
Fix reconnect on refused connection

### LINKED ISSUE #1356: Reconnecting WebSocketService after a close call
I'm wondering how to reconnect a WebSocketService after a close without re-instantiating the WebSocketService object and the related Web3j object and contract.

zilm13:
### What does this PR do?
Toggles `shouldReConnect` for connection refused case too.
Without it I get something like this in logs when trying to reconnect node 2nd time after 1st attempt got connection refused:
```
Caused by: java.lang.IllegalStateException: WebSocketClient objects are not reuseable
at org.java_websocket.client.WebSocketClient.connect(WebSocketClient.java:372) ~[Java-WebSocket-1.5.2.jar:?]
at org.java_websocket.client.WebSocketClient.connectBlocking(WebSocketClient.java:386) ~[Java-WebSocket-1.5.2.jar:?]
at org.web3j.protocol.websocket.WebSocketService.connectToWebSocket(WebSocketService.java:129) ~[core-4.8.9.jar:?]
at org.web3j.protocol.websocket.WebSocketService.connect(WebSocketService.java:117) ~[core-4.8.9.jar:?]
``` 

`WebSocketClient.java`

### Why is it needed?
Otherwise user needs to recreate `WebSocketClient` and `WebSocketService` instances for every reconnection attempt when connection is refused.
Also will fix #1356 I guess

### DIFF
--- a/core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
+++ b/core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
@@ -128,11 +128,10 @@ public class WebSocketService implements Web3jService {
                         ? webSocketClient.reconnectBlocking()
                         : webSocketClient.connectBlocking();
 
+        shouldReConnect = true;
         if (!connected) {
             throw new ConnectException("Failed to connect to WebSocket");
         }
-
-        shouldReConnect = true;
     }
 
     private void setWebSocketListener(
--- a/core/src/test/java/org/web3j/protocol/websocket/WebSocketServiceTest.java
+++ b/core/src/test/java/org/web3j/protocol/websocket/WebSocketServiceTest.java
@@ -51,6 +51,7 @@ import static org.mockito.ArgumentMatchers.startsWith;
 import static org.mockito.Mockito.atMostOnce;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
@@ -119,6 +120,21 @@ public class WebSocketServiceTest {
                 () -> {
                     service.connect();
                 });
+    }
+
+    @Test
+    public void testReconnectIfFirstConnectionFailed() throws Exception {
+        when(webSocketClient.connectBlocking()).thenReturn(false);
+        assertThrows(
+                ConnectException.class,
+                () -> {
+                    service.connect();
+                });
+        service.connect();
+        // reconnectBlocking() should be called if 1st attempt was failed, if we call
+        // connectBlocking() for a second time, we will get IllegalStateException with real
+        // webSocketClient
+        verify(webSocketClient, times(1)).reconnectBlocking();
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
```diff
@@ -128,11 +128,10 @@ private void connectToWebSocket() throws InterruptedException, ConnectException
                         ? webSocketClient.reconnectBlocking()
                         : webSocketClient.connectBlocking();
 
+        shouldReConnect = true;
```
andrii-kl, line 131: I would add a functionality to make some limited attempts of web socket connection. (webSocketClient.connectBlocking())

zilm13: > I would add a functionality to make some limited attempts of web socket connection. (webSocketClient.connectBlocking())

I don't understand. `connect`which calls `connectToWebsocket` is a public API, so the external user could decide whether and when he wants to reconnect. You want to make such logic to be internal?

zilm13: @andrii-kl added test, which fails in a current master

core/src/test/java/org/web3j/protocol/websocket/WebSocketServiceTest.java
```diff
@@ -121,6 +122,21 @@ public void testThrowExceptionIfConnectionFailed() throws Exception {
...
+        assertThrows(
+                ConnectException.class,
+                () -> {
+                    service.connect();
+                });
+        service.connect();
+        // reconnectBlocking() should be called if 1st attempt was failed, if we call
+        // connectBlocking() for a second time, we will get IllegalStateException with real
+        // webSocketClient
+        verify(webSocketClient, times(1)).reconnectBlocking();
```
gtebrean, line 137:
```suggestion
        verify(webSocketClient, atMostOnce()).reconnectBlocking();
```
nit, but to be consistent with other test cases

core/src/main/java/org/web3j/protocol/websocket/WebSocketService.java
gtebrean: @zilm13 I guess @andrii-kl wanted to say to set up a counter in order to establish a connection pool. I took a look at code and I don't think is necessary, even more there is the onClose() method that is called on every listener.
Your test makes sense. 👍

gtebrean (approved):

core/src/test/java/org/web3j/protocol/websocket/WebSocketServiceTest.java
gtebrean: I prefer times too, but also I like classes to be as clear and consistent as possible, so adding that now would mean to update all checks in test and is not really the case. Thanks for update!

andrii-kl (approved):

zilm13:
@andrii-kl any schedule, when this could hit release? We need this change to implement Websockets support in Teku

Issue #1240, opened 2020-07-23
FunctionReturnDecoder causes NPE in ServiceLoader 


MarkVilkel:
# FunctionReturnDecoder causes NPE in ServiceLoader when called from different threads
```
java.lang.NullPointerException: null
	at java.util.ServiceLoader$ModuleServicesLookupIterator.hasNext(ServiceLoader.java:1067) ~[?:?]
	at java.util.ServiceLoader$2.hasNext(ServiceLoader.java:1299) ~[?:?]
	at java.util.ServiceLoader$3.hasNext(ServiceLoader.java:1384) ~[?:?]
	at org.web3j.abi.FunctionReturnDecoder.decoder(FunctionReturnDecoder.java:87) ~[abi-5.0.0.jar:?]
	at org.web3j.abi.FunctionReturnDecoder.decode(FunctionReturnDecoder.java:49) ~[abi-5.0.0.jar:?]
	at org.web3j.tx.Contract.executeCall(Contract.java:294) ~[core-5.0.0.jar:?]
	at org.web3j.tx.Contract.executeCallSingleValueReturn(Contract.java:300) ~[core-5.0.0.jar:?]
	at org.web3j.tx.Contract.executeCallSingleValueReturn(Contract.java:311) ~[core-5.0.0.jar:?]
	at org.web3j.tx.Contract.lambda$executeRemoteCallSingleValueReturn$1(Contract.java:397) ~[core-5.0.0.jar:?]
	at org.web3j.protocol.core.RemoteCall.send(RemoteCall.java:42) ~[core-5.0.0.jar:?]
```

This happens because 
```
    private static final ServiceLoader<FunctionReturnDecoderProvider> loader =
            ServiceLoader.load(FunctionReturnDecoderProvider.class);
```
is not thread safe static field.
So method
```
    private static FunctionReturnDecoder decoder() {
        final Iterator<FunctionReturnDecoderProvider> iterator = loader.iterator();
        return iterator.hasNext() ? iterator.next().get() : defaultDecoder();
    }
```
gets different iterator instance but with the same lookupIterator1 instance inside (which is not thread safe)

## The possible solution
would be to replace this code 
```
    private static final ServiceLoader<FunctionReturnDecoderProvider> loader =
            ServiceLoader.load(FunctionReturnDecoderProvider.class);
```
with this
```
protected static final ThreadLocal<ServiceLoader<FunctionReturnDecoderProvider>> loaderThreadLocal = 
[...]

PR #1648, opened 2022-02-28
FIX-1240 - Race condition in java.util.ServiceLoader

gtebrean:
### What does this PR do?
solves https://github.com/web3j/web3j/issues/1240

changed files

### Why is it needed?
there is a bug in java.util.ServiceLoader in multi-thread environment see:
[*required*](https://bugs.openjdk.java.net/browse/JDK-8230843)

### DIFF
--- a/abi/src/main/java/org/web3j/abi/FunctionReturnDecoder.java
+++ b/abi/src/main/java/org/web3j/abi/FunctionReturnDecoder.java
@@ -82,7 +82,7 @@ public abstract class FunctionReturnDecoder {
     protected abstract <T extends Type> Type decodeEventParameter(
             String rawInput, TypeReference<T> typeReference);
 
-    private static FunctionReturnDecoder decoder() {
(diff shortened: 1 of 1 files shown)


abi/src/main/java/org/web3j/abi/FunctionReturnDecoder.java
```diff
@@ -32,10 +32,15 @@
...
 
-    private static final ServiceLoader<FunctionReturnDecoderProvider> loader =
-            ServiceLoader.load(FunctionReturnDecoderProvider.class);
+    static {
+        ServiceLoader<FunctionReturnDecoderProvider> loader =
+                ServiceLoader.load(FunctionReturnDecoderProvider.class);
+        final Iterator<FunctionReturnDecoderProvider> iterator = loader.iterator();
+
+        decoder = iterator.hasNext() ? iterator.next().get() : new DefaultFunctionReturnDecoder();
+    }
```
gtebrean, line 43: tested ✅

PR #1650, opened 2022-03-01
Make functions parsing the event logs explicitly static.

alexdupre:
### What does this PR do?
It allows to parse event logs from transaction receipts without the need to fully load the related contract.

### Why is it needed?
Because it doesn't make sense to load contracts passing address, web3j, transaction manager and gas provider objects to simply parse an event log.


alexdupre:
This bot is annoying, the PR is still valid and just needs to be merged.

PR #1656, opened 2022-03-10
Fix retrieval of revert reason by using the same weiAmount in the call.

alexdupre:
### What does this PR do?
It fixes the retrieval of the tx revert reason, by using the same weiValue amount that was used in the failed transaction.

Consider this simplified test case contract:
```
contract Test {

    function test(uint n) external payable {
        require(msg.value > 0, 'msg.value must be positive');
        require(msg.value == n, 'msg.value must be equal to n');
    }
}
```
Then call it with `n == 2` and `weiAmount == 1`, and you'll get a super misleading `msg.value must be positive` error message.

### Why is it needed?
Because it drove me crazy for 24h to debug a complex smart contract when the thrown errors were in completely different code locations.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/core/methods/request/Transaction.java
+++ b/core/src/main/java/org/web3j/protocol/core/methods/request/Transaction.java
@@ -132,6 +132,12 @@ public class Transaction {
         return new Transaction(from, nonce, gasPrice, gasLimit, to, null, data);
     }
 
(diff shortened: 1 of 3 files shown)


alexdupre:
This bot is annoying, the PR is still valid and just needs to be merged.

andrii-kl (approved):

andrii-kl (approved):

PR #1666, opened 2022-03-22
Allow to set `generateBothCallAndSend` via API.

alexdupre:
### What does this PR do?
It allows to generate wrappers with both `call_` and `send_` transactions using the API. Previously it was allowed only via the CLI.

There is just a new public constructor.

### Why is it needed?
Self-evident.

### DIFF
--- a/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapperGenerator.java
+++ b/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapperGenerator.java
@@ -116,6 +116,32 @@ public class SolidityFunctionWrapperGenerator extends FunctionWrapperGenerator {
                 abiFuncs);
     }
 
+    public SolidityFunctionWrapperGenerator(
+            File binFile,
+            File abiFile,
+            File destinationDir,
+            String contractName,
+            String basePackageName,
+            boolean useJavaNativeTypes,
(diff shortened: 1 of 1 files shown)


codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapperGenerator.java
```diff
@@ -116,6 +116,32 @@ public SolidityFunctionWrapperGenerator(
                 abiFuncs);
     }
 
+    public SolidityFunctionWrapperGenerator(
```
andrii-kl, line 119: Let's extend visibility of the existing constructor, instead of creating a new one.

alexdupre: That's fine by me, even easier (I thought there was a reason to not make it public), you an go ahead.

andrii-kl: Great, let's make it and merge this PR.

alexdupre:
Please do, take credit, I don't care.

andrii-kl:
I have add a small changes, with constructor visibility&
https://github.com/web3j/web3j/pull/1692/commits/cc07d3f2d94b338a99328281f418ee56bc1274a7

PR #1673, opened 2022-03-31
Fix array types in solidity struct constructor

vchernetskyi993:
### What does this PR do?

Fix compilation issue with array types in struct constructor.

SolidityFunctionWrapper line 509.

### Why is it needed?

Ley's say we have the following contract:

```solidity
contract MyContract {
    struct MyStruct {
        string[] items;
    }
}
```

Current code with `useNativeJavaTypes=true` produces the following solidity-native constructor:
```java
public MyStruct(DynamicArray<Utf8String> items) {
    this.items = items.getValue();
}
```
`getValue()` returns `List<Utf8String>`, but `this.items` is `List<String>`.

So, we need to convert each value of the List to the Java type.

P.S. I don't like the look of generated code with this stream transformation appended. It looks clumsy. Ideally, we need to have some method like `array.getValues()` that will do this conversion internally. But I wasn't sure how and where to put such a method, any advice is welcome.


andrii-kl:
@vchernetskyi993 Thank you, great solution!

Filter94:
@andrii-kl  When can we expect a release with this fix?

PR #1683, opened 2022-04-18
Fix EIP712 encodeType with Array

OptimisticQuan:
### What does this PR do?
if types define as below
``` JSON
types: {
  "Group": [
    {
      "name": "name",
      "type": "string"
    },
    {
      "name": "members",
      "type": "Person[]"
    }
  ],
  "Mail": [
    {
      "name": "from",
      "type": "Person"
    },
    {
      "name": "to",
      "type": "Group[]"
    },
    {
      "name": "contents",
      "type": "string"
    }
  ],
  "Person": [
    {
      "name": "name",
      "type": "string"
    },
    {
      "name": "wallets",
      "type": "address[]"
    }
  ]
}
```
`encodeType("Group")` result
expect: `"Group(string name,Person[] members)Person(string name,address[] wallets)"`
actual: `"Group(string name,Person[] members)"`

because `StructuredDataEncoder.getDependencies` does not handle the case where the fields are arrays

This PR contains Fix and Test code.

Changed Files

### Why is it needed?
EIP-712 with Array leads to the wrong result.
This PR fix it.

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
+++ b/crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
@@ -88,14 +88,15 @@ public class StructuredDataEncoder {
             remainingTypes.remove(remainingTypes.size() - 1);
             deps.add(structName);
 
-            for (StructuredData.Entry entry : types.get(primaryType)) {
-                if (!types.containsKey(entry.getType())) {
+            for (StructuredData.Entry entry : types.get(structName)) {
+                String baseDeclarationTypeName = getBaseTypeNameFromDeclaration(entry.getType());
+                if (!types.containsKey(baseDeclarationTypeName)) {
                     // Don't expand on non-user defined types
-                } else if (deps.contains(entry.getType())) {
+                } else if (deps.contains(baseDeclarationTypeName)) {
                     // Skip types which are already expanded
                 } else {
                     // Encountered a user defined type
-                    remainingTypes.add(entry.getType());
+                    remainingTypes.add(baseDeclarationTypeName);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -88,14 +88,15 @@ public Set<String> getDependencies(String primaryType) {
             remainingTypes.remove(remainingTypes.size() - 1);
             deps.add(structName);
 
-            for (StructuredData.Entry entry : types.get(primaryType)) {
-                if (!types.containsKey(entry.getType())) {
+            for (StructuredData.Entry entry : types.get(structName)) {
+                String baseDeclarationTypeName = getBaseTypeNameFromDeclaration(entry.getType());
```
gtebrean, line 92:
```suggestion
                String baseDeclarationTypeName = arrayTypePattern.matcher(declaration).find() ? declaration.substring(0, declaration.indexOf('[')) : declaration;
```
no additional method need it

gtebrean (commented):
thanks for investigation, I know was not easy to debug this. Just one comment and if it builds successfully you are good to go.

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
OptimisticQuan: ok, I have apply and fix and test the suggested code

gtebrean: please run :spotlessJavaApply in order to solve code  style issue

gtebrean (approved):

PR #1690, opened 2022-04-25
Make ClientConnectionException an IOException

luziusmeisser:
A ClientConnectionException happens when there is a failure to connect to the target url. For example, it could happen if the server responds with an 503 error code. Therefore, it is a typical IOException and not a RuntimeException.

### What does this PR do?
Make ClientConnectionException an IOException

There is only one line of code plus an import statement that changed.

### Why is it needed?
To better conform to general Java design conventions and the principle of the least surprise. If a connection problem happens, an IOException and not a RuntimeException is the expected exception type.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/exceptions/ClientConnectionException.java
+++ b/core/src/main/java/org/web3j/protocol/exceptions/ClientConnectionException.java
@@ -12,8 +12,10 @@
  */
 package org.web3j.protocol.exceptions;
 
+import java.io.IOException;
+
 /** Client connection exception. */
-public class ClientConnectionException extends RuntimeException {
+public class ClientConnectionException extends IOException {
     public ClientConnectionException(String message) {
         super(message);
     }


andrii-kl:
@luziusmeisser  We use unchecked for custom exceptions. I don't think it make sense to do some mix. 
But in any case, let's wait, other opinions.

luziusmeisser:
Why not making use of the fact that Java offers two types of exceptions?

That's like trying to avoid capital letters for simplicity. :) In fact, both checked and unchecked exceptions have their use cases. Checked exceptions should be used for things that can be expected to go wrong every now and then, even if all the code is bug-free, whereas unchecked exceptions should be used when something goes wrong that never should if everything is implemented correctly.

luziusmeisser:
Of course you don't "need" to. It's a matter of convenience and reducing the probability that your users face unexpected errors.

Also note that this is not an issue of backwards compatibility as IOException was already thrown before.

luziusmeisser:
We just had a server crash because the backend was only programmed to handle IOException gracefully, whereas RuntimeExceptions are escalated (as they should). It would be much simpler to fix this in web3j directly. Today, we have to write "catch (IOException | ClientConnectionException e)" in locations where one usually could just write "catch (IOException e)" whenever we intereact with web3j.

@andrii-kl : I'll donate 1 ETH to this project if you accept my pull request to fix this issue. It is backwards compatible, poses no risks, makes me happy, and your project get some funds for team beers. What do you think?

PR #1706, opened 2022-05-12
Fix crash in android version < 28 

hoanganhtuan95ptit:
### What does this PR do?
My project is crashing on all devices with android version < 28, caused by getTypeName() function not found. Even though I have my project set up in java 8

Utils.java in module abi

### Why is it needed?
Help the system run stably and without crashing on devices with android version < 28

### DIFF
--- a/abi/src/main/java/org/web3j/abi/Utils.java
+++ b/abi/src/main/java/org/web3j/abi/Utils.java
@@ -12,18 +12,6 @@
  */
 package org.web3j.abi;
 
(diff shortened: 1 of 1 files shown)


amal:
Issue is not stale

andrii-kl:
Hi @hoanganhtuan95ptit thank you for your big work. 
We have merged other PR with these changes because that fix was more complete.
https://github.com/web3j/web3j/pull/1717

PR #1715, opened 2022-05-30
Privacy public key length constraint removed

freemanzMrojo:
### What does this PR do?
It removes the 44-byte long constraint for enclave (privacy) public keys in base64. With the enforced 44 length is not possible to make web3j work along with the [Tessera's EC encryptor mode](https://docs.tessera.consensys.net/en/stable/HowTo/Configure/Cryptographic-elliptic-curves/#configure-an-alternative-cryptographic-elliptic-curve).

It should start [reading Tessera's EC encryptor docs](https://docs.tessera.consensys.net/en/stable/HowTo/Configure/Cryptographic-elliptic-curves/#configure-an-alternative-cryptographic-elliptic-curve) and then [this Besu issue](https://github.com/hyperledger/besu/issues/3819) opened due to the same reason. Then the change itself is pretty straightforward within web3j, it is just about replacing a fixed-length byte array by [Tuweni Bytes](https://tuweni.apache.org/tutorials/libraries/getting-started-with-bytes/).

### Why is it needed?
Without this change, Tessera's EC encryptor does not work. Not only that, for any custom encryptor configuration on the privacy side in which the public key in base64 is different to 44 bytes it does not work either.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,7 +31,6 @@ ext {
     logbackVersion = '1.2.3'
     mockitoJunitVersion = '3.1.0'
     junitPlatformLauncherVersion = '1.5.2'
-    tuweniVersion = '2.2.0'
 }
 
 
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,7 @@ ext {
     logbackVersion = '1.2.3'
     mockitoJunitVersion = '3.1.0'
     junitPlatformLauncherVersion = '1.5.2'
+    tuweniVersion = '2.2.0'
 }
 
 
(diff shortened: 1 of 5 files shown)


antonydenyer:
Is there any chance you can make the change without adding another dependency? Currently, the runtime dependencies for web3j are very minimal. You should have everything you need within web3j.

freemanzMrojo:
Sure @antonydenyer , I can do that. Will apply the change in the next commit, thanks!

antonydenyer:
@AlexandrouR any chance you can review for us?

PR #1729, opened 2022-06-15
Add QBFT API methods providing access to the QBFT consensus engine

mkrielza:
### What does this PR do?
Add QBFT API methods to the Besu package.

Besu supports consensus engine specific endpoints for IBFT and QBFT. Web3j only supports IBFT. 

### Why is it needed?
It provides access to the QBFT consensus engine.


andrii-kl (approved):

andrii-kl:
Looks good but build failed.
Could you you run command `./gradlew spotlessApply`
to fix this problem

PR #1756, opened 2022-08-04
Allow TransactionHash objects to be serialised

calmacfadden:
### What does this PR do?
- The public get() function is never called by a serialiser to get the hash string as it doesnt confirm to the public getter standard of get + field name. ie. getValue()
- Adds a public getter getValue() to return the transaction hash as a string. This allows serialisers to correctly serialise the transactionhash object.

Create an instance of TransactionHash and attempt to serialise it and confirm that it serializes to a json string in the format of:

```
                {
                    "value": "0x2a9e6ea167..........."
                }
```

### Why is it needed?
- When a TransactionHash object is serialised to json it results in an empty object eg. "{}"

### DIFF
--- a/core/src/main/java/org/web3j/protocol/core/methods/response/EthBlock.java
+++ b/core/src/main/java/org/web3j/protocol/core/methods/response/EthBlock.java
@@ -538,6 +538,10 @@ public class EthBlock extends Response<EthBlock.Block> {
         public String get() {
             return value;
         }
+        
+        public String getValue() {
+            return value;
+        }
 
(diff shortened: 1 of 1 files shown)


rdovgan (approved):

mohamedelshami:
Hi @calmacfadden, thanks for pointing this out. This is a minor change and seems harmless. I would expect this to apply to all `org.web3j.protocol.core.*` response classes. 

I believe you're referring to the Jackson JSON serializer, could you please point me to the documentation where is `getValue` is expected?

Issue #489, opened 2018-04-09
How to Decoding transaction input data


githubityu:
https://www.queryoverflow.gdn/query/decoding-transaction-input-data-with-web3j-java-29_31351.html

relaxbao:
i have the same question

ghqgfj:
Hope below code snippet will help you
` String inputData = "0xa9059cbb0000000000000000000000005c5212ed85cc957c6b656d209a7be8812ab00e330000000000000000000000000000000000000000000000008d8dadf544fc0000";
        String method = inputData.substring(0,10);
        System.out.println(method);
        String to = inputData.substring(10,74);
        String value = inputData.substring(74);
        Method refMethod = TypeDecoder.class.getDeclaredMethod("decode",String.class,int.class,Class.class);
        refMethod.setAccessible(true);
        Address address = (Address)refMethod.invoke(null,to,0,Address.class);
        System.out.println(address.toString());
        Uint256 amount = (Uint256) refMethod.invoke(null,value,0,Uint256.class);
        System.out.println(amount.getValue());`

githubityu:
@ghqgfj thanks

penuel-leo:
@githubityu I got some input data that length less than 74

nicoflores95:
> Espero que el código siguiente te ayude
> 
> ```
>  String inputData = "0xa9059cbb0000000000000000000000005c5212ed85cc957c6b656d209a7be8812ab00e330000000000000000000000000000000000000000000000008d8dadf544fc0000";
>         String method = inputData.substring(0,10);
>         System.out.println(method);
>         String to = inputData.substring(10,74);
>         String value = inputData.substring(74);
>         Method refMethod = TypeDecoder.class.getDeclaredMethod("decode",String.class,int.class,Class.class);
>         refMethod.setAccessible(true);
>         Address address = (Address)refMethod.invoke(null,to,0,Address.class);
>         System.out.println(address.toString());
>         Uint256 amount = (Uint256) refMethod.invoke(null,value,0,Uint256.class);
>         System.out.println(amount.getValue());
> ```

@ghqgfj 

If I wanted to pass it to string it would be like that?

String cadena = (String)refMethod.invoke(null,to,0,String.class);

dekinsq:
public static List<Type> decodeInputData(String inputData,
                                             String methodName,
                                             List<TypeReference<?>> outputParameters) {
        Function function = new Function(methodName,
                Collections.<Type>emptyList(),
[...]

PR #1782, opened 2022-09-28
fix: make very useful method decode(...) public

### LINKED ISSUE #1591: Why are those decode methods in class TypeDecoder not public?
```java
    @SuppressWarnings("unchecked")
    static <T extends Type> T decode(String input, int offset, Class<T> type) {
        if (NumericType.class.isAssignableFrom(type)) {
            return (T) decodeNumeric(input.substring(offset), (Class<NumericType>) type);
        } else if (Address.class.isAssignableFrom(type)) {
            return (T) decodeAddress(input.substring(offset));
        } else if (Bool.class.isAssignableFrom(type)) {
            return (T) decodeBool(input, offset);
        } else if (Bytes.class.isAssignableFrom(type)) {
            return (T) decodeBytes(input, offset, (Class<Bytes>) type);
        } else if (DynamicBytes.class.isAssignableFrom(type)) {
            return (T) decodeDynamicBytes(input, offset);
        } else if (Utf8String.class.isAssignableFrom(type)) {
            return (T) decodeUtf8String(input, offset);
        } else if (Array.class.isAssignableFrom(type)) {
            throw new UnsupportedOperationException(
                    "Array types must be wrapped in a TypeReference");
        } else {
            throw new UnsupportedOperationException("Type cannot be encoded: " + type.getClass());
        }
    }
```
I have to use reflection when I want to decode the input data of an transaction.
https://github.com/web3j/web3j/issues/489
```java
    private <T extends Type> T decode(String input, int offset, Class<T> type) {
        try {
            Method refMethod = TypeDecoder.class.getDeclaredMethod("decode", String.class, int.class, Class.class);
            refMethod.setAccessible(true);
            return (T) refMethod.invoke(null, input, offset, type);
        } catch (Exception e) {
            throw new UnsupportedOperationException("Type cannot be encoded: " + type.getClass());
        }
    }
```
Or create a new class under the same package.
```java
package org.web3j.abi;

import org.web3j.abi.datatypes.Type;

public class TypeDecoderMethodExposer {
    public static <T extends Type> T decode(String input, int offset, Class<T> type){
        return TypeDecoder.decode(input, offset, type);
    }
}
```
But neither of them are elegant.
What's the purpose of making those methods non-public?

monperrus:
### What does this PR do?
fix: make very useful method  `decode(String input, int offset, Class<T> type)` public

Fix #1591

At the modifier change

### Why is it needed?
To have a better decoding API

### DIFF
--- a/abi/src/main/java/org/web3j/abi/TypeDecoder.java
+++ b/abi/src/main/java/org/web3j/abi/TypeDecoder.java
@@ -61,8 +61,9 @@ import static org.web3j.abi.Utils.staticStructNestedPublicFieldsFlatList;
  * documented, but is the reverse of the encoding details located <a
  * href="https://github.com/ethereum/wiki/wiki/Ethereum-Contract-ABI">here</a>.
  *
- * The public API is composed of "decode*" methods and provides backward-compatbility.
- * See https://github.com/web3j/web3j/issues/1591 for a discussion about decoding and possible improvements.
+ * <p>The public API is composed of "decode*" methods and provides backward-compatbility. See
+ * https://github.com/web3j/web3j/issues/1591 for a discussion about decoding and possible
+ * improvements.
  */
 public class TypeDecoder {
 
--- a/abi/src/main/java/org/web3j/abi/TypeDecoder.java
+++ b/abi/src/main/java/org/web3j/abi/TypeDecoder.java
@@ -118,15 +118,15 @@ public class TypeDecoder {
         }
     }
 
-    static <T extends Type> T decode(String input, Class<T> type) {
+    public static <T extends Type> T decode(String input, Class<T> type) {
         return decode(input, 0, type);
     }
 
-    static Address decodeAddress(String input) {
+    public static Address decodeAddress(String input) {
         return new Address(decodeNumeric(input, Uint160.class));
     }
 
-    static <T extends NumericType> T decodeNumeric(String input, Class<T> type) {
+    public static <T extends NumericType> T decodeNumeric(String input, Class<T> type) {
         try {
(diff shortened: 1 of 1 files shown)


rdovgan (approved):

hoanglamcoin (approved):

mohamedelshami:
Hi @monperrus, could you please provide more context on the scenario where you would use `decode(String input, int offset, Class<T> type)`? would you be able to achieve what you're after with [decodeFunctionResult](https://github.com/web3j/web3j/blob/master/abi/src/main/java/org/web3j/abi/DefaultFunctionReturnDecoder.java#L44)?

By design, this is package protected and encapsulated from public access.

monperrus:
The goal is to implement a `FunctionDecoder` for historical transaction inputs, as suggested by @iikirilov in https://github.com/web3j/web3j/pull/1206

It would also avoid using runtime reflection to bypass package visibility, as suggested by @ghqgfj to @githubityu in https://github.com/web3j/web3j/issues/489

sanguivore-easyco:
> Hi @monperrus, could you please provide more context on the scenario where you would use `decode(String input, int offset, Class<T> type)`? would you be able to achieve what you're after with [decodeFunctionResult](https://github.com/web3j/web3j/blob/master/abi/src/main/java/org/web3j/abi/DefaultFunctionReturnDecoder.java#L44)?
> 
> By design, this is package protected and encapsulated from public access.

This would also provide basic support for our need of decoding function inputs before the relevant transaction has been submitted to the blockchain.

andrii-kl (approved):

monperrus:
Thanks a lot @andrii-kl

Issue #1799, opened 2022-10-23
Error on hash computation on platform where UTF-8 is not the default charset


olivierayache:
Hello I think I have detected an issue here  https://github.com/web3j/web3j/blob/7dea3d99c5bdbfcc03aaeaa8575fb0c9a9a771ab/crypto/src/main/java/org/web3j/crypto/Sign.java#L60

In the code of method getEthereumMessagePrefix(int messageLength), the default charset is used rather than specifying UTF-8. This can produce wrong hash computation on platform where UTF-8 is not the default charset.

PR #1800, opened 2022-10-23
Update Sign.java

olivierayache:
### What does this PR do?

This pull request fixes issue #1799 by setting UTF-8 charset on "personal sign" message prefix computation. 

In the **getEthereumMessagePrefix(int messageLength)** method of **Sign.java**

### Why is it needed?

It is needed to correctly generate hash for personal_sign on platform where UTF-8 is not the default encoding


mohamedelshami:
Thanks, @olivierayache!

Merging this as soon as the build completes.

PR #1803, opened 2022-10-24
Add "accepted" block name

alexdupre:
As used by Avalanche: https://github.com/ava-labs/coreth/blob/master/rpc/types.go#L74


mohamedelshami:
thank you!

PR #1810, opened 2022-11-01
Handle generated `StaticArrayNN` in `Utils#getParameterizedTypeName`

ArtificialPB:
### What does this PR do?
Handle `StaticArrayNN` instances in `Utils#getParameterizedTypeName`.

/

### Why is it needed?
It seems that fixed-sized event parameters are generated using a normal `TypeReference` instead of `StaticArrayTypeReference`. This is handles such cases in `getParameterizedTypeName` method.

### DIFF
--- a/abi/src/test/java/org/web3j/abi/UtilsTest.java
+++ b/abi/src/test/java/org/web3j/abi/UtilsTest.java
@@ -28,10 +28,7 @@ import org.web3j.abi.datatypes.StaticArray;
 import org.web3j.abi.datatypes.Ufixed;
 import org.web3j.abi.datatypes.Uint;
 import org.web3j.abi.datatypes.Utf8String;
-import org.web3j.abi.datatypes.generated.Int64;
-import org.web3j.abi.datatypes.generated.StaticArray2;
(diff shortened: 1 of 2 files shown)


abi/src/test/java/org/web3j/abi/UtilsTest.java
```diff
@@ -28,10 +28,7 @@
 import org.web3j.abi.datatypes.Ufixed;
 import org.web3j.abi.datatypes.Uint;
 import org.web3j.abi.datatypes.Utf8String;
-import org.web3j.abi.datatypes.generated.Int64;
-import org.web3j.abi.datatypes.generated.StaticArray2;
-import org.web3j.abi.datatypes.generated.Uint256;
-import org.web3j.abi.datatypes.generated.Uint64;
+import org.web3j.abi.datatypes.generated.*;
```
NickSneo, line 31: Avoid using wildcard import

NickSneo:
Created a new PR - https://github.com/web3j/web3j/pull/1873
Closing this PR since we were not able edit the current files due to lack of permissions.

Issue #1752, opened 2022-07-22
Allow parsing of padded numerical strings


antonydenyer:
# Padded numerical strings are no longer accepted

A bug was introduced in https://github.com/web3j/web3j/pull/1679

Whilst not producing a padded string may be desirable, you should certainly accept them. When calling `Numeric.decodeQuantity` an exception is now thrown if you pass a padded string. I think the [Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle) should be adhered to. For instance, an rpc endpoint may produce a padded string.

rodionlim:
@antonydenyer Has this issue been fixed with #1813? If so, can this be closed?

antonydenyer:
It looks like it should, but I haven't tested it.

PR #1813, opened 2022-11-08
Fixed leading zeroes Hex quantity not accepted

NickSneo:
### What does this PR do?
Allow parsing of padded numerical strings with leading zeroes

from decodeQuantity() function in "utils/src/main/java/org/web3j/utils/Numeric.java"

### Why is it needed?
Fixes issues - https://github.com/web3j/web3j/issues/1752

### DIFF
--- a/utils/src/main/java/org/web3j/utils/Numeric.java
+++ b/utils/src/main/java/org/web3j/utils/Numeric.java
@@ -47,11 +47,23 @@ public final class Numeric {
         if (!isValidHexQuantity(value)) {
             throw new MessageDecodingException("Value must be in format 0x[1-9]+[0-9]* or 0x0");
         }
+
         try {
-            return new BigInteger(value.substring(2), 16);
+            return new BigInteger(parsePaddedHexQuantity(value), 16);
         } catch (NumberFormatException e) {
             throw new MessageDecodingException("Negative ", e);
         }
+    }
+
+    private static String parsePaddedHexQuantity(String value) {
+        if (value.length() > 3 && value.charAt(2) == '0') {
+            for (int i = 2; i < value.length(); i++) {
+                if (value.charAt(i) != '0') {
+                    return value.substring(i);
+                }
+            }
+        }
+        return value;
     }
 
     private static boolean isLongValue(String value) {
@@ -73,10 +85,6 @@ public final class Numeric {
         }
 
         if (!value.startsWith(HEX_PREFIX)) {
-            return false;
-        }
-
-        if (value.length() > 3 && value.charAt(2) == '0') {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


utils/src/main/java/org/web3j/utils/Numeric.java
```diff
@@ -47,13 +47,25 @@ public static BigInteger decodeQuantity(String value) {
...
+
         try {
-            return new BigInteger(value.substring(2), 16);
+            return new BigInteger(parsePaddedHexQuantity(value), 16);
         } catch (NumberFormatException e) {
             throw new MessageDecodingException("Negative ", e);
         }
     }
 
+    private static String parsePaddedHexQuantity(String value) {
```
andrii-kl, line 58: I suggest to make simpler solution 
```
public static BigInteger parsePaddedNumberHex(String value) {
        String numWithoutLeadingZeros = cleanHexPrefix(value).replaceFirst("^0+", "");
        return numWithoutLeadingZeros.isEmpty() ? BigInteger.ZERO : new BigInteger(numWithoutLeadingZeros, 16);
    }
```

utils/src/test/java/org/web3j/utils/NumericTest.java
```diff
@@ -72,14 +72,8 @@ public void testQuantityDecode() {
...
-        assertThrows(MessageDecodingException.class, () -> Numeric.decodeQuantity("0x0400"));
-        assertThrows(MessageDecodingException.class, () -> Numeric.decodeQuantity("0x001"));
-    }
-
-    @Test
-    public void testQuantityDecodeLeadingZeroException() {
-
-        assertThrows(MessageDecodingException.class, () -> Numeric.decodeQuantity("0x0400"));
+        assertEquals(Numeric.decodeQuantity("0x0400"), (BigInteger.valueOf(1024L)));
+        assertEquals(Numeric.decodeQuantity("0x001"), (BigInteger.valueOf(1L)));
```
andrii-kl, line 76: I would handle more corner cases. 
For example: 0x000

utils/src/main/java/org/web3j/utils/Numeric.java
NickSneo: Yeah seems Regex could have been a better approach here for replacing leading zeros

NickSneo: Yeah coded purely in java for simple solution, but can make it more simpler using regex

utils/src/test/java/org/web3j/utils/NumericTest.java
NickSneo: yeah added one more corner case "0x00f"

andrii-kl (approved):

PR #1823, opened 2022-11-26
Fix Bug: When length < input.length, the output is wrong, because the…

Michael-Ge:
… number of times in the for loop is length - offset, and the correct one should be length.

### What does this PR do?
Fix Bug

Numeric.java file line 242

### Why is it needed?
Fix Bug
When length < input.length, the output is wrong, because the number of times in the for loop is length - offset, and the correct one should be length.

### DIFF
--- a/utils/src/main/java/org/web3j/utils/Numeric.java
+++ b/utils/src/main/java/org/web3j/utils/Numeric.java
@@ -239,7 +239,8 @@ public final class Numeric {
 
     private static char[] toHexCharArray(byte[] input, int offset, int length, boolean withPrefix) {
         final char[] output = new char[length << 1];
-        for (int i = offset, j = 0; i < length; i++, j++) {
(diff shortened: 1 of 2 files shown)


utils/src/main/java/org/web3j/utils/Numeric.java
```diff
@@ -239,7 +239,8 @@ public static String toHexString(byte[] input, int offset, int length, boolean w
 
     private static char[] toHexCharArray(byte[] input, int offset, int length, boolean withPrefix) {
         final char[] output = new char[length << 1];
-        for (int i = offset, j = 0; i < length; i++, j++) {
+        int end = length + offset;
```
andrii-kl, line 242: Could you please cover this issue with small unit test?

gtebrean (approved):

PR #1827, opened 2022-12-07
Fix offset miscalculation in nested struct

bournezjc:
### What does this PR do?
Fix a bug that existed in decoding dynamic struct:
- correct offset miscalculation in nested struct 

In the **decodeDynamicStructElements(final String input, final int offset, final TypeReference<T> typeReference, final BiFunction<List<T>, String, T> consumer)** method of **TypeDecoder.java**

### Why is it needed?
More and more users start to use nested struct in Solidity. It turns out inputdata/returndata decoder is becoming a common method.

### DIFF
--- a/abi/src/main/java/org/web3j/abi/TypeDecoder.java
+++ b/abi/src/main/java/org/web3j/abi/TypeDecoder.java
@@ -490,7 +490,7 @@ public class TypeDecoder {
                                         0,
                                         TypeReference.create(declaredField));
                         staticOffset +=
-                                staticStructNestedPublicFieldsFlatList((Class<Type>) classType)
+                                staticStructNestedPublicFieldsFlatList((Class<Type>) declaredField)
                                                 .size()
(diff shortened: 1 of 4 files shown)


abi/src/main/java/org/web3j/abi/TypeDecoder.java
```diff
@@ -490,7 +490,7 @@ private static <T extends Type> T decodeDynamicStructElements(
                                         0,
                                         TypeReference.create(declaredField));
                         staticOffset +=
-                                staticStructNestedPublicFieldsFlatList((Class<Type>) classType)
+                                staticStructNestedPublicFieldsFlatList((Class<Type>) declaredField)
```
andrii-kl, line 493: Could you please add a unit test to cover this issue?

bournezjc: Sure. The new test cases have been added to both EncoderTest and DecoderTest.

gtebrean (approved):
LGTM

PR #1832, opened 2022-12-14
Feature: make EIP712 structured data a better public API

kberezin-p6:
### What does this PR do?
Makes it easier to use structure data API from Java code without manual JSON conversions.

`StructuredData.java`

### Why is it needed?
Makes it easier to use structure data API from Java code without manual JSON conversions.


gtebrean:
merged in https://github.com/web3j/web3j/pull/1896

PR #1834, opened 2022-12-22
Fix ETH feeHistory 

gtebrean:
### What does this PR do?
Fix argument type for feeHistory

Changed files

### Why is it needed?
One discord users mentioned that the argument type is wrong and fails on Hardhat. I tested and it seems as int it was working on Ganache, Geth, but on Hardhat is failing. Now should be fine for all as it is accordingly too documentation.
https://ethereum.github.io/execution-apis/api-documentation/

### DIFF
--- a/core/src/main/java/org/web3j/protocol/core/Ethereum.java
+++ b/core/src/main/java/org/web3j/protocol/core/Ethereum.java
@@ -112,7 +112,7 @@ public interface Ethereum {
     Request<?, EthMaxPriorityFeePerGas> ethMaxPriorityFeePerGas();
 
     Request<?, EthFeeHistory> ethFeeHistory(
-            int blockCount, DefaultBlockParameter newestBlock, List<Double> rewardPercentiles);
+            String blockCount, DefaultBlockParameter newestBlock, List<Double> rewardPercentiles);
 
     Request<?, EthAccounts> ethAccounts();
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/web3j/protocol/core/Ethereum.java
```diff
@@ -112,7 +112,7 @@ public interface Ethereum {
     Request<?, EthMaxPriorityFeePerGas> ethMaxPriorityFeePerGas();
 
     Request<?, EthFeeHistory> ethFeeHistory(
-            int blockCount, DefaultBlockParameter newestBlock, List<Double> rewardPercentiles);
+            String blockCount, DefaultBlockParameter newestBlock, List<Double> rewardPercentiles);
```
andrii-kl, line 115: I'm worried about backward compatibility, let's create a new method and marck old one as @Deprecated

core/src/main/java/org/web3j/protocol/core/JsonRpc2_0Web3j.java
```diff
@@ -234,7 +234,7 @@ public Request<?, EthMaxPriorityFeePerGas> ethMaxPriorityFeePerGas() {
 
     @Override
     public Request<?, EthFeeHistory> ethFeeHistory(
-            int blockCount, DefaultBlockParameter newestBlock, List<Double> rewardPercentiles) {
+            String blockCount, DefaultBlockParameter newestBlock, List<Double> rewardPercentiles) {
```
andrii-kl, line 237: the same

andrii-kl (approved):

PR #1856, opened 2023-02-01
Add dynamic gas provider implementations.

alexdupre:
For EIP-1559 transactions:
- gasLimit is calculated from gas estimation (+ extraGas), transactions detected to fail are not submitted
- maxPriorityFeePerGas is obtained from node
- maxFeePerGas is calculated as baseFee * 2 + maxPriorityFeePerGas

For Legacy transactions:
- gasLimit is calculated from gas estimation (+ extraGas), transactions detected to fail are not submitted
- gasPrice is obtained from node

Both implementations allow to set fallback values and price/limit caps.

### What does this PR do?
It adds two new ContractGasProvider implementations that allow to dynamically set gasLimit, gasPrice, maxPriorityFeePerGas and maxBaseFeePerGas, as almost any web3 library does by default.

### Why is it needed?
To pay less fees and to increase the probability for a transaction to be included in a block.

### DIFF
--- a/core/src/main/java/org/web3j/tx/gas/DynamicGasProvider.java
+++ b/core/src/main/java/org/web3j/tx/gas/DynamicGasProvider.java
@@ -24,6 +24,7 @@ public class DynamicGasProvider extends StaticGasProvider {
 
     private BigInteger extraGas;
     private BigInteger gasLimitCap;
+    private boolean throwErrorOnFailingTransactions;
     private Web3j web3j;
 
     public DynamicGasProvider(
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/web3j/tx/gas/DynamicGasProvider.java
```diff
@@ -0,0 +1,88 @@
...
+                        ? gasUsed.min(gasLimitCap)
+                        : gasUsed;
+            }
+        } catch (IOException e) {
+        }
+        return getGasLimit(contractFunc);
+    }
+
+    @Override
+    public BigInteger getGasPrice(
```
NickSneo, line 76: Why are these parameters fromAddress, contractAddress, etc required in this function?

alexdupre: They are not required in this specific implementation, but a custom implementation might decide to use such parameters to prioritize transactions sent by o to a specific address, for instance.

gtebrean (commented):
If I understood well you want to make these Dynamic Providers a generic provider to use it by default...so when EIP 1559 is not handled it will resume on the existing implementation that gets the gas in the standard way, right?
To obtain such a behaviour a factory or strategy patter should be use between the providers because If we handle the process like this we are braking a lot of code principles.

alexdupre:
> If I understood well you want to make these Dynamic Providers a generic provider to use it by default...so when EIP 1559 is not handled it will resume on the existing implementation that gets the gas in the standard way, right?

Not really, one should use the correct provider for the correct network. The fallback is between the Dynamic and the Static version, not between EIP 1559 and Legacy.

gtebrean:
> > If I understood well you want to make these Dynamic Providers a generic provider to use it by default...so when EIP 1559 is not handled it will resume on the existing implementation that gets the gas in the standard way, right?
> 
> Not really, one should use the correct provider for the correct network. The fallback is between the Dynamic and the Static version, not between EIP 1559 and Legacy.
> 
> Note: I'm not sure why you marked your comments as duplicated

After I finished the review they appeared duplicated in my browser.

gtebrean:
We apreciate your work and contributions to our libraries but after a debate inside the team we won't merge this as it brakes several OOP patterns programming principles in terms of implementations. 
We will add the support for the DynamicGasProvider in Web3J version 5. In there will be entire refactored and made more readable and eficient in term of usage. Thanks for letting us know that we should provide also these types of requests.

PR #1857, opened 2023-02-01
Better fix for #1834, use the correct block count encoding.

alexdupre:
PR #1834 duplicated the `ethFeeHistory` function and deprecated the old one. The correct thing to do would have been to just fix the block count encoding format of the existing one.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/core/Ethereum.java
+++ b/core/src/main/java/org/web3j/protocol/core/Ethereum.java
@@ -111,21 +111,8 @@ public interface Ethereum {
 
     Request<?, EthMaxPriorityFeePerGas> ethMaxPriorityFeePerGas();
 
-    @Deprecated
-    /**
-     * Method receives wrong type parameter int for blockCount, according to documentation the type
-     * should be String which is the encoded hex of the blocks number. This is kept for backward
-     * compatibility
-     *
-     * @param blockCount Requested range of blocks
-     * @param newestBlock Highest block of the requested range.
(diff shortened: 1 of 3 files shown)


gtebrean:
We deprecated that one to don't automatically force users to update to the new method. You can use the new method that receive int instead of string, the one that is deprecated will be removed in next major release. We didn't do your workaround because we want to make users aware that they need to send a hex now instead of int in blockCount.

alexdupre:
You deprecated the one with the `int` parameter, it doesn't make sense to force the user to pass a hex string for a block number, the issue was only in the internal encoding, not in the exposed API.

gtebrean:
exposed API should ask for the same parameters type as the core method that is called, in this case String:
https://ethereum.github.io/execution-apis/api-documentation/

alexdupre:
@gtebrean Following this reasoning, why don't you deprecate the `ethGetStorageAt`, `ethGetTransactionByBlockHashAndIndex`, `ethGetTransactionByBlockNumberAndIndex`, `ethGetUncleByBlockHashAndIndex`, `ethGetUncleByBlockNumberAndIndex`, `ethUninstallFilter`, `ethGetFilterChanges` and `ethGetFilterLogs` methods as well? They all accept a `BigInteger` instead of a `String`.

gtebrean:
We didn't have the time to do it, just kidding. 
Indeed after taking a closer look and with your mentions you are right. We were wrong on this one and we will merge your changes.

Issue #1858, opened 2023-02-02
Get StringIndexOutOfBoundsException when receive result as a dynamic array with single struct 


Gelassen:
# Title
I get ```StringIndexOutOfBoundsException``` on the call of ```function getMatches(address userFirst, address userSecond) external view override returns (Match[] memory)```. Research shows client receives data, but can not parse it properly.  

## Description
The function ```getMatches()``` returns dynamic array with ```struct``` items. A test case return a single item array. 

The client successfully receives encoded bytecode, but can failed to parse it with error: 
```
Exception(error=java.lang.StringIndexOutOfBoundsException: length=512; index=522084538)
```
As you might see the index is irrelevant. 

The encoded result is:
```
0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000100000000000000000000000062f8dc8a5c80db6e8fcc042f0cc54a298f8f2ffd000000000000000000000000000000000000000000000000000000000000000000000000000000000000000052e7400ba1b956b11394a5045f8bc3682792e1ac000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001
```
which match to array code, array size and match struct with params:
```
0x62F8DC8a5c80db6e8FCc042f0cC54a298F8F2FFd // 1st user
0, // 1st user token id
0x52E7400Ba1B956B11394a5045F8BC3682792E1AC, // 2nd user
1, // 2nd user token id 
0, // approve by 1st user
1  // approve by 2nd user
```

By debugging I figured out it fails on 128 offset where it tries to represent first item of the ```struct``` as a ```DynamicArray``` and this is a reason why it returns wrong offset.  

## Context
```
struct Match {
    address _userFirst;
    uint256 _valueOfFirstUser; 
    address _userSecond;
    uint256 _valueOfSecondUser; 
    bool _approvedByFirstUser;
    bool _approvedBySecondUser;
}
```
Relevant ```TypeDecoder``` code:
```
TypeDecoder.decodeDynamicStruct(
                                        input,
                                        offset + getDataOffset(input, currOffset, typeReference), // it will take "00000000000000000000000062f8dc8a5c80db6e8fcc042f0cc54a298f8f2ffd" as substring of input find out next offset
                                        TypeReference.create(cls)
);
```

Related PR: https://github.com/web3j/web3j/pull/1321
Related issues: 
[...]

PR #1860, opened 2023-02-03
Ref issues/1858: Add test case to cover scenario decode result of DynamicArray<DynamicStruct>

Gelassen:
Ref issues/1858: Add test case to cover scenario decode result of DynamicArray<DynamicStruct>. It confirms issue, but doesn't have solution yet.

### What does this PR do?
Add test case to FunctionReturnDecoder to spot known issue. Ref: https://github.com/web3j/web3j/issues/1858

Touched just several classes. This one probably would be a right start point: 
https://github.com/web3j/web3j/blob/199914cfffda72f3233abd72d4141fdc62b873d5/abi/src/test/java/org/web3j/abi/FunctionReturnDecoderTest.java

### Why is it needed?
Confirm issue in existing codebase, but complete solution would take much more work roughly estimated in 1 man-month work by discussion with former lib developer . Ref: https://github.com/web3j/web3j/issues/1858

Depends on this PR: https://github.com/web3j/web3j/pull/1861

### DIFF
--- a/abi/src/test/java/org/web3j/abi/AbiV2TestFixture.java
+++ b/abi/src/test/java/org/web3j/abi/AbiV2TestFixture.java
@@ -120,8 +120,6 @@ public class AbiV2TestFixture {
 
     // TODO Shall we reconsider const naming convention and unified constants here?
     // TODO I follow initial code convention, but would suggest FUNC_SET_MATCH naming style
-    public static final String FUNC_SETMATCH = "setArrayOfMatchStructFunction";
-
     public static final String FUNC_GETMATCH = "getArrayOfMatchStructFunction";
 
     public static class Foo extends DynamicStruct {
@@ -944,37 +942,7 @@ public class AbiV2TestFixture {
             this.approvedBySecondUser = approvedBySecondUser.getValue();
         }
 
-        @Override
-        public String toString() {
-            return "{ " +
-                    "\"userFirst\": \"${userFirst}\", " +
-                    "\"valueOfFirstUser\": \"${valueOfFirstUser}\", " +
-                    "\"userSecond\": \"${userSecond}\", " +
-                    "\"valueOfSecondUser\" : \"${valueOfSecondUser}\", " +
-                    "\"approvedByFirstUser\" : \"${approvedByFirstUser}\", " +
-                    "\"approvedBySecondUser\" : \"${approvedBySecondUser}\" " +
-                    "}";
-        }
     }
-
-    public static final Function setArrayOfMatchStructFunction =
-            new Function(
-                    FUNC_SETMATCH,
-                    Arrays.<Type>asList(
-                            new org.web3j.abi.datatypes.DynamicArray<Match>(
-                                    Match.class,
-                                    Collections.singletonList(
-                                            new Match(
-                                                    "0x62F8DC8a5c80db6e8FCc042f0cC54a298F8F2FFd",
-                                                    BigInteger.valueOf(0),
-                                                    "0x52E7400Ba1B956B11394a5045F8BC3682792E1AC",
-                                                    BigInteger.valueOf(1),
-                                                    false,
-                                                    false)
-                                    )
-                            )
-                    ),
-                    Collections.<TypeReference<?>>emptyList());
 
     public static final Function getArrayOfMatchStructFunction =
             new Function(
--- a/abi/src/test/java/org/web3j/abi/AbiV2TestFixture.java
+++ b/abi/src/test/java/org/web3j/abi/AbiV2TestFixture.java
@@ -19,6 +19,7 @@ import java.util.List;
 import java.util.stream.Collectors;
 
 import org.web3j.abi.datatypes.Address;
+import org.web3j.abi.datatypes.Bool;
 import org.web3j.abi.datatypes.DynamicArray;
 import org.web3j.abi.datatypes.DynamicBytes;
 import org.web3j.abi.datatypes.DynamicStruct;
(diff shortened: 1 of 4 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-bin.zip
```
SweeXordious, line 2: Is there a reason to update the gradle version? IMO, this should be handled in a separate PR if needed

local.properties
```diff
@@ -0,0 +1,8 @@
+## This file must *NOT* be checked into Version Control Systems,
```
SweeXordious, line 1: As specified here, this file shouldn't be added

abi/src/test/java/org/web3j/abi/FunctionReturnDecoderTest.java
```diff
@@ -1298,4 +1298,40 @@ public void testDecodeTupleOfStaticArrays() {
...
+    /**
+     * Add extra test case to cover discovered edge-case issue
+     *
+     * Ref: https://github.com/web3j/web3j/issues/1858
+     * */
+    @Test
+    public void testDecodeDynamicArrayOfMatches() {
+        String rawInput =
+                "0x"
+                        + "0000000000000000000000000000000000000000000000000000000000000020"  // not sure about this parameter - dynamic array identifier?
```
SweeXordious, line 1311: yes, this specifies the value `20` in hex which is `32` in decimal which is the offset of the start of the dynamic type

abi/src/test/java/org/web3j/abi/AbiV2TestFixture.java
```diff
@@ -883,4 +890,101 @@ public ArrayStruct(Uint256 id, DynamicArray<Address> addresses) {
...
+                    "\"valueOfFirstUser\": \"${valueOfFirstUser}\", " +
+                    "\"userSecond\": \"${userSecond}\", " +
+                    "\"valueOfSecondUser\" : \"${valueOfSecondUser}\", " +
+                    "\"approvedByFirstUser\" : \"${approvedByFirstUser}\", " +
+                    "\"approvedBySecondUser\" : \"${approvedBySecondUser}\" " +
+                    "}";
+        }
+    }
+
+    public static final Function setArrayOfMatchStructFunction =
```
SweeXordious, line 960: this function is not tested, better remove it IMO

SweeXordious (commented):
Thanks for giving this a shot. Is the test passing locally? or as expected, it's failing?

Gelassen:
> Thanks for giving this a shot. Is the test passing locally? or as expected, it's failing?

It fails as expected. Although the test case shows a different offset in the error, different from what I saw in my personal project.

gradle/wrapper/gradle-wrapper.properties
Gelassen: It is possible to find a workaround, but it would be better to update it. There is a compatibility issue between ```gradle``` and ```java``` versions, without resolving it I am not able even a compile a project. I don't use the latest ```JDK```, just one which is comfortable to work with on the rest of my projects. I believe it is a time to migrate to more recent ```gradle``` version. 

I will submit a new PR with change in gradle version, but change in this PR code will be left as is.

abi/src/test/java/org/web3j/abi/AbiV2TestFixture.java
```diff
@@ -117,6 +118,12 @@ public class AbiV2TestFixture {
     public static final String FUNC_setArrayOfStructWithArraysFunction =
             "setArrayOfStructWithArraysFunction";
 
+    // TODO Shall we reconsider const naming convention and unified constants here?
```
Gelassen, line 121: Please note ```//TODO``` I left here. What would be your opinion on this?

gradle/wrapper/gradle-wrapper.properties
Gelassen: Please, here is a separate PR with this change: https://github.com/web3j/web3j/pull/1861

abi/src/test/java/org/web3j/abi/AbiV2TestFixture.java
SweeXordious: Certainly you could, and that way is more readable, if you open a PR, I would :+1: for it, but would need one of the maintainers to approve it and approve the renaming.

Thanks a lot for the time you're giving the library.

Gelassen: You are welcome. Thank you for all your work, otherwise I have to start all work on the similar library since the beginning. :)

SweeXordious: Generally, if you already have your smart contract written and your structs defined, you could either fork the code and tweak it to work for your case, or you could even write an encoder/decoder for the specific structs you have from scratch easily.

Gelassen: Forking a codebase would lead to constant sync with origin repo. In case of multiple commits it means possible conflict resolution and parallel library maintaining. I thought about this, but left it as a last resort.

Thank you for point out on flexibility in customisation of type encoder\decoder. I will look at this. If you have this documented somewhere, it would be good to see links too.

SweeXordious: There is unfortunately no documentation for this, I haven't checked in a while tho.

But generally, I based my work on [this](https://docs.soliditylang.org/en/v0.8.17/abi-spec.html). If you manage to understand how the encoding/decoding works for static types and dynamic types, you could easily write an encoder/decoder specific to certain structs. That is because you already know what to expect and the fields offsets.

To test your knowledge on the offsets, check the tests scenarios [here](https://github.com/web3j/web3j/blob/master/abi/src/test/java/org/web3j/abi/FunctionReturnDecoderTest.java). They provide encodings + the corresponding structs. That should be a good starting point.

PR #1862, opened 2023-02-03
StaticStruct Method Signature Is Wrong

txgyy:
### What does this PR do?
*required*
if a function's parameter  like this, the function signature will be wrong,
solidity
```
function createAccount(
        address owner,
        Config memory config,
        uint256 salt
    ) public returns (address)

struct Config {
    address entryPoint;
    address aggregator;
    uint256[4] publicKey;
}
```
java
```
public class Config extends StaticStruct {

    public final Address entryPoint;

    public final Address aggregator;

    public final StaticArray4<Uint256> publicKey;

    public Config(Address entryPoint, Address aggregator, StaticArray4<Uint256> publicKey) {
        super(entryPoint, aggregator, publicKey);
        this.entryPoint = entryPoint;
        this.aggregator = aggregator;
        this.publicKey = publicKey;
    }
}
```

real method signature is **createAccount(address,(address,address,uint256[4]),uint256)**
but the **FunctionEncoder** return **createAccount(address,(address,address,StaticArray4),uint256)**
org/web3j/abi/datatypes/StaticStruct.java:41
### Why is it needed?
*required*
it's a bug

### DIFF
--- a/abi/src/main/java/org/web3j/abi/datatypes/StaticStruct.java
+++ b/abi/src/main/java/org/web3j/abi/datatypes/StaticStruct.java
@@ -38,7 +38,7 @@ public class StaticStruct extends StaticArray<Type> implements StructType {
         final StringBuilder type = new StringBuilder("(");
         for (int i = 0; i < itemTypes.size(); ++i) {
             final Class<Type> cls = itemTypes.get(i);
-            if (StructType.class.isAssignableFrom(cls)) {
+            if (StructType.class.isAssignableFrom(cls) || Array.class.isAssignableFrom(cls)) {
                 type.append(getValue().get(i).getTypeAsString());
             } else {
                 type.append(AbiTypes.getTypeAString(cls));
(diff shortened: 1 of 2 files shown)


abi/src/main/java/org/web3j/abi/datatypes/StaticStruct.java
```diff
@@ -38,7 +38,7 @@ public String getTypeAsString() {
         final StringBuilder type = new StringBuilder("(");
         for (int i = 0; i < itemTypes.size(); ++i) {
             final Class<Type> cls = itemTypes.get(i);
-            if (StructType.class.isAssignableFrom(cls)) {
+            if (StructType.class.isAssignableFrom(cls) || Array.class.isAssignableFrom(cls)) {
```
NickSneo, line 41: Plz add a test case also as mentioned in description for this

gradle.properties
```diff
@@ -1,2 +1,2 @@
 group=org.web3j
-version=4.9.6-SNAPSHOT
+version=4.9.6
```
NickSneo, line 2: No need to remove SNAPSHOT

NickSneo (changes requested):

gtebrean:
build is failing. Please run build locally first

txgyy:
> 

done

NickSneo (approved):

PR #1866, opened 2023-02-08
feat(#): add new eth method of eth_getBlockReceipts

SuperEdison:
### What does this PR do?
*required*
eth_getReceiptByTransactionHash need to batchRequest that's so bad so eth support a new eth method of eth_getBlockReipts, the method can't sub batchRequest
*required*
There is just a one file.
### Why is it needed?
*required*

convenient

### DIFF
--- a/core/src/main/java/org/web3j/protocol/core/methods/response/EthGetBlockReceipts.java
+++ b/core/src/main/java/org/web3j/protocol/core/methods/response/EthGetBlockReceipts.java
@@ -1,0 +1,20 @@
+package org.web3j.protocol.core.methods.response;
+
+import org.web3j.protocol.core.Response;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/web3j/protocol/core/methods/response/EthGetBlockReceipts.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * @author Edison
+ * @version 1.0
+ * <p></p>
+ * @date 2023/2/8
+ * @since 17
+ */
```
gtebrean, line 14: run spotlessApply locally to format your code before pushing

gtebrean (commented):
Seems good, run build locally before pushing it

gtebrean (approved):

PR #1867, opened 2023-02-09
Add new eth method of nr_getTokenMeta

wangxianghu:
### What does this PR do?
*Add new eth method of nr_getTokenMeta*

*required*

### Why is it needed?
*required*


gtebrean:
Thanks for your work but unfortunately we won't merge this in to master because we can't support non-core API methods. We are supporting only core ethereum API methods and this is not part of them.

PR #1868, opened 2023-02-16
fix subscription

gtebrean:
### What does this PR do?
Fix subscription mecanism to network events

Changed files

### Why is it needed?
Users are not able to retrieve subscription id due to small bug.


NickSneo (approved):
LGTM

Issue #1839, opened 2022-12-27
"Filter not found" detection could be improved


arobertn:
If, for example, the web3 RPC node web3j is currently polling restarts, the filter can be lost. `Filter.java` tries to detect this condition by matching the RPC error code against -32000, which is "invalid input". However in practice various Ethereum node implementations are behaving differently here. E.g., Ganache is returning `-32700` (see log below). The current Ethereum JSON-RPC specification unfortunately simply doesn't specify which codes should be returned for different situations for each method and so implementers roll their own. In view of this, it might be better to detect "filter not found" by error message string comparison which, though imperfect, is likely to catch more cases with fewer false alarms than relying on `-32000`.

```
15:13:53.029 --> POST http://127.0.0.1:8545/
15:13:53.029 Content-Type: application/json; charset=utf-8
15:13:53.029 Content-Length: 74
15:13:53.029 
15:13:53.029 {"jsonrpc":"2.0","method":"eth_getFilterChanges","params":["0x1"],"id":55}
15:13:53.029 --> END POST (74-byte body)
15:13:53.034 <-- 200 http://127.0.0.1:8545/ (4ms)
15:13:53.034 Content-Type: application/json
15:13:53.034 Content-Length: 1450
15:13:53.034 
15:13:53.034 {"id":55,"jsonrpc":"2.0","error":{"message":"filter not found","stack":"Error: filter not found\n    at EthereumApi.eth_getFilterChanges (.../ganache/dist/node/1.js:2:52758)","code":-32700}}
15:13:53.034 <-- END HTTP (1450-byte body)

15:13:53.034 WARN  o.w.p.c.filters.Filter   - Error sending request
org.web3j.protocol.core.filters.FilterException: Invalid request: filter not found
        at org.web3j.protocol.core.filters.Filter.throwException(Filter.java:194)
        at org.web3j.protocol.core.filters.Filter.pollFilter(Filter.java:144)
        at org.web3j.protocol.core.filters.Filter.lambda$run$0(Filter.java:92)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305)
        at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
[...]

PR #1869, opened 2023-02-22
"Filter not found" detection improved

mattiamonari:
### What does this PR do?
*Add a condition on the EthFilter polling.*

*There is just a one file.*

### Why is it needed?
*As pointed out in Issue #1839, every client returns a different error code to notify the loss of a filter. Thus, we can check firstly the error code, and if that is not matched, we can check the error message and see if it matches the string "filter not found". I used regex to create a case insensitve and multi-space pattern in order to incude as many cases as possible.*

### DIFF
--- a/core/src/main/java/org/web3j/protocol/core/filters/Filter.java
+++ b/core/src/main/java/org/web3j/protocol/core/filters/Filter.java
@@ -143,9 +143,8 @@ public abstract class Filter<T> {
                     reinstallFilter();
                     break;
                 default:
-                    if (Pattern.compile("(?i)\\bfilter\\s+not\\s+found\\b").matcher(message).find())
-                        reinstallFilter();
-                    else throwException(error);
+                    Pattern.compile("FILTER_NOT_FOUND_PATTERN").matcher(message).find())? 
+                    reinstallFilter():throwException(error);
                     break;
             }
         } else {
--- a/core/src/main/java/org/web3j/protocol/core/filters/Filter.java
+++ b/core/src/main/java/org/web3j/protocol/core/filters/Filter.java
@@ -50,6 +50,8 @@ public abstract class Filter<T> {
 
     private long blockTime;
 
+    private static final String FILTER_NOT_FOUND_PATTERN = "(?i)\\bfilter\\s+not\\s+found\\b";
+
     public Filter(Web3j web3j, Callback<T> callback) {
         this.web3j = web3j;
         this.callback = callback;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/web3j/protocol/core/filters/Filter.java
```diff
@@ -136,12 +137,15 @@ private void pollFilter(EthFilter ethFilter) {
...
+            String message = error.getMessage();
             switch (error.getCode()) {
                 case RpcErrors.FILTER_NOT_FOUND:
                     reinstallFilter();
                     break;
                 default:
-                    throwException(error);
+                    if (Pattern.compile("(?i)\\bfilter\\s+not\\s+found\\b").matcher(message).find())
+                        reinstallFilter();
+                    else throwException(error);
```
gtebrean, line 148: Add regex to a final static var and also:

```suggestion
                    Pattern.compile("FILTER_NOT_FOUND_PATTERN").matcher(message).find())? 
                    reinstallFilter():throwException(error);
```

mattiamonari: Hey @gtebrean, should I commit your changes to my branch and then push it in order to get my PR approved? Sorry, but I'm new to the open source world.

gtebrean: Add regex to a final static var  as I mentioned and after implement according to my suggestion ....here in this branch is good

gtebrean: also run build locally before push

mattiamonari: Initially I thought your suggestions were already implemented by you in the code, that's why I commited. 
Anyway, thanks for help. I'll make the right modifications and push.

mattiamonari: Sorry again..

mattiamonari: @gtebrean Ternary operator can't be used in this case, since [it would be used as a statement, not as an assignment](https://stackoverflow.com/a/51197166/10276248). Eventually it also generates side effects, thus it should be better to leave it in the normal if-else structure.

core/src/main/java/org/web3j/protocol/core/filters/Filter.java
```diff
@@ -136,12 +139,15 @@ private void pollFilter(EthFilter ethFilter) {
...
+            String message = error.getMessage();
             switch (error.getCode()) {
                 case RpcErrors.FILTER_NOT_FOUND:
                     reinstallFilter();
                     break;
                 default:
-                    throwException(error);
+                    if (Pattern.compile(FILTER_NOT_FOUND_PATTERN).matcher(message).find())
+                        reinstallFilter();
+                    else throwException(error);
```
gtebrean, line 150:
```suggestion
                     Pattern.compile(FILTER_NOT_FOUND_PATTERN).matcher(message).find())? 
                    reinstallFilter():throwException(error);
```

Also run build locally before push

mattiamonari: As I pointed out [above](https://github.com/web3j/web3j/pull/1869#discussion_r1121893565) it is not possible to use ternary operator, build will fail.
Also, I should have built locally with _./gradlew check_ and I've got no errors with last commit.

Issue #1781, opened 2022-09-28
Conflict for Events with same name for generated abi


ShawnWu0x:
# _Bug_title_
In Solidity, the events may have same names, but with different parameters.
Such as Compound's [controller contract](https://etherscan.io/address/0xbafe01ff935c7305907c33bf824352ee5979b526#code)

There are 2  ActionPaused events with different parameters.

```
/// @notice Emitted when an action is paused globally
    event ActionPaused(string action, bool pauseState);

    /// @notice Emitted when an action is paused on a market
    event ActionPaused(CToken cToken, string action, bool pauseState);
```
If we generate the Java file with web3jgen for it, there will be 2 global parameters with name ACTIONPAUSED_EVENT in the generated JAVA file.
This is not allowed in Java.

```
public static final Event ACTIONPAUSED_EVENT = new Event("ActionPaused", 
            Arrays.<TypeReference<?>>asList(new TypeReference<Utf8String>() {}, new TypeReference<Bool>() {}));
    ;

    public static final Event ACTIONPAUSED_EVENT = new Event("ActionPaused", 
            Arrays.<TypeReference<?>>asList(new TypeReference<Address>() {}, new TypeReference<Utf8String>() {}, new TypeReference<Bool>() {}));
    ;

```

## Steps To Reproduce

1. Open Compound's etherscan link [controller contract](https://etherscan.io/address/0xbafe01ff935c7305907c33bf824352ee5979b526#code)
2. Copy abi and bin codes to 2 files.
3. Run `web3j generate solidity -a abc.abi -b abc.bin -o ./ -p abi `
  * -a is abi file
  * -b is bin file
  * -o is the directory where generate the file
  * -p is the package of java file

### Expected behavior
The generated java file should has no ERROR.

### Actual behavior
The generated java file built failed with ERROR.

## Environment
- Web3j core 4.8.7
- Java 1.8
- MacOS

## Additional context

- [etherscan](https://etherscan.io/address/0xbafe01ff935c7305907c33bf824352ee5979b526#code) references

mohamedelshami:
@ShawnWu0x investigating this

CJ42:
Any updates on this? Also encountering the same issue in https://github.com/lukso-network/lsp-smart-contracts/pull/469

mattiamonari:
How would you like to have the name of duplicated event @CJ42? I'm currently working on it.

It is possible to create in the wrapper a 'numbered list', e.g. : 

```
[...]

PR #1872, opened 2023-03-03
Add a condition on the duplicated event names

mattiamonari:
### What does this PR do?
Add a condition on the name of the events of the java wrapper

There is just a one file.

### Why is it needed?
Based on Issue #1781, for contracts which have '**overloaded**' events (same name, but different signatures) the Java Wrapper creates those events as different instances of the Event class, but with the same variable name, thus resulting in an error. This PR adds the `changeEventsName()` method which uses an HashMap to count the occurences of an event name, and for those with more than one occurence, it will change the name putting a number at its end, e.g. 

```
public static final Event ACTIONPAUSED_EVENT = new Event("ActionPaused",
            Arrays.<TypeReference<?>>asList(new TypeReference<Utf8String>() {}, new TypeReference<Bool>() {}));
    ;

    public static final Event ACTIONPAUSED1_EVENT = new Event("ActionPaused", 
            Arrays.<TypeReference<?>>asList(new TypeReference<Address>() {}, new TypeReference<Utf8String>() {}, new TypeReference<Bool>() {}));
    ;
```

### DIFF
--- a/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
+++ b/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
@@ -423,6 +423,7 @@ public class SolidityFunctionWrapper extends Generator {
             throws ClassNotFoundException {
 
         Set<String> duplicateFunctionNames = getDuplicateFunctionNames(functionDefinitions);
+        changeEventsName(functionDefinitions);
         List<MethodSpec> methodSpecs = new ArrayList<>();
         for (AbiDefinition functionDefinition : functionDefinitions) {
             if (functionDefinition.getType().equals(TYPE_FUNCTION)) {
@@ -434,6 +435,25 @@ public class SolidityFunctionWrapper extends Generator {
             }
         }
         return methodSpecs;
+    }
+
+    private void changeEventsName(List<AbiDefinition> functionDefinitions) {
+
+        Map<String, Integer> countMap = new HashMap<>();
+
+        for (AbiDefinition functionDefinition : functionDefinitions) {
+            if (TYPE_EVENT.equals(functionDefinition.getType())
+                    && functionDefinition.getName() != null) {
+                String s = functionDefinition.getName();
+                if (countMap.containsKey(s)) {
+                    int count = countMap.get(s);
+                    functionDefinition.setName(s + count);
+                    countMap.put(s, count + 1);
+                } else {
+                    countMap.put(s, 1);
+                }
+            }
+        }
     }
 
     private List<TypeSpec> buildStructTypes(final List<AbiDefinition> functionDefinitions)
@@ -723,6 +743,7 @@ public class SolidityFunctionWrapper extends Generator {
         Set<String> fieldNames = new HashSet<>();
         fieldNames.add(Contract.FUNC_DEPLOY);
         Set<String> duplicateFunctionNames = getDuplicateFunctionNames(functionDefinitions);
+        changeEventsName(functionDefinitions);
         if (!duplicateFunctionNames.isEmpty()) {
             System.out.println(
                     "\nWarning: Duplicate field(s) found: "
--- a/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
+++ b/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
@@ -402,9 +402,19 @@ public class SolidityFunctionWrapper extends Generator {
                 .build();
     }
 
-    private FieldSpec createEventDefinition(String name, List<NamedTypeName> parameters) {
+    private FieldSpec createEventDefinition(
+            String name,
+            List<NamedTypeName> parameters,
+            Map<String, Integer> eventsCount,
+            AbiDefinition event) {
 
         CodeBlock initializer = buildVariableLengthEventInitializer(name, parameters);
+        Integer occurrences = eventsCount.get(name);
+        if (occurrences > 1) {
+            event.setName(name + (occurrences - 1));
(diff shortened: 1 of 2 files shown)


codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -436,6 +437,25 @@ private List<MethodSpec> buildFunctionDefinitions(
         return methodSpecs;
     }
 
+    private void changeEventsName(List<AbiDefinition> functionDefinitions) {
```
gtebrean, line 440:
```suggestion
    private void changeDuplicateEventsName(List<AbiDefinition> functionDefinitions) {
```

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -436,6 +437,25 @@ private List<MethodSpec> buildFunctionDefinitions(
...
+                String s = functionDefinition.getName();
+                if (countMap.containsKey(s)) {
+                    int count = countMap.get(s);
+                    functionDefinition.setName(s + count);
+                    countMap.put(s, count + 1);
+                } else {
+                    countMap.put(s, 1);
+                }
+            }
+        }
```
gtebrean, line 456:
```suggestion
        functionDefinitions.stream().filter(function -> TYPE_EVENT.equals(function.getType())
                && function.getName() != null).forEach(function -> {
            String functionName = function.getName();
            if (countMap.containsKey(functionName)) {
                int count = countMap.get(functionName);
                function.setName(functionName + count);
                countMap.put(functionName, count + 1);
            } else {
                countMap.put(functionName, 1);
            }
        });
```

gtebrean (commented):
This is a good workaround for now. Please add also a unit test to can include it in today release.

mattiamonari:
Added the unit test. Not sure if that's the proper way of doing it, please check. I had to put the buildFunctionDefinitions package-private since it was not accessible inside tests.

I also removed the usage of changeEventsName() in buildFuncNameConstants(), l.742 since the function operate just on TYPE_FUNCTION objects (correct me if I'm wrong).

mattiamonari:
Apparently my Unit Test is not working. On my local machine, anyway, it correctly builds and passes tests.

codegen/src/test/java/org/web3j/codegen/SolidityFunctionWrapperTest.java
```diff
@@ -902,6 +902,133 @@ public void testBuildFuncNameConstants() throws Exception {
...
+                        builder,
+                        Arrays.asList(firstEventDefinition, secondEventDefinition)));
+
+        String expected =
+                "class testClass {\n"
+                        + "  public static final org.web3j.abi.datatypes.Event EVENTNAME_EVENT = new org.web3j.abi.datatypes.Event(\"eventName\", \n"
+                        + "      java.util.Arrays.<org.web3j.abi.TypeReference<?>>asList(new org.web3j.abi.TypeReference<org.web3j.abi.datatypes.Utf8String>() {}, new org.web3j.abi.TypeReference<org.web3j.abi.datatypes.Bool>() {}));\n"
+                        + "  ;\n"
+                        + "\n"
+                        + "  public static final org.web3j.abi.datatypes.Event EVENTNAME1_EVENT = new org.web3j.abi.datatypes.Event(\"eventName1\", \n"
```
gtebrean, line 942: Event name shouldn't be modified  `new org.web3j.abi.datatypes.Event(\"eventName1\"...` will fail in request.

gtebrean:
I won't added it to this release as it affects the wrapper and has to be carefully tested also by us. 
We keep the PR open for finding a better solution.

codegen/src/test/java/org/web3j/codegen/SolidityFunctionWrapperTest.java
mattiamonari: Can you please give me some more details about that?
Do you mean it is not possible to use the abiDef.setName() function on TYPE_EVENTS objects?

Thanks

gtebrean:
> @gtebrean can you please give me some more information if you have them? With the new commit the eventName passed in the constructor should be the same. In my local env. the tests passed, but apparently there's still something which I'm doing wrongly since the checks failed.

jobs are complaining that:
org.web3j.codegen.SolidityFunctionWrapperTest > testBuildFunctionDuplicatedEventNames() FAILED
    org.opentest4j.AssertionFailedError at SolidityFunctionWrapperTest.java:1038.
    
It fails to create the wrapper, make sure you have latest changes from master.

mattiamonari:
> > @gtebrean can you please give me some more information if you have them? With the new commit the eventName passed in the constructor should be the same. In my local env. the tests passed, but apparently there's still something which I'm doing wrongly since the checks failed.
> 
> jobs are complaining that: org.web3j.codegen.SolidityFunctionWrapperTest > testBuildFunctionDuplicatedEventNames() FAILED org.opentest4j.AssertionFailedError at SolidityFunctionWrapperTest.java:1038.
> 
> It fails to create the wrapper, make sure you have latest changes from master.

My branch when I commited was up-to-date with master. org.web3j.codegen.SolidityFunctionWrapperTest > testBuildFunctionDuplicatedEventNames() is the unit test I created my self for this PR, but in my local env is passing the test, but not in the jobs. Which could be the cause? Error logs aren't really useful.

SergioIOB:
Hi! @mattiamonari @gtebrean 

I would like to know the status of this, as we have recently encountered this problem as well.

Best regards

gtebrean:
#2124 fixed this, thank you @mattiamonari for taking the initiative and @SergioIOB for fixing the pipeline

PR #1884, opened 2023-03-28
Change buildInputStream to fetch and retain response if includeRawReponse is true

cfelde:
..otherwise return unprocessed responseBody.byteStream().

This is inline with historical behavior, but due to later changes we started fetching and returning the content also when includeRawResponse was false. Due to this, also changing the default value of includeRawResponse to true.

### What does this PR do?
Clean up some historic issues and attempts to fix is so that expected behavior is achieved.

Testing should be done using integration testing and existing applications, to ensure behavior is as expected. Existing unit tests might not cover this.

### Why is it needed?
There have been issues reported by users due to the existing implementation not behaving as expected.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/http/HttpService.java
+++ b/core/src/main/java/org/web3j/protocol/http/HttpService.java
@@ -12,7 +12,6 @@
  */
 package org.web3j.protocol.http;
 
-import java.io.BufferedInputStream;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -29,8 +28,6 @@ import okhttp3.OkHttpClient;
 import okhttp3.RequestBody;
 import okhttp3.ResponseBody;
 import okhttp3.logging.HttpLoggingInterceptor;
-import okio.Buffer;
-import okio.BufferedSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/web3j/protocol/http/HttpService.java
```diff
@@ -185,9 +196,13 @@ protected void processHeaders(Headers headers) {
     private InputStream buildInputStream(ResponseBody responseBody) throws IOException {
         if (includeRawResponse) {
             return new ByteArrayInputStream(responseBody.bytes());
-        } else {
-            return responseBody.byteStream();
         }
+
+        BufferedSource source = responseBody.source();
+
+        source.request(Long.MAX_VALUE); // Buffer the body
```
cfelde, line 195: This will force the response to be loaded, which shouldn't be the case if includeRawResponse is false

core/src/main/java/org/web3j/protocol/http/HttpService.java
```diff
@@ -150,6 +153,14 @@ private static void configureLogging(OkHttpClient.Builder builder) {
         }
     }
 
+    public static void main(String[] args) throws IOException {
```
cfelde: This needs to be cleaned up

core/src/main/java/org/web3j/protocol/http/HttpService.java
gtebrean: then a solution will be to clone the responseBody.byteStream()

or refactor HttpsService (also for batch requests) and process them inside the try block

PR #1896, opened 2023-04-13
Structured data public api

gtebrean:
### What does this PR do?
extended from https://github.com/web3j/web3j/pull/1832 because user doesn't answer

changed files

### Why is it needed?
To have better access to Structured data api


cfelde:
LGTM

PR #1920, opened 2023-06-14
add the eth_baseFee call

rvullriede:
### What does this PR do?
Add another RPC call call "eth_baseFee". 

Ethereum.java

### Why is it needed?
Some EVM-Chains (e.g. Avalanche, see https://docs.avax.network/quickstart/transaction-fees) provide a dedicated RPC call that returns the current base fee. This allows web3j to use such networks as a evm-compatible client that is not restricted to Ethereum alone.


NickSneo:
@rvullriede Can you add some tests for your feature?

rvullriede:
@NickSneo I've added a basic test for this. I also looked at the TransactionDecoderTests but that would require adding the baseFee also to the Transaction1559 (which it technically is) and I'm not sure if you'd support such a bigger change since I assume your main focus is Ethereum, right?

NickSneo:
> @NickSneo I've added a basic test for this. I also looked at the TransactionDecoderTests but that would require adding the baseFee also to the Transaction1559 (which it technically is) and I'm not sure if you'd support such a bigger change since I assume your main focus is Ethereum, right?

I think this is good enough, as eth_baseFee api is exclusive to some evm-clients like Avalanche only

NickSneo (approved):
LGTM

NickSneo:
@rvullriede Plz add one more test in ResponseTest.java file, just like this - https://github.com/web3j/web3j/blob/99fb8f7512a110a61a1ca4ac6d3a9563c7ccbc46/core/src/test/java/org/web3j/protocol/core/ResponseTest.java#L191

rvullriede:
Thx for the hint! Done :-)

NickSneo:
> Thx for the hint! Done :-)

pipeline is failing, plz run spotlessApply. Thanks

rvullriede:
> > Thx for the hint! Done :-)
> 
> pipeline is failing, plz run spotlessApply. Thanks

Sorry for the noise. Done.

PR #1923, opened 2023-06-16
Update Class Loader

gtebrean:
### What does this PR do?
Updates Besu and Class Loader

*required*

### Why is it needed?
To can release CLI


gtebrean:
The reason for which I didn't add a unit test express for it is because the relevat tests are already in Generator and wrapper class. The issue that this is solving showed up only in the CLI repo, and causes the units  to fail there. Now with the current changes classes are processed correctly.

cfelde:
LGTM

PR #1929, opened 2023-07-15
Add Bloom filter utility

mhrsalehi:
## What does this PR do?

this PR adds the Ethereum Bloom filter utility class. that can be used to test a data (topic) for its presence and also to create a Bloom filter.

## Why is it needed?

this is a extremely useful feature that is missing from web3j. others also requested this feature (https://github.com/web3j/web3j/discussions/1795).
so I decided to implement it.
it is implemented based on Ethereum yellow paper.
I tested it extensively(I also hard-coded an Ethereum block LogsBloom data and its logs into my tests because it is great test case)

## Where should the reviewer start?
`org.web3j.utils.Bloom`

no side effect, and no file was edited. only two were added.


NickSneo:
@mhrsalehi Your build is failing, kindly build it in your local env first.
You might need to Run './gradlew :spotlessApply' to fix these violations, also fix other violations if any

PR #1958, opened 2023-09-03
Fix integration tests

NickSneo:
### What does this PR do?
Fix integration tests

review all files except in folder `integration-tests/src/test/resources/quorum-test-network`

### Why is it needed?
Currently many test were disabled as they were failing and many deprecated tests were included which are needed to be removed


integration-tests/src/test/java/org/web3j/protocol/besu/BesuPrivacyQuickstartIntegrationTest.java
```diff
@@ -65,32 +61,47 @@ public class BesuPrivacyQuickstartIntegrationTest {
...
     @Test
     public void testConnection() throws IOException {
-        assertEquals(nodeAlice.web3ClientVersion().send().getWeb3ClientVersion(), (CLIENT_VERSION));
-        assertEquals(nodeBob.web3ClientVersion().send().getWeb3ClientVersion(), (CLIENT_VERSION));
-        assertEquals(
-                nodeCharlie.web3ClientVersion().send().getWeb3ClientVersion(), (CLIENT_VERSION));
+        assertNotNull(rpcNode.web3ClientVersion().send().getWeb3ClientVersion());
+        assertNotNull(nodeAlice.web3ClientVersion().send().getWeb3ClientVersion());
+        assertNotNull(nodeBob.web3ClientVersion().send().getWeb3ClientVersion());
+        assertNotNull(nodeCharlie.web3ClientVersion().send().getWeb3ClientVersion());
```
gtebrean, line 104: it would be nice to assert also the chain ID

integration-tests/src/test/java/org/web3j/protocol/core/CoreIT.java
```diff
@@ -413,8 +403,9 @@ public void testEthGetUncleByBlockHashAndIndex() throws Exception {
     }
 
     @Test
-    @Disabled
+    @Disabled("Specified block has no uncle blocks, so result returning as null")
```
gtebrean, line 406: could this be enabled if we set a data folder for the network?

integration-tests/src/test/java/org/web3j/protocol/geth/GethIT.java
```diff
@@ -41,14 +40,14 @@ public void setUp() {
...
 
-    @Disabled
     @Test // Method not Implemented.
     public void testWeb3ClientVersion() throws Exception {
         TxPoolContent content = web3j.txPoolContent().send();
         assertNotNull(content.getResult());
     }
 
     @Test
+    @Disabled("Deprecated. Replaced by Clef's equivalent method - 'account_new'")
```
gtebrean, line 50: but is it failing on geth as you disabled it?

integration-tests/src/test/resources/quorum-test-network/.env
```diff
@@ -0,0 +1,22 @@
+# This file defines environment variables defaults for Docker-compose
+# but we also use it for shell scripts as a sourced file
+
+BESU_VERSION=23.4.1
```
gtebrean, line 4: why not 23.7.0?

gtebrean (commented):
The config files were uploaded from the local network generated files?

integration-tests/src/test/java/org/web3j/protocol/besu/BesuPrivacyQuickstartIntegrationTest.java
NickSneo: okay will add assert for chain ID too

NickSneo:
> The config files were uploaded from the local network generated files?

These configs are taken directly from Besu-dev-quickstart repo and as generated by npx besu-dev-quickstart script

gtebrean (approved):

Issue #1967, opened 2023-09-26
Structures without members are not encoded properly


Supertuba12:
# Structures without members are not encoded properly
The JSON RPC API specification for 'eth_signTypedData_v4' does not
explicitly state that the required struct 'EIP712Domain' has to define
any members [1]. The current implementation cannot correctly encode
such a structure, as it assumes that at least one member has been
parsed and added to the struct's string representation.

[1] https://eips.ethereum.org/EIPS/eip-712#specification-of-the-eth_signtypeddata-json-rpc

## Steps To Reproduce
Pass a valid structured data object where a type structure lacks a member:

```{json}
{
  "types": {
    "EIP712Domain": []
  },
  "primaryType": "EIP712Domain",
  "domain": {},
  "message": {}
}
```

### Expected behavior
The internal function `encodeStruct` should return `"EIP712Domain()"` when encoding `EIP712Domain`.
### Actual behavior
The internal function returns `"EIP712Domain)"`

## Environment
- Web3j version: First found using the latest android release `4.8.8-android`. Reproduced on master `ac0618c4d38ca780a7781d58255ff8b38626d7fc`

NickSneo:
Thanks for reporting and contributing.

NickSneo:
PR merged

PR #1968, opened 2023-09-26
Correctly encode structs without members

Supertuba12:
The JSON RPC API specification for 'eth_signTypedData_v4' does not explicitly state that the required struct 'EIP712Domain' has to define any members [1]. The current implementation cannot correctly encode such a structure, as it assumes that at least one member has been parsed and added to the struct's string representation.

[1] https://eips.ethereum.org/EIPS/eip-712#specification-of-the-eth_signtypeddata-json-rpc

### What does this PR do?
This commit refactors the struct encoding function such that it now accepts structures that does not define any members. This PR aims to resolve issue #1967.

There's only one function changed. I've also gone ahead and added a new test.

### Why is it needed?
This is needed to fully support the EIP-712 specification. Without it, web3j fails to properly handle valid types of structured data.

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
+++ b/crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
@@ -24,6 +24,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


NickSneo (approved):
LGTM

Issue #1726, opened 2022-06-12
Handle  ABI V2 nested array case


SweeXordious:
As discussed by @874341642 in https://github.com/web3j/web3j/issues/1321#issuecomment-1153054859, the following

SweeXordious:
The first step to solve this would be to add this ABI definition to the `AbiV2TestFixture.java` file, such as:
https://github.com/web3j/web3j/blob/a9cc039ad3da656049e8b7bc004656ba1113b40d/abi/src/test/java/org/web3j/abi/AbiV2TestFixture.java#L183-L228

This would add the above to the test cases we can work with.
Then, add type tests under: [web3j](https://github.com/web3j/web3j)/[abi](https://github.com/web3j/web3j/tree/master/abi)/[src](https://github.com/web3j/web3j/tree/master/abi/src)/[test](https://github.com/web3j/web3j/tree/master/abi/src/test)/[java](https://github.com/web3j/web3j/tree/master/abi/src/test/java)/[org](https://github.com/web3j/web3j/tree/master/abi/src/test/java/org)/[web3j](https://github.com/web3j/web3j/tree/master/abi/src/test/java/org/web3j)/[abi](https://github.com/web3j/web3j/tree/master/abi/src/test/java/org/web3j/abi).
You can be inspired by this PR: https://github.com/web3j/web3j/pull/1321.

After finishing with this, and having encoding decoding tests working.  Move on to the codegen module. And, following the same above way, add the tests and see where the problem is happening.

This PR should be your reference for changes https://github.com/web3j/web3j/pull/1321 as it is doing almost the same thing to implement support for struct arrays support.

874341642:
new TypeReference<DynamicArray<Foo>>() {}

I still don't understand what you mean, writing like this will throw an exception.

SweeXordious:
If I understand right, you will write your own Foo, call it something I don't know, that has the same ABI definition as the one above and test against it.

874341642:
> If I understand right, you will write your own Foo, call it something I don't know, that has the same ABI definition as the one above and test against it.

  public static class PlayerWork extends DynamicStruct {
        public String workType;

        public String careerAddr;

        public BigInteger startTime;

        public BigInteger endTime;

        public String owner;

        public List<BigInteger> blocks;

        public BigInteger beginTime;

[...]

PR #1972, opened 2023-10-12
decodeDynamicStructElements NPE fix

lzgh0102:
### What does this PR do?
*fix decodeDynamicStructElements NPE*
lower version codegen does not generate @Parameterized annotation when use DynamicArray<T> parameter, which will cause the error as below that confuse people
image

*use the test case: org.web3j.abi.FunctionReturnDecoderTest#testDecodeDynamicStruct3*
remove the Annotation can see the case
image

### Why is it needed?
*(https://github.com/web3j/web3j/issues/1726#issuecomment-1605250651*


NickSneo:
Seems the build is failing

PR #1977, opened 2023-10-31
Support public key recovery for EIP-155 transactions

jprudent:
### What does this PR do?
It supports recovery of public keys when the transaction use the anti replay protection (EIP 155)

PR is quite short 

### Why is it needed?
Most of txs are protected with EIP 155

### DIFF
--- a/utils/src/main/java/org/web3j/commons/ChainId.java
+++ b/utils/src/main/java/org/web3j/commons/ChainId.java
@@ -17,7 +17,6 @@ public enum ChainId {
     GOERLI(5),
     SEPOLIA(58008);
 
-
     private int id;
 
     ChainId(int id) {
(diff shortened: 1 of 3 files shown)


Fishin195 (approved):

utils/src/main/java/org/web3j/commons/ChainId.java
```diff
@@ -13,7 +13,10 @@
 package org.web3j.commons;
 
 public enum ChainId {
-    MAIN_NET(1);
+    MAIN_NET(1),
+    GOERLI(5),
+    SEPOLIA(58008);
```
NickSneo, line 18: Sepolia ChainId is 11155111?

utils/src/main/java/org/web3j/commons/ChainId.java
```diff
@@ -13,7 +13,10 @@
...
+
```
NickSneo, line 19: Plz add Holesky chainId too 17000

gtebrean:
closing due to lack of activity

Issue #1966, opened 2023-09-21
EIP-1559 for private transactions


NickSneo:
## Feature description
Adds support for EIP-1559 transactions via eth/eea_sendTransaction
encoding/decoding the EIP-1559 private transactions

As discussed here - https://discord.com/channels/779382027614158919/905428373218689054/1153588585044377602

NickSneo:
PR merged, closing

PR #1980, opened 2023-11-03
Adding support for EIP1559 Private Transactions

NickSneo:
### What does this PR do?
refactored private txs logic, to include more type of private txs
Adds support for EIP1559 private tx

Review all files

### Why is it needed?
https://github.com/web3j/web3j/issues/1966


besu/src/main/java/org/web3j/tx/PrivateTransactionManager.java
```diff
@@ -74,15 +74,15 @@ public PrivateTransactionManager(
             final Base64String privateFrom,
             final Base64String privacyGroupId,
             final Restriction restriction,
-            final TxSignService txSignService) {
+            final PrivateTxSignServiceImpl privateTxSignService) {
```
cfelde, line 77: Would this not cause API changes that might require this to be a major upgrade?

besu/src/main/java/org/web3j/tx/PrivateTransactionManager.java
```diff
@@ -134,15 +134,15 @@ public EthSendTransaction sendTransaction(
             throws IOException {
 
         final BigInteger nonce =
-                besu.privGetTransactionCount(txSignService.getAddress(), privacyGroupId)
+                besu.privGetTransactionCount(privateTxSignService.getAddress(), privacyGroupId)
                         .send()
                         .getTransactionCount();
 
-        final RawPrivateTransaction transaction;
+        final LegacyPrivateTransaction transaction;
```
cfelde, line 141: Would it not be less intrusive to keep the old class name?

besu/src/main/java/org/web3j/tx/PrivateTransactionManager.java
```diff
@@ -179,34 +179,34 @@ public EthSendTransaction sendEIP1559Transaction(
...
+                besu.privGetTransactionCount(privateTxSignService.getAddress(), privacyGroupId)
                         .send()
                         .getTransactionCount();
 
-        final RawPrivateTransaction transaction;
+        final PrivateTransaction1559 transaction;
         if (privateFor != null) {
             transaction =
-                    RawPrivateTransaction.createTransaction(
+                    new PrivateTransaction1559(
```
cfelde, line 189: It seem some places the pattern is LegacyPrivateTransaction.createTransaction(..) and here you're using a different pattern by calling the constructor directly on PrivateTransaction1559. Might be better to use the same pattern throughout, and if you go with the createTransaction(..) apporach, could you be able to support private 1559 transactions without breaking any existing APIs?

eea/src/main/java/org/web3j/protocol/eea/crypto/PrivateTransactionDecoder.java
```diff
@@ -30,25 +35,113 @@
...
+        else return PrivateTransactionType.PRIVATE_LEGACY;
+    }
+
+    private static LegacyPrivateTransaction decodePrivateTransaction1559(final byte[] transaction) {
+        final byte[] encodedTx = Arrays.copyOfRange(transaction, 1, transaction.length);
+        final RlpList rlpList = RlpDecoder.decode(encodedTx);
+        final RlpList temp = (RlpList) rlpList.getValues().get(0);
+        final List<RlpType> values = temp.getValues();
+
+        System.out.println(rlpList);
```
cfelde, line 62: I assume the sysout's are temporary? You might want to mark the PR as draft?

cfelde (changes requested):
Some questions around the approach, let's discuss.

eea/src/main/java/org/web3j/protocol/eea/crypto/PrivateTransactionDecoder.java
NickSneo: Fixed

besu/src/main/java/org/web3j/tx/PrivateTransactionManager.java
NickSneo: Fixed this, should not break this API

besu/src/main/java/org/web3j/tx/PrivateTransactionManager.java
NickSneo: Moved to .createTransaction(..) pattern everywhere

eea/src/main/java/org/web3j/protocol/eea/crypto/PrivateTransactionEncoder.java
```diff
@@ -61,35 +59,25 @@ public static byte[] encode(final RawPrivateTransaction rawTransaction, final lo
...
+        return encoded;
     }
 
     private static byte[] longToBytes(long x) {
         ByteBuffer buffer = ByteBuffer.allocate(Long.BYTES);
         buffer.putLong(x);
         return buffer.array();
     }
-
-    public static List<RlpType> asRlpValues(
```
cfelde, line 76: Any issue with this being removed for backwards compatibility?

eea/src/main/java/org/web3j/protocol/eea/crypto/transaction/type/PrivateTransaction1559.java
```diff
@@ -0,0 +1,260 @@
...
+import org.web3j.crypto.Sign;
+import org.web3j.crypto.transaction.type.ITransaction;
+import org.web3j.rlp.RlpString;
+import org.web3j.rlp.RlpType;
+import org.web3j.utils.Base64String;
+import org.web3j.utils.Bytes;
+import org.web3j.utils.Numeric;
+import org.web3j.utils.Restriction;
+
+public class PrivateTransaction1559 extends RawPrivateTransaction implements ITransaction {
```
cfelde, line 28: Is it not a bit odd that RawTransaction does not implement ITransaction, if PrivateTransaction1559 does? It seems like a bit of a broken pattern?

Genuine question: Why should PrivateTransaction1559 implement the interface is RawTransaction does not?

besu/src/main/java/org/web3j/protocol/besu/privacy/OnChainPrivacyTransactionBuilder.java
```diff
@@ -24,7 +24,7 @@
 import org.web3j.abi.datatypes.generated.Bytes32;
 import org.web3j.crypto.Credentials;
 import org.web3j.protocol.eea.crypto.PrivateTransactionEncoder;
-import org.web3j.protocol.eea.crypto.RawPrivateTransaction;
+import org.web3j.protocol.eea.crypto.transaction.type.RawPrivateTransaction;
```
cfelde, line 27: As we talked about, I guess this package change will break backwards compatibility?

cfelde (changes requested):

besu/src/main/java/org/web3j/protocol/besu/privacy/OnChainPrivacyTransactionBuilder.java
NickSneo: Yeah we can change the class path to original so that it doesn't break anything

eea/src/main/java/org/web3j/protocol/eea/crypto/PrivateTransactionEncoder.java
NickSneo: So every Tx type has its own different asRlpValues(), so it is being transferred from here.

gtebrean (approved):

jframe:
Hi @NickSneo, Thanks for adding this functionality to web3j, these are the changes we needed for web3signer

PR #1986, opened 2023-11-23
Enrich generateBothCallAndSend feature in TruffleJsonFunctionWrapperGenerator

tonykwok1992:
### What does this PR do?
- Enrich generateBothCallAndSend feature in TruffleJsonFunctionWrapperGenerator
- Refactor arg parsing logic using picoli rather than hand crafted parsing logic
- Make CommandLine ExceptionHandler re-throw rather than system exit so that test cases can catch failure case caused by exception

All files

### Why is it needed?
To be able to use `web3j generate truffle` with flag `--generateBoth` / `-B` to generator both send and call in wrapper generation using truffle option


NickSneo (approved):
LGTM

NickSneo:
@tonykwok1992 Please rebase to master and do spotlessApply

tonykwok1992:
rebased

codegen/src/main/java/org/web3j/codegen/TruffleJsonFunctionWrapperGenerator.java
```diff
@@ -40,6 +42,8 @@
 
 import static org.web3j.codegen.Console.exitError;
 import static org.web3j.utils.Collection.tail;
+import static picocli.CommandLine.*;
```
NickSneo, line 45: Plz remove this wildcard import

NickSneo (changes requested):

codegen/src/main/java/org/web3j/codegen/TruffleJsonFunctionWrapperGenerator.java
```diff
@@ -50,66 +54,107 @@
...
+        String jsonFileLocation;
+
+        @Option(
+                names = {"-o", "--outputDir"},
+                description = "destination base directory.",
+                required = true)
+        private String destinationDirLocation;
+
+        @Option(
+                names = {"-B", "--generateBoth"},
```
gtebrean, line 118: rename this to generateCallAndSend with "-cs"

tonykwok1992: I think it will be better to be consistent with `SolidityFunctionWrapperGenerator` which use `"-B", "--generateBoth"` ?

https://github.com/web3j/web3j/blob/master/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapperGenerator.java#L263

codegen/src/main/java/org/web3j/codegen/TruffleJsonFunctionWrapperGenerator.java
tonykwok1992: done

tonykwok1992:
rebased

gtebrean (approved):

Issue #1979, opened 2023-11-03
EIP-4844 support


Filter94:
## Feature description
[EIP-4844](https://eips.ethereum.org/EIPS/eip-4844) is coming up and Web3j doesn't support blob carrying transactions (EIP-2718 transaction type 0x03).

Here is an [example](https://github.com/Inphi/blob-utils/tree/master) implementation in Go.
Here is a [repo](https://github.com/ConsenSys/jc-kzg-4844) which implements JNI bindings for KZG stuff

PR #2000, opened 2024-01-16
EIP-4844 support

NickSneo:
### What does this PR do?
Adds support for EIP-4844

Review all files

### Why is it needed?
Closes https://github.com/web3j/web3j/issues/1979


crypto/src/main/java/org/web3j/crypto/BlobUtils.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+
+import ethereum.ckzg4844.CKZG4844JNI;
+
+import org.web3j.abi.datatypes.Bytes;
+
+public class BlobUtils {
```
gtebrean, line 24: There is an inconsistency in this util class related to static and non-static methods.
I would keep everything here as static method and send Blob every time as an argument when you run these helper methods. Else you might end having code as bellow, and it doesn't look good:
BlobUtils.method();
BlobUtils util = new BlobUtils(blob)

gtebrean (commented):
Few comments, also add  test cases for this.

besu/build.gradle
```diff
@@ -3,6 +3,10 @@ plugins {
 }
 description 'Besu JSON-RPC API'
 
+repositories {
+    maven { url "https://artifacts.consensys.net/public/maven/maven/" }
```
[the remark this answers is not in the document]
NickSneo: This is to fix the error I was getting when I am trying to run spotlessApply -
```
> Task :besu:test FAILED

Execution failed for task ':besu:test'.
> Could not resolve all files for configuration ':besu:testRuntimeClasspath'.
   > Could not find tech.pegasys:jc-kzg-4844:0.8.0.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/tech/pegasys/jc-kzg-4844/0.8.0/jc-kzg-4844-0.8.0.pom
       - https://plugins.gradle.org/m2/tech/pegasys/jc-kzg-4844/0.8.0/jc-kzg-4844-0.8.0.pom
       - https://jcenter.bintray.com/tech/pegasys/jc-kzg-4844/0.8.0/jc-kzg-4844-0.8.0.pom
       - https://oss.sonatype.org/content/repositories/releases/tech/pegasys/jc-kzg-4844/0.8.0/jc-kzg-4844-0.8.0.pom
       - https://oss.sonatype.org/content/repositories/snapshots/tech/pegasys/jc-kzg-4844/0.8.0/jc-kzg-4844-0.8.0.pom
     Required by:
         project :besu > project :core > project :crypto

Possible solution:
 - Declare repository providing the artifact, see the documentation at https://docs.gradle.org/current/userguide/declaring_repositories.html
```

Only build.graddle of crypto required this -
```
repositories {
    maven { url "https://artifacts.consensys.net/public/maven/maven/" }
}
```

But issue got resolved when I added the repo in other modules too

crypto/src/main/java/org/web3j/crypto/BlobUtils.java
```diff
@@ -0,0 +1,80 @@
...
+    public static void loadTrustedSetupFromResource() {
+        CKZG4844JNI.loadNativeLibrary();
+        try (InputStream resourceStream =
+                BlobUtils.class.getClassLoader().getResourceAsStream("trusted_setup.txt")) {
+            if (resourceStream == null) {
+                throw new IllegalArgumentException("Resource not found");
+            }
+            Path tempFile = Files.createTempFile("trusted_setup", "txt");
+            Files.copy(resourceStream, tempFile, StandardCopyOption.REPLACE_EXISTING);
+            CKZG4844JNI.loadTrustedSetup(tempFile.toString());
```
cfelde, line 46: Instead of using a temp file, look at using loadTrustedSetupFromResource

crypto/src/main/java/org/web3j/crypto/RawTransaction.java
```diff
@@ -118,6 +120,64 @@ public static RawTransaction createTransaction(
...
+            long chainId,
+            BigInteger nonce,
+            BigInteger maxPriorityFeePerGas,
+            BigInteger maxFeePerGas,
+            BigInteger gasLimit,
+            String to,
+            BigInteger value,
+            String data,
+            BigInteger maxFeePerBlobGas,
+            List<Bytes> versionedHashes) {
```
cfelde, line 162: If I provide the blobs, couldn't it calculate some of these other parameters for me, like versionedHashes and proofs? Or am I getting this wrong? See my comment below on Transaction4844.java as well.

crypto/src/main/java/org/web3j/crypto/transaction/type/Transaction4844.java
```diff
@@ -0,0 +1,293 @@
...
+        super(chainId, nonce, gasLimit, to, value, data, maxPriorityFeePerGas, maxFeePerGas);
+        this.maxFeePerBlobGas = maxFeePerBlobGas;
+        this.versionedHashes = versionedHashes;
+        this.blobs = Optional.ofNullable(blobs);
+        this.kzgCommitments = Optional.ofNullable(kzgCommitments);
+        this.kzgProofs = Optional.ofNullable(kzgProofs);
+    }
+
+    public Transaction4844(
+            List<Blob> blobsData,
```
cfelde, line 82: I see here it looks like we have a method of calculating the proofs and related based on the blob data, which was what my earlier question/comment was about. I think that makes sense, so just wanted to confirm my understanding.

cfelde (commented):
Added some minor comments/suggestions and questions.

I think test cases will confirm the question, specifically, the ability to submit a 4844 while just providing the blobs and have Web3j manage the proofs and related.

besu/build.gradle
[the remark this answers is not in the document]
NickSneo: Resolved

abi/src/main/java/org/web3j/abi/datatypes/Bytes.java
```diff
@@ -17,16 +17,16 @@ public class Bytes extends BytesType {
 
     public static final String TYPE_NAME = "bytes";
 
-    protected Bytes(int byteSize, byte[] value) {
+    public Bytes(int byteSize, byte[] value) {
         super(value, TYPE_NAME + value.length);
         if (!isValid(byteSize)) {
             throw new UnsupportedOperationException(
-                    "Input byte array must be in range 0 < M <= 32 and length must match type");
+                    "Input byte array must be in range 0 < M <= 48 and length must match type");
```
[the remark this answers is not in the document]
NickSneo: Change reverted, now using `org.apache.tuweni.bytes.Bytes`

crypto/src/main/java/org/web3j/crypto/BlobUtils.java
```diff
@@ -0,0 +1,80 @@
...
+public class BlobUtils {
+    public static boolean libraryLoaded = false;
+    private static final byte blobCommitmentVersionKZG = 0x01;
+    private final Blob blobData;
+
+    public BlobUtils(Blob blobData) {
+        if (!libraryLoaded) {
+            throw new RuntimeException(
+                    "Please load CKZG4844JNI Library by running BlobUtils.loadTrustedSetupFromResource()");
+        }
```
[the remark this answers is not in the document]
NickSneo: Should I remove loadLibrary logic? and just give a method for users to use that when doing blobTx?
This is from their readme, on how to use - 
```
// load the library once at startup
CKZG4844JNI.loadNativeLibrary();

// once a trusted setup is loaded, all methods will use it
CKZG4844JNI.loadTrustedSetup("trusted-setup.txt");

byte[] blob = ...;
byte[] commitment = CKZG4844JNI.blobToKzgCommitment(blob);
byte[] proof = CKZG4844JNI.computeBlobKzgProof(blob, commitment);

boolean isValidProof = CKZG4844JNI.verifyBlobKzgProof(blob, commitment, proof);

// the current trusted setup should be freed before a new one is loaded
CKZG4844JNI.freeTrustedSetup();
```

crypto/src/main/java/org/web3j/crypto/BlobUtils.java
NickSneo: Makes sense, thanks. Will implement all static methods

crypto/src/main/java/org/web3j/crypto/RawTransaction.java
NickSneo: Yes makes sense, but we are using this when decoding 4844 tx, that time we are decoding blobs, commitments, proofs and hashes. So we can directly decode and create a 4844 tx.

crypto/src/main/java/org/web3j/crypto/BlobUtils.java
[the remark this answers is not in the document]
gtebrean: I was referring to automatically loading it

crypto/src/main/java/org/web3j/crypto/BlobUtils.java
```diff
@@ -0,0 +1,70 @@
...
+    public static boolean checkProofValidity(Blob blobData, Bytes commitment, Bytes proof) {
+        return CKZG4844JNI.verifyBlobKzgProof(
+                blobData.data.toArray(), commitment.toArray(), proof.toArray());
+    }
+
+    public static Bytes kzgToVersionedHash(Bytes commitment) {
+        byte[] hash = Hash.sha256(commitment.toArray());
+        hash[0] = blobCommitmentVersionKZG;
+        return Bytes.wrap(hash);
+    }
```
abhijeetbhagat, line 64: does it make sense to use the respective array types in the function signatures instead since all we are doing is  converting the params to arrays. for .e.g.,
```
public static Bytes getCommitment(Array<whatever> param) {
```

crypto/src/main/java/org/web3j/crypto/Blob.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package org.web3j.crypto;
+
+import org.apache.tuweni.bytes.Bytes;
+
+public class Blob {
```
cfelde, line 17: This class should have hashCode and equals implemented so Blob instances can be compared based on data, ideally also some helpful toString

crypto/src/main/java/org/web3j/crypto/BlobUtils.java
```diff
@@ -0,0 +1,82 @@
...
+            final ByteBuffer g1 = ByteBuffer.allocate(g1Count * CKZG4844JNI.BYTES_PER_G1);
+            final ByteBuffer g2 = ByteBuffer.allocate(g2Count * CKZG4844JNI.BYTES_PER_G2);
+
+            for (int i = 0; i < g1Count; i++) {
+                g1.put(Bytes.fromHexString(reader.readLine()).toArray());
+            }
+            for (int i = 0; i < g2Count; i++) {
+                g2.put(Bytes.fromHexString(reader.readLine()).toArray());
+            }
+            CKZG4844JNI.loadTrustedSetup(g1.array(), g1Count, g2.array(), g2Count);
```
cfelde, line 52: Still not sure why you can't just use loadTrustedSetupFromResource?

crypto/src/main/java/org/web3j/crypto/BlobUtils.java
```diff
@@ -0,0 +1,82 @@
...
+
+public class BlobUtils {
+
+    static {
+        CKZG4844JNI.loadNativeLibrary();
+    }
+
+    private static final byte blobCommitmentVersionKZG = 0x01;
+
+    public static void loadTrustedSetupParameters() {
```
cfelde, line 34: If we call this once, do we ever need to call it again?

Could we not just call this after CKZG4844JNI.loadNativeLibrary() above, and never call freeTrustedSetup() ?

cfelde (changes requested):
Some questions for review/discussion

crypto/src/main/java/org/web3j/crypto/BlobUtils.java
NickSneo: I tried using this method `loadTrustedSetupFromResource` but couldn't get it working, getting error - File does not exist.
I gave absolute path too, but still same issue

crypto/src/main/java/org/web3j/crypto/BlobUtils.java
NickSneo: Yes sounds like a better approach

NickSneo: Given both option, BlobUtil will automatically load file in static method. If someone wants to use his own file, he can free the setup once and give path to the file

crypto/src/main/java/org/web3j/crypto/BlobUtils.java
NickSneo: Hey, thanks for the suggestion. Actually users might use the BlobUtils for single blob to get commitment, proof, etc So we want make it simple for them to use this.

build.gradle
```diff
@@ -21,10 +21,12 @@ ext {
     jnr_unixsocketVersion = '0.38.17'
     okhttpVersion = '4.9.0'
     rxjavaVersion = '2.2.2'
-    slf4jVersion = '1.7.30'
+    slf4jVersion = '2.0.11'
```
gtebrean, line 24: let's move to 2.0.12, is the latest one

gtebrean (commented):
👍

gtebrean (approved):

PR #2009, opened 2024-03-04
fix versionedHashes to blobVersionedHashes

thinkAfCod:
### What does this PR do?
The correct property name in the transaction object returned by the eth_getBlockByNumber Api should be 'blobVersionedHashes', not 'versionedHashed'.

This is a short PR

### Why is it needed?
The current version does not correctly parse these returned messages.


NickSneo (approved):
LGTM

PR #2016, opened 2024-03-15
Bug fix for large binary with unlink libraries codegen

tonykwok1992:
### What does this PR do?
Bug fix for large binary with unlinked libraries codegen

All files

### Why is it needed?
solidity bytecode with unlinked libraries can contain dollar sign (https://docs.soliditylang.org/en/latest/using-the-compiler.html) in the form of `__$hashed$__`. This could potentially being interpreted as string placeholder for args replacement in Java poet, in the case that cause failure in particular is `__$927c5a12e2f339676f56d42ec1c0537964$__` (the $927c part). The code that contain this bug only at the case for large binary (length > 65534). The safest way to fix is just to escape all dollar using javapoet's escape rule

https://github.com/square/javapoet/issues/670


codegen/src/test/java/org/web3j/codegen/SolidityFunctionWrapperTest.java
```diff
@@ -18,11 +18,7 @@
 import java.util.Collections;
 import java.util.List;
 
-import com.squareup.javapoet.ClassName;
-import com.squareup.javapoet.MethodSpec;
-import com.squareup.javapoet.ParameterizedTypeName;
-import com.squareup.javapoet.TypeName;
-import com.squareup.javapoet.TypeSpec;
+import com.squareup.javapoet.*;
```
gtebrean, line 21: please remove wildcard import

tonykwok1992: done

NickSneo (approved):
LGTM

PR #2017, opened 2024-03-19
fix contract generator

gtebrean:
### What does this PR do?
Fix contract wrapper generation

*required*

### Why is it needed?
*required*


NickSneo (approved):
LGTM

tonykwok1992:
Looks good but curious - what is the things that is broken without this change?

gtebrean:
> Looks good but curious - what is the things that is broken without this change?

will brake the generation of test contract wrappers....atm every method needs to be public in order for unit tests to be created properly

PR #2020, opened 2024-03-20
Fix for test java wrappers with duplicate name

NickSneo:
### What does this PR do?
Adds logic to fix bug where test java wrapper methods have duplicate name

review all files

### Why is it needed?
Bug fix for test java wrapper


codegen/src/test/java/org/web3j/codegen/unit/gen/java/MethodParserTest.java
```diff
@@ -27,7 +27,8 @@ public void testThatDeployMethodWasGenerated() {
         Optional<Method> deployMethod =
                 filteredMethods.stream().filter(m -> m.getName().equals("deploy")).findAny();
         MethodSpec deployMethodSpec =
```
gtebrean, line 29: add in this class a test for a smart contract that contains 2 methods with the same name...even erc721

codegen/src/main/java/org/web3j/codegen/unit/gen/MethodFilter.java
```diff
@@ -54,21 +56,38 @@ private static boolean parametersAreMatching(final Method method) {
...
                 .forEach(
-                        method ->
-                                listOfMethodSpecs.add(
-                                        new FunParser(method, theContract).getFunSpec()));
-        return listOfMethodSpecs;
+                        method -> {
+                            String baseName = method.getName();
+                            int count = functionNameCountMap.getOrDefault(baseName, 0);
+                            functionNameCountMap.put(baseName, count + 1);
+                            String uniqueName = count > 0 ? baseName + count : baseName;
```
gtebrean, line 85: please extract this common code to a method

gtebrean (changes requested):

codegen/src/test/java/org/web3j/codegen/unit/gen/java/MethodParserTest.java
NickSneo: ERC721 is hard to add in here, I have added duplicate functions in greeter contract, and added the test

gtebrean (approved):

PR #2025, opened 2024-04-01
Fix for test wrappers generation

NickSneo:
### What does this PR do?
Adds a check to skip test wrappers generation for getDeploymentBinary

review all

### Why is it needed?
Fixes test wrapper generation bug

## Checklist


gtebrean (commented):
A better approach would be to make that method private again and to avoid in here adding the private methods

gtebrean (approved):

PR #2045, opened 2024-04-23
Fix two typos in package org.web3j.abi.

hanmz:
### What does this PR do?
Fix two typos in package org.web3j.abi.
`compatbility` => `compatibility`
`errror` => `error`

### Why is it needed?
Improves readability


NickSneo (approved):
LGTM

PR #2060, opened 2024-06-06
Bug fix: Same inner struct name of different libraries result in failure in compilation for wrapper generated 

tonykwok1992:
### What does this PR do?
Bug fix: Same inner struct name of different libraries result in failure in compilation for wrapper generated. Example of such contract is in the test case 

All files

### Why is it needed?
It is a bug fix

## Checklist


NickSneo (approved):
LGTM

PR #2066, opened 2024-06-11
Remove registration of bouncy castle as JCE provider

killergerbah:
### What does this PR do?
This PR removes static registration of `BouncyCastleProvider` as a JCE provider.

In `Keys.java`, review the change for key pair generation to refer to `BouncyCastleProvider` by reference instead of by string.

### Why is it needed?
It prevents the usage of the `Keys` class from inadvertently changing the behavior of the depending application.

## Checklist

- [ x ] I've read the contribution guidelines.
- [  ] I've added tests (if applicable). (I am assuming that additional tests are not required because `Keys` is already well-tested)
- [  ] I've added a changelog entry if necessary.


NickSneo:
@killergerbah DCO is failing, can you signoff all the commits?

NickSneo (approved):
LGTM

PR #2068, opened 2024-06-14
Big fix for codegen when Array of struct and struct used together

tonykwok1992:
### What does this PR do?
 After fixing https://github.com/hyperledger/web3j/pull/2061, there is still an edge case where array of struct and struct are used together in the same contract e.g `Foo` and `Foo[2]`. Fix this by having `normalizeNamedType` to take care of static array

Also, structIdentifer is using hashCode which does not have uniqueness Property, it could subject to quite significant collision.

This should also fix https://github.com/hyperledger/web3j/issues/2056. Added test case for the example in the issue

All files

### Why is it needed?
Bug fix

## Checklist


core/src/main/java/org/web3j/protocol/core/methods/response/AbiDefinition.java
```diff
@@ -275,12 +275,11 @@ public void setComponents(final List<NamedType> components) {
             this.components = components;
         }
 
-        public int structIdentifier() {
+        public String structIdentifier() {
             return ((internalType == null ? type : internalType.isEmpty() ? type : internalType)
                             + components.stream()
-                                    .map(namedType -> String.valueOf(namedType.structIdentifier()))
-                                    .collect(Collectors.joining()))
-                    .hashCode();
```
tonykwok1992, line 283: hashCode does not have uniqueness Property, it could subject to quite significant collision as a identifier

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -637,7 +646,7 @@ private List<AbiDefinition.NamedType> extractStructs(
                             parameters.addAll(definition.getOutputs());
                             return parameters.stream()
                                     .map(this::normalizeNamedType)
-                                    .filter(namedType -> namedType.getType().startsWith("tuple"));
```
tonykwok1992, line 640: This was added in my previous PR merged https://github.com/hyperledger/web3j/pull/2061/files#diff-11944d4d901c6c86e0bb4f4bb1af18f9b3644a325a39e7cc7437878613330294R590

However, after fixing `normalizeNamedType` to handle static array, this is no longer needed

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -620,6 +617,18 @@ private NamedType normalizeNamedType(NamedType namedType) {
                             .getInternalType()
                             .substring(0, namedType.getInternalType().length() - 2),
                     namedType.isIndexed());
+        } else if (namedType.getType().startsWith("tuple[")
```
tonykwok1992, line 620: Main change is to add handling for static array in `normalizedType`

gtebrean:
Looks good from my side and also we really apreciate the effort! I'm not approving it yet because I want also to do some checks to make sure, but it should be fine as it is.

gtebrean (approved):

PR #2070, opened 2024-06-22
Bug fix for Int256 decode range [2^248, type(int256).max] and [ type(int256.min), -(2^248) )

tonykwok1992:
### What does this PR do?
Bug fix for decoding Int256 range [2^248, type(int256).max] and [ type(int256.min), -(2^248) )

All files

### Why is it needed?
The decoding of int256 number range  [2^248, type(int256).max] and [ type(int256.min), -(2^248) ) will throw exception,
type(int256.min) and type(int256).max is the particular case that we found

## Checklist

### DIFF
--- a/abi/src/main/java/org/web3j/abi/TypeDecoder.java
+++ b/abi/src/main/java/org/web3j/abi/TypeDecoder.java
@@ -141,17 +141,14 @@ public class TypeDecoder {
         try {
             byte[] inputByteArray = Numeric.hexStringToByteArray(input);
             int typeLengthAsBytes = getTypeLengthInBytes(type);
-
-            byte[] resultByteArray = new byte[typeLengthAsBytes + 1];
-
-            if (Int.class.isAssignableFrom(type) || Fixed.class.isAssignableFrom(type)) {
-                resultByteArray[0] = inputByteArray[0]; // take MSB as sign bit
-            }
-
             int valueOffset = Type.MAX_BYTE_LENGTH - typeLengthAsBytes;
-            System.arraycopy(inputByteArray, valueOffset, resultByteArray, 1, typeLengthAsBytes);
 
-            BigInteger numericValue = new BigInteger(resultByteArray);
+            BigInteger numericValue;
+            if (Uint.class.isAssignableFrom(type) || Ufixed.class.isAssignableFrom(type)) {
+                numericValue = new BigInteger(1, inputByteArray, valueOffset, typeLengthAsBytes);
+            } else {
+                numericValue = new BigInteger(inputByteArray, valueOffset, typeLengthAsBytes);
+            }
             return type.getConstructor(BigInteger.class).newInstance(numericValue);
 
         } catch (NoSuchMethodException
--- a/abi/src/test/java/org/web3j/abi/TypeDecoderTest.java
+++ b/abi/src/test/java/org/web3j/abi/TypeDecoderTest.java
@@ -97,6 +97,7 @@ import org.web3j.abi.datatypes.generated.Uint8;
 import org.web3j.abi.datatypes.generated.Uint80;
 import org.web3j.abi.datatypes.generated.Uint88;
 import org.web3j.abi.datatypes.generated.Uint96;
+import org.web3j.utils.Numeric;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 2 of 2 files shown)


abi/src/test/java/org/web3j/abi/TypeDecoderTest.java
```diff
@@ -858,11 +859,63 @@ public void testIntDecode() throws Exception {
...
+                        Int256.class),
+                new Int256(new BigInteger("-57896044618658097711785492504343953926634992332820282019728792003956564819968")));
+
         assertEquals(TypeDecoder.instantiateType("int", 123), (new Int(BigInteger.valueOf(123))));
 
         assertEquals(TypeDecoder.instantiateType("int", -123), (new Int(BigInteger.valueOf(-123))));
     }
 
+    @Test
+    public void testInt16All() throws Exception {
```
tonykwok1992, line 897: pick one type that won't take too long to run test to test the entire range for sanity

abi/src/main/java/org/web3j/abi/TypeDecoder.java
```diff
@@ -141,17 +141,14 @@ public static <T extends NumericType> T decodeNumeric(String input, Class<T> typ
         try {
             byte[] inputByteArray = Numeric.hexStringToByteArray(input);
             int typeLengthAsBytes = getTypeLengthInBytes(type);
-
-            byte[] resultByteArray = new byte[typeLengthAsBytes + 1];
-
-            if (Int.class.isAssignableFrom(type) || Fixed.class.isAssignableFrom(type)) {
-                resultByteArray[0] = inputByteArray[0]; // take MSB as sign bit
```
tonykwok1992, line 148: These contains bug for decoding int256 range mentioned in PR

abi/src/main/java/org/web3j/abi/TypeDecoder.java
```diff
@@ -141,17 +141,14 @@ public static <T extends NumericType> T decodeNumeric(String input, Class<T> typ
...
-            }
-
             int valueOffset = Type.MAX_BYTE_LENGTH - typeLengthAsBytes;
-            System.arraycopy(inputByteArray, valueOffset, resultByteArray, 1, typeLengthAsBytes);
 
-            BigInteger numericValue = new BigInteger(resultByteArray);
+            BigInteger numericValue;
+            if (Uint.class.isAssignableFrom(type) || Ufixed.class.isAssignableFrom(type)) {
+                numericValue = new BigInteger(1, inputByteArray, valueOffset, typeLengthAsBytes);
```
tonykwok1992, line 148: The BigInteger constructor already support reading offset from input byte array, do not need the array copying here

gtebrean:
@tonykwok1992 looks good from my side, thanks for bringing this up. Please update also the Changelog.md and I will merge this.

abi/src/main/java/org/web3j/abi/TypeDecoder.java
NickSneo: makes sense

abi/src/test/java/org/web3j/abi/TypeDecoderTest.java
NickSneo: So `2^255 - 1` and `-2^255` used to fail in decoding?

NickSneo: testing all Int16 and uint16 can be a overkill, add tests for boundary cases should be enough.
What do you think  @gtebrean ?

NickSneo (commented):
Left some comments

abi/src/test/java/org/web3j/abi/TypeDecoderTest.java
tonykwok1992: > testing all Int16 and uint16 can be a overkill, adding tests for boundary cases should be enough. What do you think @gtebrean ?

it is a loop with 2^16 = 65536 entries, I think it runs at a reasonable time. But if there is concern I can change it to boundary

tonykwok1992: > So `2^255 - 1` and `-2^255` used to fail in decoding?

yes, that's the case I'm using, later on found that the bug is in a large range mentioned in PR title

tonykwok1992:
> @tonykwok1992 looks good from my side, thanks for bringing this up. Please update also the Changelog.md and I will merge this.

added

abi/src/test/java/org/web3j/abi/TypeDecoderTest.java
tonykwok1992: done

NickSneo (approved):
LGTM

PR #2084, opened 2024-07-25
Bug fix for FastRawTransactionManager.resetNonce

### LINKED ISSUE #2002: FastRawTransactionManager's resetNonce() will bring its nonce field out of sync
# FastRawTransactionManager's function resetNonce() will bring its nonce field out of sync

## Steps To Reproduce
1. Using FastRawTransactionManager to instantiate a web3j contract instance.
MyContract myContractInstance = MyContract.load(myContractAddress, myWeb3jInstance, myFastRawTxManager, myGasProvider); 

2. Call resetNonce on the FastRawTransactionManager instance:
myFastRawTxManager.resetNonce();

3. Send a contract transaction:
RemoteFunctionCall<TransactionReceipt> rfc = myContractInstance.myMethod().send();

4. The result is that the transaction times out:
"Transaction receipt was not generated after 40 seconds for transaction: 0x21d063bb9057abc172bb7b590a3ea03d3a76eabc5409675fed961269a9d3f1e7"

### Expected behavior
The expectation is that the transaction will succeed.

### Actual behavior
Originally:
- the FastRawTransactionManager's nonce is "-1" 
- the network's TransactionCount for the current user is "n".

Calling resetNonce() will result in:
- the FastRawTransactionManager's nonce is "n" 
- the network's TransactionCount for the current user is "n".

Creating the TX will perform a call to:
```
    protected synchronized BigInteger getNonce() throws IOException {
        if (nonce.signum() == -1) {
            // obtain lock
            nonce = super.getNonce();
        } else {
            nonce = nonce.add(BigInteger.ONE);
        }
        return nonce;
    }
```
that will return an incremented nonce (one with value "n+1"), so the transaction will remain in the node's Transaction Pool until another transaction with a nonce with value "n" is sent. 
In the most scenario's this will occur only after:
- restarting the application containing the web3j instance;
or
- calling FastRawTransactionManager's setNonce (-1); // of course with casting to BigInteger
or
- calling FastRawTransactionManager's setNonce (n-1); // of course with casting to BigInteger

## Environment
_Describe the environment in which the issue occurs_
- Web3j 4.10.3
- Java 17
- ubuntu 22.04

## Additional context
_Add any other context about the problem here._
- Logs
- Sample code and/or code snippets
- Unit/integration tests to highlight the issue
- [etherscan](https://etherscan.io/) references

junsung-cho:
### What does this PR do?
Make `FastRawTransactionManager.resetNonce()` reset the nonce

#2002

#### AS-IS
```java
    public synchronized void resetNonce() throws IOException {
        nonce = super.getNonce();
    }
```
#### TO-BE
```java
    public synchronized void resetNonce() throws IOException {
        nonce = BigInteger.valueOf(-1);
    }
```

While the **AS-IS** reset nonce with the current nonce, if `getNonce()` is called, it will return `currentNonce + 1`.

```java
    @Override
    protected synchronized BigInteger getNonce() throws IOException {
        if (nonce.signum() == -1) {
            // obtain lock
            nonce = super.getNonce();
        } else {
            nonce = nonce.add(BigInteger.ONE); // this line will be called
        }
        return nonce;
    }
```

## Checklist

fix #2002

### DIFF
--- a/core/src/main/java/org/web3j/tx/FastRawTransactionManager.java
+++ b/core/src/main/java/org/web3j/tx/FastRawTransactionManager.java
@@ -73,7 +73,11 @@ public class FastRawTransactionManager extends RawTransactionManager {
     }
 
     public synchronized void resetNonce() throws IOException {
-        nonce = super.getNonce();
+        nonce = super.getNonce().subtract(BigInteger.ONE);
+    }
+
+    public synchronized void clearNonce() {
+        nonce = BigInteger.valueOf(-1);
     }
 
     public synchronized void setNonce(BigInteger value) {
--- a/integration-tests/src/test/java/org/web3j/protocol/scenarios/FastRawTransactionManagerIT.java
+++ b/integration-tests/src/test/java/org/web3j/protocol/scenarios/FastRawTransactionManagerIT.java
@@ -12,20 +12,9 @@
  */
 package org.web3j.protocol.scenarios;
 
-import java.math.BigDecimal;
-import java.math.BigInteger;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.Future;
-
 import com.carrotsearch.junitbenchmarks.BenchmarkOptions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
-
 import org.web3j.EVMTest;
 import org.web3j.NodeType;
 import org.web3j.protocol.Web3j;
@@ -38,9 +27,17 @@ import org.web3j.tx.response.PollingTransactionReceiptProcessor;
 import org.web3j.tx.response.QueuingTransactionReceiptProcessor;
 import org.web3j.utils.Convert;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.Future;
+
+import static org.junit.jupiter.api.Assertions.*;
 import static org.web3j.tx.TransactionManager.DEFAULT_POLLING_ATTEMPTS_PER_TX_HASH;
 
 @EVMTest(type = NodeType.BESU)
@@ -77,11 +74,11 @@ public class FastRawTransactionManagerIT extends Scenario {
         }
 
         for (int i = 0;
-                i < DEFAULT_POLLING_ATTEMPTS_PER_TX_HASH && !transactionReceipts.isEmpty();
-                i++) {
+             i < DEFAULT_POLLING_ATTEMPTS_PER_TX_HASH && !transactionReceipts.isEmpty();
+             i++) {
 
             for (Iterator<Future<TransactionReceipt>> iterator = transactionReceipts.iterator();
-                    iterator.hasNext(); ) {
+                 iterator.hasNext(); ) {
                 Future<TransactionReceipt> transactionReceiptFuture = iterator.next();
 
                 if (transactionReceiptFuture.isDone()) {
@@ -111,8 +108,16 @@ public class FastRawTransactionManagerIT extends Scenario {
 
         createTransaction(transfer, gasPrice).send();
         createTransaction(transfer, gasPrice).send();
+
+        BigInteger expected = transactionManager.getCurrentNonce();
+
         transactionManager.resetNonce();
+
+        BigInteger actual = transactionManager.getCurrentNonce();
+
         createTransaction(transfer, gasPrice).send();
+
+        assertEquals(expected, actual);
     }
 
     @Test
@@ -135,7 +140,8 @@ public class FastRawTransactionManagerIT extends Scenario {
                                     }
 
                                     @Override
-                                    public void exception(Exception exception) {}
+                                    public void exception(Exception exception) {
+                                    }
                                 },
                                 DEFAULT_POLLING_ATTEMPTS_PER_TX_HASH,
                                 POLLING_FREQUENCY));
@@ -150,8 +156,8 @@ public class FastRawTransactionManagerIT extends Scenario {
         }
 
         for (int i = 0;
-                i < DEFAULT_POLLING_ATTEMPTS_PER_TX_HASH && !pendingTransactions.isEmpty();
-                i++) {
+             i < DEFAULT_POLLING_ATTEMPTS_PER_TX_HASH && !pendingTransactions.isEmpty();
+             i++) {
             for (TransactionReceipt transactionReceipt : transactionReceipts) {
                 assertFalse(transactionReceipt.getBlockHash().isEmpty());
                 pendingTransactions.remove(transactionReceipt.getTransactionHash());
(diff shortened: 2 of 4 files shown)


gtebrean:
@junsung-cho Looks good to me, please also update the changelog.md and I will merge it

junsung-cho:
@gtebrean I updated CHANGELOG.md, thanks

junsung-cho:
@gtebrean Integration test failed, but I don't know why.
```
BesuPrivacyQuickstartIntegrationTest > simplePrivateTransactions() FAILED
    org.opentest4j.AssertionFailedError at BesuPrivacyQuickstartIntegrationTest.java:193
```

NickSneo:
> @gtebrean Integration test failed, but I don't know why.
> 
> ```
> BesuPrivacyQuickstartIntegrationTest > simplePrivateTransactions() FAILED
>     org.opentest4j.AssertionFailedError at BesuPrivacyQuickstartIntegrationTest.java:193
> ```

It is a bit flaky test, I will rerun it

core/src/main/java/org/web3j/tx/FastRawTransactionManager.java
NickSneo, line 1: I have a question with this approach, the FastRawTransactionManager has a local copy of nonce which doesn't call  and check the current value of nonce for the Account.
Now if there is any mismatch in nonce the reset function in FastRawTransactionManager with Current implementation resets the nonce to the correct nonce of the account.

But with your new approach it will reset to -1 which is not desirable in real world scenario. I would suggest rather than changing the current implementation of reset function you add another method called clear() which will set nonce to -1.

NickSneo (changes requested):

core/src/main/java/org/web3j/tx/FastRawTransactionManager.java
junsung-cho: Here's the proposed code change. Could you please review and confirm if this approach is acceptable?

Original code:
```java
    public synchronized void resetNonce() throws IOException {
        nonce = super.getNonce();
    }
```
Proposed changes:

```java
    public synchronized void resetNonce() throws IOException {
        nonce = super.getNonce().subtract(BigInteger.ONE);
    }

    public synchronized void clearNonce() throws IOException {
        nonce = BigInteger.valueOf(-1);
    }
```
In this approach, the `resetNonce` method is modified to decrement the nonce by one, as suggested in the feedback. Additionally, a new `clearNonce` method is added to set the nonce to -1. This way, we address the concern about the reset behavior and provide an alternative method to clear the nonce.

Please let me know if this approach works for you or if there are any other improvements needed.

NickSneo: Why do we want to `nonce = super.getNonce().subtract(BigInteger.ONE);` ?
I think the resetNonce method should be left unchanged, as it is working as expected. Let me know you thoughts if you think otherwise.

Your clearNonce method looks good, please go ahead and commit this change

NickSneo: Also please add a test case for clearNonce() method

junsung-cho: Also I added a test case for `clearNonce`.

NickSneo: But isn't RawTransactionManager's getNonce call - 
`web3j.ethGetTransactionCount(this.getFromAddress(), DefaultBlockParameterName.PENDING)`

So if we resetNonce() it will give the current nonce of account, and getNonce() in FastRawTransactionManager will give the next Nonce as it should be, since FastRawTransactionManager manages nonces itself thats why it should be currentNonce+1

Let me know your thoughts

junsung-cho: Here's example.

**Original code:**

1. Initial state: `nonce = -1` `txCount = 0`
2. Call `transfer()`
   1. Call `getNonce() = 0`: `nonce = 0` `txCount = 0`
   2. Call `send(nonce = 0)`: `nonce = 0` `txCount = 1`
3. Call `transfer()`
   1. Call `getNonce() = 1`: `nonce = 1` `txCount = 1`
   2. Call `send(nonce = 1)`: `nonce = 1` `txCount = 2`
4. Call `resetNonce()`: `nonce = 2` `txCount = 2`
5. Call `transfer()`
   1. Call `getNonce() = 3`: `nonce = 3` `txCount = 2`
   2. Call `send(nonce = 3)`: `nonce = 3` `txCount = 3`

We can see tx with `nonce = 2` never send.

**Proposed changes:**
1. Initial state: `nonce = -1` `txCount = 0`
2. Call `transfer()`
   1. Call `getNonce() = 0`: `nonce = 0` `txCount = 0`
   2. Call `send(nonce = 0)`: `nonce = 0` `txCount = 1`
3. Call `transfer()`
   1. Call `getNonce() = 1`: `nonce = 1` `txCount = 1`
   2. Call `send(nonce = 1)`: `nonce = 1` `txCount = 2`
4. Call `resetNonce()`: `nonce = 1` `txCount = 2`
5. Call `transfer()`
   1. Call `getNonce() = 2`: `nonce = 2` `txCount = 2`
   2. Call `send(nonce = 2)`: `nonce = 2` `txCount = 3`

I also revert changes to original code:
```java
    public synchronized void resetNonce() throws IOException {
        nonce = super.getNonce();
    }
```
and add test case for `resetNonce()` in `FastRawTransactionManagerIT.java`
```java
// ...
        for (int i = 0; i < COUNT; i++) {

            Future<TransactionReceipt> transactionReceiptFuture =
                    createTransaction(transfer, gasPrice).sendAsync();
            transactionReceipts.add(transactionReceiptFuture);
        }

        transactionManager.resetNonce();

        for (int i = 0; i < COUNT; i++) {

            Future<TransactionReceipt> transactionReceiptFuture =
                    createTransaction(transfer, gasPrice).sendAsync();
            transactionReceipts.add(transactionReceiptFuture);
        }
// ...
```

as results, we can see the test is [failed](https://github.com/hyperledger/web3j/actions/runs/10293503638/job/28489919798?pr=2084).

junsung-cho: I also add simple test case for `resetNonce()`

```java
    @Test
    public void testTransactionResetNonce() throws Exception {
        FastRawTransactionManager transactionManager =
                new FastRawTransactionManager(
                        web3j,
                        ALICE,
                        new PollingTransactionReceiptProcessor(
                                web3j, POLLING_FREQUENCY, DEFAULT_POLLING_ATTEMPTS_PER_TX_HASH));

        Transfer transfer = new Transfer(web3j, transactionManager);
        BigInteger gasPrice = transfer.requestCurrentGasPrice();

        createTransaction(transfer, gasPrice).send();
        transactionManager.resetNonce();
        createTransaction(transfer, gasPrice).send();
    }
```

NickSneo: Hey @junsung-cho , LGTM but I am testing some functionality in my local before I approve.
Thanks

NickSneo: Also can you add the test case for the same scenario as described in example.

junsung-cho: Test updated!

integration-tests/src/test/java/org/web3j/protocol/scenarios/FastRawTransactionManagerIT.java
```diff
@@ -96,6 +97,24 @@ public void testTransactionPolling() throws Exception {
...
+                        ALICE,
+                        new PollingTransactionReceiptProcessor(
+                                web3j, POLLING_FREQUENCY, DEFAULT_POLLING_ATTEMPTS_PER_TX_HASH));
+
+        Transfer transfer = new Transfer(web3j, transactionManager);
+        BigInteger gasPrice = transfer.requestCurrentGasPrice();
+
+        createTransaction(transfer, gasPrice).send();
+        createTransaction(transfer, gasPrice).send();
+        transactionManager.resetNonce();
```
NickSneo, line 114: Add check after reset what is the value of nonce

junsung-cho: Done!

PR #2028, opened 2024-04-11
change default branch to main

NickSneo:
### What does this PR do?
*required*

### Where should the reviewer start?
*required*

### Why is it needed?
*required*

## Checklist

- [ ] I've read the contribution guidelines.
- [ ] I've added tests (if applicable).
- [ ] I've added a changelog entry if necessary.

PR #2086, opened 2024-07-30
Fixing URLs and branches

Julius278:
### What does this PR do?
fix for new URL (web3j to hyperledger) including one branch fix (master to main)

doesn't matter, changes are indenpendent for each class / file

### Why is it needed?
branch name and repository owner changed

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Web3j is a lightweight, highly modular, reactive, type safe Java and
 Android library for working with Smart Contracts and integrating with
 clients (nodes) on the Ethereum network:
 
-![image](https://github.com/web3j/web3j-docs/blob/master/docs/img/web3j_network.png)
+![image](https://github.com/hyperledger/web3j-docs/blob/main/docs/img/web3j_network.png)
 
 This allows you to work with the [Ethereum](https://www.ethereum.org/)
 blockchain, without the additional overhead of having to write your own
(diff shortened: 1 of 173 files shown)


Julius278:
comparable adjustments as @NickSneo did in #2028

Julius278:
Test pipeline: https://github.com/Julius278/web3j/actions/runs/10162422088

no new tests needed, and I would guess no changelog addition needed

NickSneo (commented):
Well, these are all auto generated code, so it is not really required to change code in every file, we can anyways generate those and replace with new one. But thanks for your efforts.

Julius278:
Hey Nick, yes most of them are auto generated and could be removed from the merge request. But some changes like the README or Generator.java are changes that will make a difference.

Also I would suggest creating a class which holds the code comment String to not have duplicated code in case they are all identical. If you want, I could check that.

NickSneo:
> Hey Nick, yes most of them are auto generated and could be removed from the merge request. But some changes like the README or Generator.java are changes that will make a difference.
> 
> Also I would suggest creating a class which holds the code comment String to not have duplicated code in case they are all identical. If you want, I could check that.

Sure, you can look into it and create a PR we are happy to review

NickSneo (approved):
LGTM

Issue #1983, opened 2023-11-11
Make Web3j autocloseable


jarnaud:
## Feature description_

It would be nice to make the `Web3j` interface implement autocloseable, this will prevent users to forget about closing it.
Implementation effort should be minimal.

steffenboe:
this can be closed, resolved by https://github.com/hyperledger/web3j/pull/2102 ;)

PR #2102, opened 2024-09-23
Web3j extends AutoClosable Interface

steffenb91:
### What does this PR do?
Makes Web3j autoclosable to use it in `try-with-resources` blocks, as wished for in #1983 .

At the Web3j Interface, extending the AutoClosable Interface.

### Why is it needed?
So users do not forget to close the Web3j instance.

## Checklist


NickSneo (approved):
LGTM

PR #2105, opened 2024-10-01
Add HSM kms implementation

gtebrean:
### What does this PR do?
*required*

*required*

### Why is it needed?
*required*

## Checklist


NickSneo (commented):
Do we need to add some test cases?

gtebrean:
> Do we need to add some test cases?

can't be executed as a tes because will need to provide an AWS env with the KMS in order to can execute it, I tested it locally

NickSneo (approved):
LGTM

PR #2116, opened 2024-11-25
ENS metadata and reverse ENS record set

NickSneo:
### What does this PR do?
Adds retrieving ENS metadata and lets user set reverse ENS record

all

### Why is it needed?
Functionality was not available

## Checklist


core/src/main/java/org/web3j/ens/EnsResolver.java
```diff
@@ -150,6 +154,31 @@ protected OffchainResolverContract obtainOffchainResolver(String ensName) {
...
+            } catch (Exception e) {
+                throw new EnsResolutionException("Unable to determine sync status of node", e);
+            }
+
+            if (!isSynced) {
+                throw new EnsResolutionException("Node is not currently synced");
+            }
+
+            try {
+                return lookupOffchainResolver(ensName, credentials);
```
gtebrean, line 173: use only one try block as you catch a generic exception and throw the same one

core/src/main/java/org/web3j/ens/EnsMetadataResponse.java
```diff
@@ -0,0 +1,124 @@
...
+    public String description;
+    public Attribute[] attributes;
+    public String url;
+    public long last_request_date;
+    public int version;
+    public String background_image;
+    public String image;
+    public String image_url;
+
+    public boolean isIs_normalized() {
```
gtebrean, line 30: I think we should start to use Lombok in oder to avoid all these setters and getters

core/src/main/java/org/web3j/ens/EnsResolver.java
gtebrean: also the code should be a bit refactored to have only one line with "throw new"

core/src/main/java/org/web3j/ens/EnsResolver.java
```diff
@@ -376,13 +418,14 @@ private OffchainResolverContract lookupOffchainResolver(String ensName) throws E
...
     }
 
-    private String getResolverAddress(String ensName) throws Exception {
-        NetVersion netVersion = web3j.netVersion().send();
-        String registryContract = Contracts.resolveRegistryContract(netVersion.getNetVersion());
-
-        ENS ensRegistry =
-                ENS.load(registryContract, web3j, transactionManager, new DefaultGasProvider());
+    private OffchainResolverContract lookupOffchainResolver(String ensName, Credentials credentials)
+            throws Exception {
```
gtebrean, line 422: it's recommended to throw specific exception and avoid the Exception class for better track, only and advice for future

core/src/main/java/org/web3j/ens/NameWrapperApi.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package org.web3j.ens;
+
+import org.web3j.tx.ChainIdLong;
+
+public class NameWrapperApi {
```
gtebrean, line 17:
```suggestion
public class NameWrapperUrl {
```

core/src/main/java/org/web3j/ens/contracts/generated/ReverseRegistrar.java
```diff
@@ -0,0 +1,450 @@
...
+ * <p>Auto generated code.
+ * <p><strong>Do not modify!</strong>
+ * <p>Please use the <a href="https://docs.web3j.io/command_line.html">web3j command line tools</a>,
+ * or the org.web3j.codegen.SolidityFunctionWrapperGenerator in the 
+ * <a href="https://github.com/hyperledger/web3j/tree/main/codegen">codegen module</a> to update.
+ *
+ * <p>Generated with web3j version 4.12.2.
+ */
+@SuppressWarnings("rawtypes")
+public class ReverseRegistrar extends Contract {
```
gtebrean, line 41: is there any possibility to generate this wrapper when we need it, as we are doing for the integration tests in order to avoid this hardcode here

gtebrean (commented):
few suggestions

core/src/main/java/org/web3j/ens/EnsResolver.java
NickSneo: I have taken same logic as used before, just overloading the function with credentials param.
Shall I update the logic for both the functions?
I dont think its an ideal thing, current logic makes more sense to me, checkout this - https://docs.web3j.io/4.11.0/advanced/ethereum_name_service/#web3j-implementation

core/src/main/java/org/web3j/ens/EnsMetadataResponse.java
NickSneo: I agree, I can include Lombok in another PR and update all the classes

core/src/main/java/org/web3j/ens/NameWrapperApi.java
NickSneo: Will commit this change along with file name change

core/src/main/java/org/web3j/ens/EnsResolver.java
NickSneo: Not able to throw particular Exception as - 
```
public T send() throws Exception {
        return callable.call();
    }
```

If I set it to IOException, getting error - Unhandled exception: java.lang.Exception
  org.web3j.protocol.core.RemoteCall<T> 
  public T send()
throws Exception

gtebrean (approved):

PR #2124, opened 2024-12-09
Counter on duplicated event names

SergioIOB:
### What does this PR do?
Fix for when more than 1 event with the same name but different signature is found in a contract

This PR is a continuation of https://github.com/hyperledger-web3j/web3j/pull/1872 testing and fixing that the event corresponds to the function in buildEventFunctions

It is a simple PR, it only has a modified file and the corresponding test

### Why is it needed?
Same reasons as https://github.com/hyperledger-web3j/web3j/pull/1872
Relative to issue https://github.com/hyperledger-web3j/web3j/issues/1781


NickSneo (approved):
LGTM

gtebrean (approved):

PR #2133, opened 2024-12-20
fix Transaction.getChainId when v=27 must return null

jpnovais:
### What does this PR do?
fix `deriveChainId(long v)`, when `v=27` or `v=28` must return null instead of `0`, as per [EIP155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md#specification) 

crypto/src/main/java/org/web3j/crypto/TransactionUtils.java#deriveChainId

### Why is it needed?
The current seems to be buggy and comply with the EIP155 spec

## Checklist


NickSneo (approved):
LGTM, thanks for pointing out

jpnovais:
@NickSneo thanks for merging it. When do you think it will be released?

NickSneo:
Hey @jpnovais , we will make a release by this month end. Currently we are fixing web3j-android release.

PR #2142, opened 2025-01-16
Dynamic gas provider

NickSneo:
### What does this PR do?
Adds support for Dynamic Gas Providers

review all

### Why is it needed?
Currently web3j only supports static and defaut gas Provider, which is not ideal in real world scenarios

## Checklist


NickSneo:
Improves and provide features like priority over closed PR - https://github.com/hyperledger-web3j/web3j/pull/1856

core/src/main/java/org/web3j/tx/Contract.java
```diff
@@ -393,27 +395,63 @@ TransactionReceipt executeTransaction(
...
+                } else {
+                    tx =
+                            Transaction.createFunctionCallTransaction(
+                                    this.transactionManager.getFromAddress(),
+                                    BigInteger.ONE,
+                                    gasProvider.getGasPrice(),
+                                    gasProvider.getGasLimit(),
+                                    contractAddress,
+                                    data);
+                }
```
gtebrean, line 447: this is duplicated code, should be exported to a method for a better maintaining

core/src/main/java/org/web3j/tx/Contract.java
```diff
@@ -393,27 +395,63 @@ TransactionReceipt executeTransaction(
                 ContractEIP1559GasProvider eip1559GasProvider =
                         (ContractEIP1559GasProvider) gasProvider;
                 if (eip1559GasProvider.isEIP1559Enabled()) {
+                    Transaction tx;
```
gtebrean, line 398: assign this only once at creation

core/src/main/java/org/web3j/tx/gas/ContractGasProvider.java
```diff
@@ -14,13 +14,13 @@
...
 public interface ContractGasProvider {
-    BigInteger getGasPrice(String contractFunc);
 
-    @Deprecated
     BigInteger getGasPrice();
 
-    BigInteger getGasLimit(String contractFunc);
+    BigInteger getGasLimit(Transaction transaction);
 
     @Deprecated
```
gtebrean, line 25: this should be removed as we don't recommend other method to be used

core/src/main/java/org/web3j/tx/gas/DynamicEIP1559GasProvider.java
```diff
@@ -0,0 +1,126 @@
...
+            Web3j web3j, long chainId, Priority priority, BigDecimal customMultiplier) {
+        this.web3j = web3j;
+        this.chainId = chainId;
+        this.priority = priority;
+        this.customMultiplier = customMultiplier;
+    }
+
+    @Override
+    public boolean isEIP1559Enabled() {
+        return true;
```
gtebrean, line 51: as this comes from ContractEIP1559GasProvide when using this gas provider this check is redundant because it returns true by default, it doesn't check if eip1559 is enabled on the network. This method should be removed.

core/src/main/java/org/web3j/tx/Contract.java
NickSneo: Done

core/src/main/java/org/web3j/tx/Contract.java
NickSneo: Done

core/src/main/java/org/web3j/tx/gas/ContractGasProvider.java
NickSneo: This is being used to return a default gas limit value in case user uses DynamicGasProvider or EIP1559DynamicGasProvider without passing any transaction

core/src/main/java/org/web3j/tx/gas/DynamicEIP1559GasProvider.java
NickSneo: ContractEIP1559GasProvider is an interface extending this method isEIP1559Enabled()
This is being implemented by StaticEIP1559GasProvider too and hence in EIP1559DynamicGasProvider also i have implemented this to return true.

Do you mean shall I remove this method from both the classes and in interface default it to true?

core/src/main/java/org/web3j/tx/gas/ContractGasProvider.java
gtebrean: than let's remove the @deprecated annotation because is misleading

gtebrean:
only 2 comments, and resolve conflicts, rest is g2g

core/src/main/java/org/web3j/tx/gas/DynamicEIP1559GasProvider.java
NickSneo: Makes sense, removed

PR #2143, opened 2025-01-21
Fix Java version parsing for pre- and post-Java 9 versions

0xwitty:
### What does this PR do?

I’ve updated the `getJavaVersionAsDouble()` method to correctly handle Java versions both before and after Java 9. The code is now more robust and adaptable to changes in the Java version format.

This should resolve any issues related to version parsing.

## Checklist

### DIFF
--- a/utils/src/main/java/org/web3j/commons/JavaVersion.java
+++ b/utils/src/main/java/org/web3j/commons/JavaVersion.java
@@ -12,13 +12,38 @@
  */
 package org.web3j.commons;
 
+/**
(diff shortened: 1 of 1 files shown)


NickSneo (changes requested):
Can you add test also for this?

gtebrean:
We have double versions only for java 8, which has no longer support. As this library will migrate soon to java 21 it will become incompatible with java 8 environments, making this new method worthless, so I'm closing this PR.

PR #2162, opened 2025-03-08
7702 support : Transaction Response Add authorization

cherry-yl-sh:
### What does this PR do?
The display of transaction and block interfaces can show information about 7702.
start with  authorizationList
### Why is it needed?
need show the authoriaction list
## Checklist


gtebrean:
Modifying the functionality of the current Transaction class won't be enough for this and also will break the existing compatibility. In this video the EIP-7702 it's explained really well https://www.youtube.com/watch?v=4G2KwbxjKi8. At a first view I think will need a new Transaction class which extends the current one to do the magic...I  also need to do a research on this.

gtebrean:
closing this due to no update

PR #2164, opened 2025-03-12
fix: ENS CCIP Read improvements

Syn-McJ:
### What does this PR do?
The current implementation of [ERC-3668: CCIP Read](https://eips.ethereum.org/EIPS/eip-3668) for ENS name resolution is too strict and doesn't follow the standard exactly.
- Currently, the `{sender}` parameter is enforced in URL while the standard specifies that it can be sent in POST as part of DTO.
- Currently, the callback request is always made to the hard-coded `resolveWithProof` method while the standard specifies that the `callbackFunction` from OffchainLookup should be used.
- Data in the `isEIP3668` method might have double quotes that must be stripped.

The ERC-3668: https://eips.ethereum.org/EIPS/eip-3668#abstract

### Why is it needed?
These problems prevent L2 subdomains created with [Durin](https://durin.dev/) from being resolved, as well as others like `base.eth` subdomains.

## Checklist

### DIFF
--- a/core/src/main/java/org/web3j/dto/EnsGatewayRequestDTO.java
+++ b/core/src/main/java/org/web3j/dto/EnsGatewayRequestDTO.java
@@ -16,11 +16,17 @@ package org.web3j.dto;
 public class EnsGatewayRequestDTO {
 
     private String data;
+    private String sender;
 
     public EnsGatewayRequestDTO() {}
 
     public EnsGatewayRequestDTO(String data) {
         this.data = data;
+    }
+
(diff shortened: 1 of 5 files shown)


NickSneo:
Hey @Syn-McJ , thanks for the useful PR, can you plz sign the DCO to all the commits.

Syn-McJ:
> Hey @Syn-McJ , thanks for the useful PR, can you plz sign the DCO to all the commits.

Done

NickSneo (approved):
LGTM

Issue #2151, opened 2025-02-14
Auto generated toString method for structs


Julius278:
## Feature description_
Hi all, I guess this could be a smaller feature.
Working with structs can be very helpful to retrieve objects.
But an auto generated toString method with all variables would be pretty nice for fast prototyping and logging.

I didnt really find a good spot to put it, otherwise I would have taken a deeper look into it.

aceppaluni:
Hi @Julius278 
Does this conform to what you had anticipated? 
#2169

Julius278:
Hi @aceppaluni,
looks good :) thanks for your effort

PR #2169, opened 2025-03-19
Adding auto generated toString method for structs

gtebrean:
### What does this PR do?
*required*

*required*

### Why is it needed?
*required*

## Checklist


aceppaluni:
What does this PR do?
- This PR aims to address issue #2151, adding in auto generated toString() method for structs. 
- The review should begin with testing the generateToStringMethod() for any errors. Currently there is an assertion error within the testing suit. 
Why is it needed?
- The method is needed so as to retrieve objects and assist with prototyping and logging.

codegen/src/test/java/org/web3j/codegen/GenerateToStringMethodTest.java
```diff
@@ -0,0 +1,73 @@
...
+            format.setLength(format.length() - 2);
+        };
+        format.append("}\"");
+        //assertTrue(actualMethodBody.contains("return \"{ name='\" + name + \"', age='\" + age + \"' }\";"));
+
+        // Step 4: Ensure the generated method matches the expected format
+        //String actualMethodBody = toStringMethod.toString();
+        //assertTrue(actualMethodBody.contains("return \"{ name='\" + name + \"', age='\" + age + \"' }\";"));
+        //assertTrue(actualMethodBody.contains(expectedFormat), "The toString method was not generated correctly.");
+        assertTrue(actualMethodBody.contains("return \"{ name='\" + name + \"', age='\" + age + \"' }\";"),
```
gtebrean, line 47:
```suggestion
        assertTrue(actualMethodBody.contains("return \"{ name='\" + name + \"', age='\" + age + \"',  }\";"),
```
This will fix the current test.

However I have some points about the current approach.

1. If this is just a dreft is fine, else copy /pasting the code in this class in a private method to test it is not the right thing.
2. For the moment I will stop do any changes on Solidity wrapper generator and have a look at what we have right now in this file you can find how to generate smart contracts with stucts https://blog.web3labs.com/hyperledger-web3j-truly-decode-support-for-dynamic-solidity-structs/.
3. There are 2 ways to do it, first I would recommend to check what to string returns on a TypeReference struct and if we can update only that class.
4. 2nd check the final wrapper where exactly should be the to string written when there is a struct (in both cases TypeReference and Struct class type),
By covering this we can make a plan and start implementing.

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -519,6 +519,7 @@ private List<TypeSpec> buildStructTypes(final List<AbiDefinition> functionDefini
             final TypeSpec.Builder builder =
                     TypeSpec.classBuilder(structName)
                             .addModifiers(Modifier.PUBLIC, Modifier.STATIC);
+                    builder.addMethod(generateToStringMethod(namedType.getComponents())); // ADDED THIS LINE
```
Julius278, line 522: Hey @aceppaluni,
I guess you wont need a comment here

codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
```diff
@@ -602,6 +603,7 @@ private List<TypeSpec> buildStructTypes(final List<AbiDefinition> functionDefini
 
             builder.superclass(namedType.isDynamic() ? DynamicStruct.class : StaticStruct.class);
             builder.addMethod(constructorBuilder.build());
+            builder.addMethod(generateToStringMethod(namedType.getComponents())); // Added this line
```
Julius278, line 606: same here, no comment needed

aceppaluni:
@gtebrean I noticed the branch is behind. Should I try to sync the branch to test the build?

NickSneo:
@aceppaluni @gtebrean Hey, what is the status of this PR? Would love to merge this in the upcoming release

PR #2174, opened 2025-03-26
fix: dynamic array encoding

xj894976677:
# Dynamic Array Encoding Fix in Web3j

## Problem Statement
The original implementation did not correctly handle dynamic arrays according to Solidity's ABI encoding specification. In Solidity, when using `abi.encode()` with array parameters, the output includes dynamic offsets pointing to the actual array data.

### Example in Solidity
```solidity
// Solidity contract
function encodeExample(uint256[] memory tokenIds, uint256[] memory amounts) public pure returns (bytes memory) {
    return abi.encode(tokenIds, amounts);
}
```

When calling this function with:
```javascript
tokenIds = [1, 2]
amounts = [100, 200, 300]
```

The encoded output structure should be:
```
[offset_1][offset_2][array1_length][array1_data...][array2_length][array2_data...]
```

## Issue
The original Web3j implementation was concatenating array data directly without proper offset handling:
```java
// Original problematic implementation
ByteArrayOutputStream concatenatedArrayEncodingBuffer = new ByteArrayOutputStream();
for (Object arrayItem : arrayItems) {
    // Directly writing array data without offsets
    concatenatedArrayEncodingBuffer.write(arrayItemEncoding, 0, arrayItemEncoding.length);
}
```

## Solution
Modified the implementation to follow Solidity's ABI encoding specification:
```java
// Fixed implementation
// 1. Calculate and write offsets in the header
encValues.add(BigInteger.valueOf(headSize + dynamicDataSize));

// 2. Store dynamic data separately
ByteArrayOutputStream dynamicBuffer = new ByteArrayOutputStream();
// Write array length
byte[] lengthBytes = Numeric.toBytesPadded(BigInteger.valueOf(arrayItems.size()), 32);
dynamicBuffer.write(lengthBytes, 0, lengthBytes.length);
// Write array elements
for (Object arrayItem : arrayItems) {
    byte[] itemBytes = Numeric.toBytesPadded(convertToBigInt(arrayItem), 32);
    dynamicBuffer.write(itemBytes, 0, itemBytes.length);
}
```

## Impact
- Ensures compatibility with Solidity's `abi.encode()` function
- Fixes signature verification issues in contracts using EIP-712
- Correctly handles multiple dynamic arrays in structured data

This change aligns Web3j's implementation with Ethereum's ABI specification for dynamic array encoding.

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
+++ b/crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
@@ -15,7 +15,6 @@ package org.web3j.crypto;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
 import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -307,10 +306,24 @@ public class StructuredDataEncoder {
 
         List<String> encTypes = new ArrayList<>();
         List<Object> encValues = new ArrayList<>();
+        List<byte[]> dynamicData = new ArrayList<>(); // Store dynamic data
 
         // Add typehash
         encTypes.add("bytes32");
         encValues.add(typeHash(primaryType));
+
+        // Calculate header size (static part)
+        int headSize = 32; // typehash
+        for (StructuredData.Entry field : types.get(primaryType)) {
+            if (field.getType().contains("[]")) {
+                headSize += 32; // Dynamic array offset takes 32 bytes
+            } else {
+                headSize += 32; // Static type takes 32 bytes
+            }
+        }
+
+        // Track total size of dynamic data
+        int dynamicDataSize = 0;
 
         // Add field contents
         for (StructuredData.Entry field : types.get(primaryType)) {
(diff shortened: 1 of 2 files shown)


crypto/src/test/java/org/web3j/crypto/StructuredDataEncoderTest.java
```diff
@@ -0,0 +1,100 @@
...
+        StructuredDataEncoder encoder = new StructuredDataEncoder(jsonMessage);
+        encoder.hashStructuredData();
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+
+        byte[] dataHash =
+                encoder.encodeData(
+                        encoder.jsonMessageObject.getPrimaryType(),
+                        (HashMap<String, Object>) encoder.jsonMessageObject.getMessage());
+        System.out.println("################ hash is " + Numeric.toHexString(dataHash));
+        baos.write(dataHash, 0, dataHash.length);
```
gtebrean, line 98: This has nothing to do with a unit test. Please create a proper test smart contract abi and assert the expected output of the encode result

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -307,11 +306,25 @@ public byte[] encodeData(String primaryType, HashMap<String, Object> data)
...
         encValues.add(typeHash(primaryType));
 
+        // Calculate header size (static part)
+        int headSize = 32; // typehash
+        for (StructuredData.Entry field : types.get(primaryType)) {
+            if (field.getType().contains("[]")) {
+                headSize += 32; // Dynamic array offset takes 32 bytes
+            } else {
+                headSize += 32; // Static type takes 32 bytes
+            }
```
gtebrean, line 322: the check here makes no sense as the same line of code is executed every time

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -337,86 +350,117 @@ public byte[] encodeData(String primaryType, HashMap<String, Object> data)
...
+                        byte[] temp =
+                                Numeric.hexStringToByteArray(
+                                        TypeEncoder.encode(
+                                                typeClazz
+                                                        .getDeclaredConstructor(parameterTypes)
+                                                        .newInstance(value)));
+                        baos.write(temp, 0, temp.length);
+                        encoded = true;
+                        break;
+                    } catch (Exception ignored) {
```
gtebrean, line 449: this should be handled

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -337,86 +350,117 @@ public byte[] encodeData(String primaryType, HashMap<String, Object> data)
...
+                    }
                 }
-            }
 
-            if (!atleastOneConstructorExistsForGivenParametersType) {
```
gtebrean, line 411: please keep the same old check, it gives more context than  "decoded"

gtebrean (changes requested):
I see what you mean and I apreciate the work. Please sign you commits, you can find ho in the details of the failing check. Also please check the existing unit tests for 712 and create some meaningful ones for this.

xj894976677:
I observed that there is no sign in my previous commits, but it is a bit troublesome for me to modify it. I can close the merge and resubmit a code

PR #2184, opened 2025-04-25
feat: dynamic array encode

xj894976677:
# Dynamic Array Encoding Fix in Web3j

## Problem Statement
The original implementation did not correctly handle dynamic arrays according to Solidity's ABI encoding specification. In Solidity, when using `abi.encode()` with array parameters, the output includes dynamic offsets pointing to the actual array data.

### Example in Solidity
```solidity
// Solidity contract
function encodeExample(uint256[] memory tokenIds, uint256[] memory amounts) public pure returns (bytes memory) {
    return abi.encode(tokenIds, amounts);
}
```

When calling this function with:
```javascript
tokenIds = [1, 2]
amounts = [100, 200, 300]
```

The encoded output structure should be:
```
[offset_1][offset_2][array1_length][array1_data...][array2_length][array2_data...]
```

## Issue
The original Web3j implementation was concatenating array data directly without proper offset handling:
```java
// Original problematic implementation
ByteArrayOutputStream concatenatedArrayEncodingBuffer = new ByteArrayOutputStream();
for (Object arrayItem : arrayItems) {
    // Directly writing array data without offsets
    concatenatedArrayEncodingBuffer.write(arrayItemEncoding, 0, arrayItemEncoding.length);
}
```

## Solution
Modified the implementation to follow Solidity's ABI encoding specification:
```java
// Fixed implementation
// 1. Calculate and write offsets in the header
encValues.add(BigInteger.valueOf(headSize + dynamicDataSize));

// 2. Store dynamic data separately
ByteArrayOutputStream dynamicBuffer = new ByteArrayOutputStream();
// Write array length
byte[] lengthBytes = Numeric.toBytesPadded(BigInteger.valueOf(arrayItems.size()), 32);
dynamicBuffer.write(lengthBytes, 0, lengthBytes.length);
// Write array elements
for (Object arrayItem : arrayItems) {
    byte[] itemBytes = Numeric.toBytesPadded(convertToBigInt(arrayItem), 32);
    dynamicBuffer.write(itemBytes, 0, itemBytes.length);
}
```

## Impact
- Ensures compatibility with Solidity's `abi.encode()` function
- Fixes signature verification issues in contracts using EIP-712
- Correctly handles multiple dynamic arrays in structured data

This change aligns Web3j's implementation with Ethereum's ABI specification for dynamic array encoding.

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
+++ b/crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
@@ -15,7 +15,6 @@ package org.web3j.crypto;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
 import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -102,7 +101,7 @@ public class StructuredDataEncoder {
                 String baseDeclarationTypeName =
                         arrayTypePattern.matcher(declarationFieldTypeName).find()
                                 ? declarationFieldTypeName.substring(
-                                        0, declarationFieldTypeName.indexOf('['))
+                                0, declarationFieldTypeName.indexOf('['))
                                 : declarationFieldTypeName;
                 if (!types.containsKey(baseDeclarationTypeName)) {
                     // Don't expand on non-user defined types
@@ -307,16 +306,21 @@ public class StructuredDataEncoder {
 
         List<String> encTypes = new ArrayList<>();
         List<Object> encValues = new ArrayList<>();
+        List<byte[]> dynamicData = new ArrayList<>(); // Store dynamic data
 
         // Add typehash
         encTypes.add("bytes32");
         encValues.add(typeHash(primaryType));
 
+
+
         // Add field contents
         for (StructuredData.Entry field : types.get(primaryType)) {
             Object value = data.get(field.getName());
 
-            if (value == null) continue;
+            if (value == null) {
+                continue;
+            }
 
             if (field.getType().equals("string")) {
                 encTypes.add("bytes32");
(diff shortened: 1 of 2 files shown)


crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -335,88 +339,123 @@ public byte[] encodeData(String primaryType, HashMap<String, Object> data)
                 encTypes.add(field.getType());
                 encValues.add(Numeric.hexStringToByteArray((String) value));
             } else if (arrayTypePattern.matcher(field.getType()).find()) {
+                // Calculate header size (static part)
+                int headSize = 32;
```
subhramit, line 343: You are not calculating the header size, but hardcoding it - so the comment is misleading.
It's better to convert this into a constant like `static final HEADER_SIZE` just after the class declaration above the constructor at and use it wherever needed.
If you name the constant descriptively enough, the comment can also be let go of.

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -335,88 +339,123 @@ public byte[] encodeData(String primaryType, HashMap<String, Object> data)
...
+                // Track total size of dynamic data
+                int dynamicDataSize = 0;
```
subhramit, line 345: Redundant comment

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -335,88 +339,123 @@ public byte[] encodeData(String primaryType, HashMap<String, Object> data)
...
-                                convertToEncodedItem(
-                                        baseTypeName,
-                                        arrayItem); // add raw item, packed to 32 bytes
+
+                if (baseTypeName.startsWith("uint")
+                        || baseTypeName.startsWith("int")
+                        || baseTypeName.equals("address")
+                        || baseTypeName.equals("bool")) {
+                    // Handle dynamic array
+                    encTypes.add(baseTypeName); // Use base type instead of array type
```
subhramit, line 354: This seems to be AI-generated.
Nevertheless, if this is correct (I don't have much context), the comment here is useless as it talks about a change, which makes sense in the context of this PR, but won't once merged. Should be removed.
Think - how would the code look like once merged?

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -335,88 +339,123 @@ public byte[] encodeData(String primaryType, HashMap<String, Object> data)
...
+                    // Add offset position, considering actual size of all previous dynamic data
+                    encValues.add(BigInteger.valueOf(headSize + dynamicDataSize));
+
+                    // Prepare dynamic data
+                    ByteArrayOutputStream dynamicBuffer = new ByteArrayOutputStream();
```
subhramit, line 359: Redundant comment.

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -335,88 +339,123 @@ public byte[] encodeData(String primaryType, HashMap<String, Object> data)
...
+                    // Write array length
+                    byte[] lengthBytes =
+                            Numeric.toBytesPadded(BigInteger.valueOf(arrayItems.size()), 32);
```
subhramit, line 362: Extract magic number to a constant

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -335,88 +339,123 @@ public byte[] encodeData(String primaryType, HashMap<String, Object> data)
...
+                    dynamicBuffer.write(lengthBytes, 0, lengthBytes.length);
+
+                    // Write array elements
+                    for (Object arrayItem : arrayItems) {
+                        BigInteger itemValue = convertToBigInt(arrayItem);
+                        byte[] itemBytes = Numeric.toBytesPadded(itemValue, 32);
```
subhramit, line 368: Same here ^

crypto/src/main/java/org/web3j/crypto/StructuredDataEncoder.java
```diff
@@ -335,88 +339,123 @@ public byte[] encodeData(String primaryType, HashMap<String, Object> data)
...
+                        dynamicBuffer.write(itemBytes, 0, itemBytes.length);
                     }
 
-                    concatenatedArrayEncodingBuffer.write(
-                            arrayItemEncoding, 0, arrayItemEncoding.length);
+                    byte[] dynamicBytes = dynamicBuffer.toByteArray();
+                    dynamicData.add(dynamicBytes);
+                    // Update total size of dynamic data
+                    dynamicDataSize += dynamicBytes.length;
```
subhramit, line 375: Redundant comment - possibly because of AI

subhramit (changes requested):
Hi, I am not a maintainer (in fact, I just started exploring this project) and was having a hard time finding good first issues to solve myself, so thought of beginning to contribute by reviewing code.

I don't know if the PR is correct or not, as I don't have a lot of context - so pretty sure there will be many more changes when the maintainers take a look, but I left some comments below on code quality which would help you regardless of whether the PR goes in.

gtebrean:
@xj894976677 is this still relevant? asking because pipeline keeps failing

gtebrean:
closing this due to no update

PR #2186, opened 2025-05-20
fix for multidimensional array endcoding problem

ko5tik:
### What does this PR do?

Fixes this bug:
https://github.com/LFDT-web3j/web3j/issues/1912

By looking into function signature generation.  Actual release  produces  wrong result  for multidimensional arrays. 

This function:
````
    function execute(
        uint256 foo,  // start amount or 0
        uint256 bar,  // also start amount or 0
        uint256 baz, // last block we are willing to execute
        address[] calldata glyrge, //  list of pairs
        uint256[][] calldata whatever, //  expected reserves
        bytes calldata _data // all other data for further exchanges
    ) external;
````

Produces this parameter  signature while encoding:
````
uint256,uint256,uint256,address[],dynamicarray[],bytes    
````

Which is plain wrong, and does not match deployed contract. 

### Why is it needed?
It would be nice to be able to call    contracts methods with  multidimensional array parameters:

uint256[][] 

Higher dimensions also work as demonstrated by test

## Checklist

### CODE REVIEW

File: abi/src/test/java/org/web3j/abi/datatypes/DynamicArrayTest.java [hunk 1]
```diff
@@ -12,12 +12,14 @@
  */
 package org.web3j.abi.datatypes;
 
-import java.util.Collections;
-import java.util.List;
+import java.math.BigInteger;
+import java.util.*;
```
gtebrean, line 16: please remove the wildcard import

gtebrean (commented):
update also the changelog.md please

ko5tik:
Fuck.     Rebase for signoff went  fubar.      I will  create new clean branch and  new pull request from there.

ko5tik:
Due to rebasing mishap on my side,   please use new PR instead:

https://github.com/LFDT-web3j/web3j/pull/2188

PR #2187, opened 2025-05-29
EIP-4337, Account Abstraction support

gtebrean:
### What does this PR do?
*required*

*required*

### Why is it needed?
*required*

## Checklist


gtebrean:
it's in testing for receipt support

NickSneo:
Do we need to add it in https://github.com/LFDT-web3j/web3j/blob/main/crypto/src/main/java/org/web3j/crypto/transaction/type/TransactionType.java (TransactionType ) also?

And have encoding/decoding like in https://github.com/LFDT-web3j/web3j/blob/main/crypto/src/main/java/org/web3j/crypto/TransactionDecoder.java

NickSneo:
@gtebrean LGTM, lets merge it as an experiment feature

Syn-McJ:
I think `EIP4337Transaction` should be renamed to `UserOperation` or `UserOperationStruct` as this is not an actual transaction sent to the pool, but a user operation sent to a bundler. This is a term used across other account abstraction infra: https://www.alchemy.com/docs/wallets/infra/send-user-operations

The existing `UserOperation` can be either folded into the same class or renamed to `UserOperationResult` or something.

PR #2188, opened 2025-05-30
Dynamic array bug fix  2.  attempt

ko5tik:
### What does this PR do?

Attention!   This is a second version  of https://github.com/LFDT-web3j/web3j/pull/2186  -  there was  rebasing mishap on my side due to sign off comments 

Fixes this bug:

https://github.com/LFDT-web3j/web3j/issues/1912

By looking into function signature generation. Actual release produces wrong result for multidimensional arrays.

This function:
```

    function execute(
        uint256 foo,  // start amount or 0
        uint256 bar,  // also start amount or 0
        uint256 baz, // last block we are willing to execute
        address[] calldata glyrge, //  list of pairs
        uint256[][] calldata whatever, //  expected reserves
        bytes calldata _data // all other data for further exchanges
    )  external;
```
Produces this parameter signature while encoding:

`uint256,uint256,uint256,address[],dynamicarray[],bytes    `

Which is plain wrong, and does not match deployed contract.

### Why is it needed?
It would be nice to be able to call contracts methods with multidimensional array parameters:

`uint256[][]`

Higher dimensions also work as demonstrated by test

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 ### Bug Fixes
 
-* 
+* fix multidimensional array encoding [#1912](https://github.com/LFDT-web3j/web3j/issues/1912)
+
 ### Features
 
 * bump snapshot version to 4.14.1 [#2176](https://github.com/hyperledger-web3j/web3j/pull/2176)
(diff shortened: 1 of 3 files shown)


ko5tik:
Replaces   https://github.com/LFDT-web3j/web3j/pull/2186

Issue #2180, opened 2025-04-21
Expose revert `data` field in `TransactionException` for decoding `CustomError`


dgimenez27:
# # _Issue_title_
Expose revert `data` field in `TransactionException` class for decoding `CustomError`

## _Issue_description_
Currently, when executing a transaction using `.send()` and it fails on-chain (e.g. due to a `revert`), Web3J throws a `TransactionException`. This exception provides access to the error message (`execution reverted`) and optionally the transaction hash or receipt.

However, it does **not expose the `data` field** from the JSON-RPC error response, which is **critical** for decoding custom errors (`CustomError`) returned by the contract.

## _Issue_context_
Imagine calling a contract function with `.send()` and catching a `TransactionException`. 

### 💡 Example

```java
try {
    contractInstance.withdraw(...params).send();
} catch (TransactionException ex) {
    // Can't access revert reason data here
}
```

The actual JSON-RPC response from the node may include a field like `"data": "0x..."` which contains the ABI-encoded custom error selector and parameters.

```json
{
  "jsonrpc": "2.0",
  "id": 4,
  "error": {
    "code": 3,
    "message": "execution reverted",
    "data": "0xe2517d3f000000000000000000000000..." // selector + ABI-encoded params
  }
}
```

The `data` field contains the selector and ABI-encoded parameters of the custom error, which can be decoded using the new CustomError feature added in [PR #2173](https://github.com/LFDT-web3j/web3j/pull/2173).

### 🧩 Problem

- The `data` field is currently not mapped in `org.web3j.protocol.core.Response.Error`
- `TransactionException` does not expose it either
- As a result, it's **not possible to decode `CustomError` revert reasons from `.send()`**
- Developers must work around this by using manual JSON-RPC calls via external HTTP clients

---

### ✅ Proposed enhancement

1. Extend `Response.Error` to include a `data` field (i.e., the full revert payload)
2. Add support in `TransactionException` to receive and expose that field
3. Ensure that if the JSON-RPC error includes `data`, it is accessible from the exception

```java
@JsonProperty("data")
private String data;

private final String revertData;
public String getRevertData();

```

---

### ✅ Optional

Consider adding utilities to:
- Extract the 4-byte selector from the data
- Match it against `CustomError` instances generated by Web3J wrappers

---

### 🎯 Benefits

- Enables decoding of custom error revert reasons directly from Web3J `.send()` workflows
- Eliminates the need to manually replicate `eth_call` logic using HTTP clients
[...]

PR #2192, opened 2025-06-20
Include the data field returned from custom error Issue(#2180)

bodyhedia44:
### What does this PR do?
solves the issue #2180 to expose the revert data field in TransactionException for decoding CustomError

from the Contract file, where the issue starts from the investigations 

### Why is it needed?
to get the custom error from the chain to show at web3j 

## Checklist

- [.] I've read the contribution guidelines.

### DIFF
--- a/core/src/main/java/org/web3j/protocol/exceptions/TransactionException.java
+++ b/core/src/main/java/org/web3j/protocol/exceptions/TransactionException.java
@@ -68,7 +68,12 @@ public class TransactionException extends Exception {
     public Optional<TransactionReceipt> getTransactionReceipt() {
         return transactionReceipt;
     }
-
+    
+    /**
+     * Obtain the Encoded data field from response.
+     *
+     * @return optional Encoded data field .
+     */
     public Optional<String> getEncodedDataResponse() {
         return encodedDataResponse;
     }


core/src/main/java/org/web3j/utils/RevertReasonExtractor.java
```diff
@@ -105,6 +120,24 @@ public static String retrieveRevertReason(
                 .getRevertReason();
     }
 
+    public static String retrieveRevertReasonEncodedData(
```
gtebrean, line 123: add a doc comment

gtebrean (commented):
fine, but please create also some  a uit test. As example you can take the testTransactionFailedWithRevertReason unit test

core/src/main/java/org/web3j/protocol/exceptions/TransactionException.java
```diff
@@ -60,4 +68,8 @@ public Optional<String> getTransactionHash() {
     public Optional<TransactionReceipt> getTransactionReceipt() {
         return transactionReceipt;
     }
+
+    public Optional<String> getEncodedDataResponse() {
```
gtebrean, line 72: doc coment on this too

PR #2205, opened 2025-07-24
fix: transaction serialized with access list

zouxyan:
https://github.com/LFDT-web3j/web3j/issues/2204

### What does this PR do?
To fix https://github.com/LFDT-web3j/web3j/issues/2204

https://github.com/LFDT-web3j/web3j/blob/45dcf2611ff66dd1a64cc9969815f2ed176a4add/crypto/src/main/java/org/web3j/crypto/transaction/type/Transaction4844.java#L142

### Why is it needed?
To send blob tx with access list correctly

## Checklist

### DIFF
--- a/crypto/src/main/java/org/web3j/crypto/transaction/type/Transaction4844.java
+++ b/crypto/src/main/java/org/web3j/crypto/transaction/type/Transaction4844.java
@@ -139,7 +139,7 @@ public class Transaction4844 extends Transaction1559 implements ITransaction {
         resultTx.add(RlpString.create(data));
 
         // access list
-        resultTx.add(new RlpList());
+        result.add(new RlpList(rlpAccessListRlp()));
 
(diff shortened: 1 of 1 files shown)


gtebrean:
@zouxyan please build and run locally to fix it

crypto/src/main/java/org/web3j/crypto/transaction/type/Transaction4844.java
```diff
@@ -139,7 +139,7 @@ public List<RlpType> asRlpValues(Sign.SignatureData signatureData) {
         resultTx.add(RlpString.create(data));
 
         // access list
-        resultTx.add(new RlpList());
+        result.add(new RlpList(rlpAccessListRlp()));
```
mskrzypkows, line 142: Are you sure you want to change it from `resultTx` to `result`?  I think it should stay in `resultTx`: https://eips.ethereum.org/EIPS/eip-4844

PR #2206, opened 2025-07-28
Fix missing weiValue in transaction when estimating gas for payable function

### LINKED ISSUE #1268: Java wrapper generator does not properly handle payable functions
# Java wrapper generator for smart contract does not properly handle payable functions

There is no overload generated to send Ether into the smart contract function call, even though the 
contract function is marked payable.

## Steps To Reproduce

Consider this solidity function

```
    function sendEthWithContract(address payable _recipient) external payable {
        require(msg.value > 0);
        _recipient.transfer(msg.value);
    }
```

Compile the contract and generate the Java wrapper. It generates

```
    public RemoteFunctionCall<TransactionReceipt> sendEthWithContract(String _recipient) {
        final org.web3j.abi.datatypes.Function function = new org.web3j.abi.datatypes.Function(
                FUNC_SENDETHWITHCONTRACT, 
                Arrays.<Type>asList(new org.web3j.abi.datatypes.Address(160, _recipient)), 
                Collections.<TypeReference<?>>emptyList());
        return executeRemoteCallTransaction(function);
    }
```

This is wrong as you have to generate an overload that passes Eth all the way down to the execution of the
transaction. This is clear as the function is marked as payable. 

Instead it forwards the execution of the transaction to

```
    protected TransactionReceipt executeTransaction(Function function)
            throws IOException, TransactionException {
        return executeTransaction(function, BigInteger.ZERO);
    }
```

This is clearly wrong.

## Environment
- Web3j version.  core 4.6.0, codgen 4.5.18, 
- Java 8
- Operating System Mac OS X Catalina

liulinhui:
### What does this PR do?
to fix #1268 

Fix an issue where the **weiValue** is not set in the constructed transaction during gas estimation for a payable function, which leads to failure.

org.web3j.tx.Contract

### Why is it needed?
When estimating gas for a payable function, the transaction must include the correct **weiValue** to reflect the actual call conditions. Without setting the **weiValue**, the gas estimation does not accurately represent a real invocation, and in many contracts, this leads to a revert or failure during the estimation phase.

This fix ensures the **weiValue** is properly included in the transaction used for gas estimation, which allows the estimation to succeed and reflect real execution costs.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 * bump snapshot version to 4.14.1 [#2176](https://github.com/hyperledger-web3j/web3j/pull/2176)
 * add encoding/decoding for EIP-7702 transactions [#2178](https://github.com/LFDT-web3j/web3j/pull/2178)
 * add support for Account Abstraction EIP-4337 transactions [#2187](https://github.com/LFDT-web3j/web3j/pull/2187)
+* fix missing weiValue in transaction when estimating gas for payable function
 
 ### BREAKING CHANGES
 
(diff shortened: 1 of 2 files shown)


FangshuoCao:
I’ve tested this fix on Sepolia using a simple payable contract (`deposit()` with `require(msg.value > 0)`) and confirmed that gas estimation now correctly includes the weiValue. This should be good to go.

PR #2241, opened 2025-11-02
Expose loading credentials via input stream

pawellozinski:
### What does this PR do?
Expose WalletUtils.loadCredentials via input stream. With this change it is no longer required that the wallet definitions is a file system file. This is useful for loading a wallet i.e. in a web service.

In the WalletUtils class.

### Why is it needed?
To enable loading wallet definition from sources other than a physical file (i.e. through a web interface).

## Checklist


pawellozinski:
Closing as just found WalletUtils.loadJsonCredentials method.

PR #2246, opened 2026-01-09
feat: Add EIP-7702 authorization list support to Transaction

Filter94:
### What does this PR do?
This PR adds support for 7702 transactions returned from various ETH API methods, like eth_getBlockByX, eth_getTransactionByX

[ResponseTest.java‎](https://github.com/LFDT-web3j/web3j/pull/2246/changes#diff-df1cafe6e01f6bd51dcb7477015a3ae8920c171d1b1bd8c59de1b310301fd297)
Or 
[EthBlock.java](https://github.com/LFDT-web3j/web3j/pull/2246/changes#diff-e07d404a4303a6083a5d3d7192c9c6d82be44fcd5d935aa048d9cfb7f18da0b5)

### Why is it needed?
Because otherwise it's impossible to get valid deserialized 7702 transactions from an Ethereum node with Web3j

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 ### Features
 
-* 
+* Add EIP-7702 authorization list support to Transaction [#2246](https://github.com/LFDT-web3j/web3j/pull/2246)
 
 ### BREAKING CHANGES
 


gtebrean:
@Filter94 the support for this was already added with https://github.com/LFDT-web3j/web3j/pull/2178, most of the transactions EIPs were created as separated objects. Could you please try to see if what we have now is working for you?

Filter94:
@gtebrean So it means that with the current latest version of Web3j I can fetch and deserialize a complete 7702 transaction from the chain, right? And if that's the case, `org.web3j.protocol.core.methods.response.Transaction` should have some field for the authorization list

Can you please point me where this field is in the existing codebase? You should be able to easily check this using a 7702 transaction sample and the ResponseTest

gtebrean:
@Filter94, now I get it, the 7702 transaction from now can be included in the reponse of those rpcs and that's why Transaction has to be update because else will just increase complexity of the code by managing the custom class there.

Please update the pr with the autorization tuple from crypto module (is already referenced in this module) and I will merge it, also update the changelog too. Will be included in the next release

PR #2247, opened 2026-01-20
fix(encoder): fix encoding for non-ASCII chars.

### LINKED ISSUE #1741: dynamic data encode
After dynamic array encoding, the value is not expected, there are error characters!

Kun3375:
Use getBytes(StandardCharsets.UTF_8).length instead of String.length() when calculating offsets for dynamic string arrays. This fixes incorrect ABI encoding for non-ASCII characters (Chinese, Korean, etc.) where multi-byte UTF-8 characters were incorrectly measured by character count instead of byte length.

Fixes #1741

### What does this PR do?

fix the bug when TypeEncoder encode dynamic string array with non-ASCII chars.

org.web3j.abi.TypeEncoder#encodeArrayValuesOffsets line:355

## Checklist

- [ ☑️] I've read the contribution guidelines.
- [ ☑️] I've added tests (if applicable).
- [ ❌] I've added a changelog entry if necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 ### BREAKING CHANGES
 
 * bump to snapshot version 5.0.2 [#2231](https://github.com/LFDT-web3j/web3j/pull/2231)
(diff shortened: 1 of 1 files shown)


gtebrean (commented):
Please:
1. add a entry in the changelog on bug section 
2. sign your commits, click on the failing check job and you will find how

PR #2255, opened 2026-04-03
Fix ClassLoader leak caused by static ExecutorService in Async

### LINKED ISSUE #2244: Static ExecutorService in Async.java causes ClassLoader leaks in Web Containers
# Bug: Memory Leak in Web Containers due to Static ExecutorService

The `org.web3j.utils.Async` class utilizes a `static` instance of `ExecutorService` which relies solely on a JVM Shutdown Hook for termination. This design prevents the threads from stopping when a Web Application is undeployed, causing ClassLoader leaks.

## Steps To Reproduce

1. Create a Java Web Application (WAR) that depends on the Web3j library.
2. Deploy the application to a Servlet Container (e.g., Apache Tomcat, Jetty).
3. Execute code that triggers the initialization of `org.web3j.utils.Async` (this starts the static cached thread pool).
4. **Reload** or **Undeploy** the web application via the container's management interface (without restarting the JVM).
5. Take a Heap Dump and analyze it (e.g., using Eclipse MAT).

### Expected behavior
The `ExecutorService` threads should terminate (or provide a mechanism to be terminated) when the application context is destroyed, releasing references to the `WebAppClassLoader` and allowing it to be garbage collected.

### Actual behavior
The threads created by the static `executor` remain alive because the JVM Shutdown Hook (`Runtime.getRuntime().addShutdownHook`) is **not triggered** during application redeployment (only during JVM termination).

These "Not Terminated" threads hold strong references to the `WebAppClassLoader`, preventing it from being garbage collected. Repeated redeployments lead to `java.lang.OutOfMemoryError: Metaspace` (or PermGen).

## Environment

_Describe the environment in which the issue occurs_
- Web3j version: [填入你的 Web3j 版本，例如 4.8.7]
- Java or Android version: [填入你的 Java 版本，例如 Java 8 / 11]
- Operating System: [填入你的操作系统，例如 Windows / Linux]
- Container: [填入你的容器，例如 Apache Tomcat 9.0]

## Additional context

**Root Cause Analysis:**
The issue lies in `org.web3j.utils.Async`:

```java
private static final ExecutorService executor = Executors.newCachedThreadPool();

static {
    Runtime.getRuntime().addShutdownHook(new Thread(() -> shutdown(executor)));
}

Since the executor is private and static, there is no way for a developer to manually shut it down using a ServletContextListener.

Suggested Fix: 
Please provide a public static method (e.g., Async.shutdown()) to allow manual resource cleanup, or avoid using a static thread pool that assumes it owns the JVM lifecycle.

### PR DESCRIPTION
**Pull request by Dev10-sys**
### What does this PR do?

This PR adds a manual shutdown method for the static ExecutorService used in Async.java.

Right now the executor is only shut down using a JVM shutdown hook. This works for
normal applications, but in web containers like Tomcat or Jetty the JVM does not stop
when a web application is reloaded or undeployed. Because of this, the executor threads
can stay alive and keep references to the old ClassLoader, which may cause memory leaks
over time.

This change adds a public static shutdown() method so the executor can be stopped
manually when the application is being shut down or undeployed.

The shutdown method safely:
- Calls shutdown()
- Waits for termination
- Calls shutdownNow() if needed
- Handles InterruptedException
- Can be called multiple times safely

This change does not modify existing behavior and is backward compatible.

### Why is it needed?

In web container environments, applications can be reloaded without stopping the JVM.
In that case, the static executor may keep running and hold references to the old
ClassLoader, which can lead to ClassLoader memory leaks. Providing a manual shutdown
method allows proper cleanup.

## Checklist

- [x] I've read the contribution guidelines
- [ ] I've added tests (not applicable for this small fix)
- [ ] I've added a changelog entry if required

Fixes #2244

### CODE REVIEW

**Comment by gtebrean** [comment]
over all this is: 
- Good for explicit cleanup
- Not good enough for post-shutdown safety
- Still worse than the earlier lazy-reinit version for container-managed lifecycle
compared to https://github.com/LFDT-web3j/web3j/pull/2271

Also:
- DCO is failing
- there is no changelog.md update

**Comment by Dev10-sys** [comment]
@gtebrean Thanks for feedback, that helped.
I have updated the implementation to support lazy reinitialization so the executor can safely be recreated after shutdown, which should handle container lifecycle cases better.
Also added a lifecycle test to verify shutdown and reuse, fixed the DCO issue and updated the changelog.
Everything is passing locally and CI is green now.

Let me know if there’s anything else you’d like me to adjust.

File: core/src/test/java/org/web3j/utils/AsyncLifecycleTest.java [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+package org.web3j.utils;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class AsyncLifecycleTest {
```
gtebrean: there is already AsyncTest class

core/src/main/java/org/web3j/utils/Async.java
```diff
@@ -24,10 +24,44 @@ public class Async {
...
+     * ClassLoader leaks during web app undeployment.
+     */
+    private static volatile ExecutorService executor;
+
+    private static ExecutorService getExecutor() {
+        ExecutorService result = executor;
+        if (result == null || result.isShutdown()) {
+            synchronized (Async.class) {
+                result = executor;
+                if (result == null || result.isShutdown()) {
```
gtebrean:
```suggestion
                if (result == null || result.isShutdown() || result.isTerminated()) {
```

core/src/main/java/org/web3j/utils/Async.java
gtebrean: can still throw RejectedExecutionException

PR #2264, opened 2026-04-14
fix: ensure maxPriorityFeePerGas does not exceed maxFeePerGas in DynamicEIP1559GasProvider

### LINKED ISSUE #2225: DynamicEIP1559GasProvider: max priority fee per gas higher than max fee per gas
# _Bug_title_
DynamicEIP1559GasProvider produced JsonRpcError: max priority fee per gas higher than max fee per gas

## Steps To Reproduce
it only happened once, depends on the network

### Expected behavior
gas fee calculation like it does all the time. The dynamic gas provider is pretty good, it only happened like twice 

### Actual behavior

Exception in thread "main" org.web3j.protocol.exceptions.TransactionException: JsonRpcError thrown with code -32000. Message: max priority fee per gas higher than max fee per gas

## Environment
- Web3j version, core 4.13.0
- Java 17
- Operating System: Windows 11

Dev10-sys:
This PR fixes an edge case where maxPriorityFeePerGas could exceed maxFeePerGas due to inconsistent RPC results.

The fix ensures that maxFeePerGas is always greater than or equal to maxPriorityFeePerGas by introducing a high-water mark for the priority fee during calculation.

A unit test is added to cover this scenario and verify the fix.

Fixes #2225


core/src/main/java/org/web3j/tx/gas/DynamicEIP1559GasProvider.java
```diff
@@ -54,20 +56,43 @@ public long getChainId() {
     @Override
     public BigInteger getMaxFeePerGas() {
         try {
-            BigInteger baseFee =
-                    web3j.ethGetBlockByNumber(DefaultBlockParameterName.LATEST, false)
-                            .send()
-                            .getBlock()
-                            .getBaseFeePerGas();
+            EthBlock ethBlock =
+                    web3j.ethGetBlockByNumber(DefaultBlockParameterName.LATEST, false).send();
```
NickSneo, line 59: this feels like an unnecessary change, already in try catch block, any issue should give  "Failed to get ethMaxFeePerGas" which is enough

NickSneo (commented):
I think `fetchMaxPriorityFeePerGas()` and changes in `maxPriorityFeePerGas()` are not required given the calculations are already correct.

Beside one thing I observed is Priority and custom multiplier, seems there is some dead code or priority is not being applied properly. Can you plz check that once and fix the issue. Refer to other implementation from ethersJs, viem, etc

Dev10-sys:
@NickSneo, I will look into this and make the necessary updates.

Dev10-sys:
@NickSneo please review!

PR #2265, opened 2026-04-14
fix: ensure QueuingTransactionReceiptProcessor executor is properly shut down

### LINKED ISSUE #2148: QueuingTransactionReceiptProcessor should provide a mechanism of closing the scheduledExecutorService thread pool
# Bug description
When using QueuingTransactionReceiptProcessor  in a Spring Boot, a crash (an exception thrown) during the singleton beans initialization phase leaves the Spring Boot application hanging.

## Steps to reproduce
In a SpringBoot Application using web3j, in a bean constructor call:
1. Create a QueuingTransactionReceiptProcessor:
`            transactionReceiptProcessor = new QueuingTransactionReceiptProcessor(this.readWeb3j,
                    callback,
                    txManagerPollAttempts,
                    txManagerPollSleepDuration);
`
2. Throw an exception afterwards.

### Expected behavior
The resources will be released and the Spring-boot application will exit.

### Actual behavior
The resources are not released and the Spring boot application stays hanging, waiting all the hreads to quit. The reason is that the shutdown().
The cause is that the shutdown hook is set up only after the singleton beans are created and in this case the bean creation fails.

## Environment
- Web3j version: 4.12.3
- Java : 21
- Operating System: Ubuntu 22

Dev10-sys:
This PR fixes an issue where the ScheduledExecutorService used by QueuingTransactionReceiptProcessor is not properly shut down, which can prevent the JVM from exiting in certain scenarios (e.g., Spring Boot application bootstrap failures).

A public shutdown() method is added to allow manual resource cleanup and ensure threads are properly terminated.

A unit test is included to verify that the executor is correctly shut down.

This change is backward compatible and does not affect existing behavior unless shutdown() is explicitly called.

Fixes #2148


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 ### Bug Fixes
 
+- Fix ensure QueuingTransactionReceiptProcessor executor is properly shut down to avoid JVM hang [#2265](https://github.com/LFDT-web3j/web3j/pull/2265)
```
gtebrean: remove

Dev10-sys: done!

gtebrean (approved):

PR #2266, opened 2026-04-14
feat: make Web3jService AutoCloseable and improve resource management

### LINKED ISSUE #2144: Have Web3jService implement AutoCloseable
## Feature description_
I think Web3jService should implement AutoCloseable if it sometimes needs to be closed.

Also I think Web3j should get an override for close that throws something more restricted than Exception.

Dev10-sys:
This PR makes Web3jService implement AutoCloseable, allowing it to be used with try-with-resources.

Key improvements:
- Web3jService interface now extends AutoCloseable and explicitly overrides close() to remove the generic 'throws Exception' and checked 'throws IOException'.
- Web3j interface provides a default close() method that delegates to its existing shutdown() logic.
- Service implementations (HttpService, IpcService) and callers (JsonRpc2_0Web3j) have been updated to remove unnecessary checked exception handling during resource cleanup.

This change enables idiomatic resource management in modern Java applications without breaking existing behavior or requiring verbose try-catch blocks for closing services.

Fixes #2144


NickSneo (approved):
LGTM, @gtebrean Shall we merge this `AutoCloseable` or previous approach seems better?

PR #2279, opened 2026-04-27
perf(abi): direct construction for common leaf types in TypeDecoder

snuderl:
## Summary

The most frequently decoded ABI types (`Uint256`, `Address`, `Bytes32`, etc.) currently go through reflection on every call:
`type.getConstructor(BigInteger.class).newInstance(numericValue)`. For the small set of hot-path types, this PR adds a direct-constructor fast path (`new Uint256(numericValue)`) and falls back to the existing reflective path for everything else.

`decodeAddress` is also rewritten to construct the `Uint160` directly from the input bytes (`new BigInteger(1, inputByteArray, 12, 20)`), bypassing the full `decodeNumeric` path.

## Why

Reflection is cheap on a warmed-up JIT but not free, and `Class.getConstructor` does a guarded lookup through the class's reflection cache. For high-traffic code paths (event log decoding, batched eth_call result decoding) this shows up in CPU profiles. Direct construction lets the JIT inline the constructor entirely.

## Benchmark

Single-JVM micro-benchmark, 200k warmup + 2M measure ops, best of 3 trials, run on macOS arm64, OpenJDK 25.

| Type    | main (ns/op) | patch (ns/op) | speedup |
|---------|-------------:|--------------:|--------:|
| Uint256 |        172   |        152    |  1.13x  |
| Int128  |        193   |        170    |  1.13x  |
| Address |        190   |         32    |  **5.9x** |
| Bytes32 |        132   |        113    |  1.17x  |

The dramatic Address gain comes from short-circuiting the whole `decodeNumeric` path; numeric/bytes gains are smaller because reflection is well-cached after warm-up. In cold-cache and contended-call scenarios the wins are larger.

## Compatibility

- `decodeAddress`: equivalent to before (still produces `new Address(new Uint160(value))` with the same numeric value), just constructed without intermediate hex substring + reflective lookup.
- `decodeNumeric` / `decodeBytes`: behavior is unchanged for all types — fast path runs only for an explicit identity check (`type == Uint256.class`); any other class falls through to the existing reflective constructor lookup.

## Test plan

- [x] `./gradlew :abi:test` passes
- [x] Micro-benchmark numbers above

### DIFF
--- a/abi/src/main/java/org/web3j/abi/TypeDecoder.java
+++ b/abi/src/main/java/org/web3j/abi/TypeDecoder.java
@@ -47,7 +47,7 @@ import org.web3j.abi.datatypes.Type;
 import org.web3j.abi.datatypes.Ufixed;
 import org.web3j.abi.datatypes.Uint;
 import org.web3j.abi.datatypes.Utf8String;
-import org.web3j.abi.datatypes.generated.Uint160;
+import org.web3j.abi.datatypes.generated.*;
 import org.web3j.abi.datatypes.primitive.Double;
 import org.web3j.abi.datatypes.primitive.Float;
 import org.web3j.utils.Numeric;
@@ -140,7 +140,9 @@ public class TypeDecoder {
     }
 
     public static Address decodeAddress(String input) {
-        return new Address(decodeNumeric(input, Uint160.class));
+        byte[] inputByteArray = Numeric.hexStringToByteArray(input);
+        BigInteger numericValue = new BigInteger(1, inputByteArray, 12, 20);
+        return new Address(new Uint160(numericValue));
     }
 
     public static <T extends NumericType> T decodeNumeric(String input, Class<T> type) {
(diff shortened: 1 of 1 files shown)


Dev10-sys:
Hey @snuderl , nice improvement idea here, especially avoiding reflection on hot paths.
Since this PR appears to be bot generated, I think you should review the changes carefully yourself.

I noticed one unsafe assumption in decodeAddress.

The new implementation directly slices the byte array using new BigInteger(1, inputByteArray, 12, 20). This assumes the input is always 32 bytes padded.

That is true when decoding ABI buffers internally, but TypeDecoder.decodeAddress can also be used with raw inputs. If someone passes a normal 20 byte address, this will throw IndexOutOfBoundsException.

So this introduces a potential crash for valid inputs outside the padded ABI context.

It would be safer to add a length check before using the fast path, and fall back to the existing logic for shorter inputs.

The optimization is good, just needs this guard to avoid breaking behavior.

PR #2281, opened 2026-04-27
fix: add websocket fallback for pending transactions when filter is unsupported

### LINKED ISSUE #1941: I failed to retrieve pending data using web3j with a wss node, but the same node can retrieve the data successfully using nodejs. How can I resolve this?
I failed to retrieve pending data using web3j with a wss node, but the same node can retrieve the data successfully using nodejs. How can I resolve this?

Dev10-sys:
Fixes an issue where pendingTransactionFlowable fails when the node does not support filter-based APIs like eth_newPendingTransactionFilter.

Some nodes (for example BSC or Avalanche) only support websocket subscriptions and return errors when filter APIs are used.

This change adds a fallback to use eth_subscribe (newPendingTransactions) when filter-based polling fails, ensuring pending transactions can still be received over WebSocket connections.

Also improves handling of unexpected result types returned by some nodes.

Fixes #1941


NickSneo:
Please fix conflicts

PR #2283, opened 2026-04-27
feat(ens): route resolve() through the Universal Resolver (ENSv2 readiness)

Dhaiwat10:
## Summary

Makes `EnsResolver.resolve(name)` route through the Universal Resolver, so ENS name lookups go through the ENSv2-ready path by default. **Drop-in**: consumers get UR behaviour by bumping the package version — no API change, no new method to learn.

This closes the Universal-Resolver half of the ENSv2 readiness requirement ([spec](https://docs.ens.domains/web/ensv2-readiness)). The library already had CCIP-Read plumbing; this PR wires resolution through the UR contract that orchestrates registry lookup, ENSIP-10 wildcard traversal, and CCIP-Read trampoline in a single call.

## Behaviour change

**Before**: `resolve()` hand-rolled ENS resolution — `ENSRegistry.resolver(namehash)` → `resolver.supportsInterface(ENSIP-10)` → either `resolver.addr(namehash)` or `resolver.resolve(dnsEncoded, addrCalldata)` + CCIP trampoline. 3–4 RPC calls, and missed ENSv2 scenarios the UR handles natively.

**After**: `resolve()` makes one call to the mainnet Universal Resolver at `0xeeeeeeee14d718c2b47d9923deab1335e144eeee` — the same contract and call shape viem 2.47.x uses — via `resolveWithGateways(bytes name, bytes data, string[] gateways)`. The UR handles registry, wildcard, and CCIP-Read internally.

Input/output contract is unchanged:
- Blank / `"."` → `null`
- Plain address → returned as-is
- Valid ENS name → resolved address

## Target test case

ENSv2 resolution-tests [`universal-resolver`](https://github.com/ensdomains/resolution-tests/blob/main/test-cases.json):

```json
{ "id": "universal-resolver", "input": { "name": "ur.integration-tests.eth" },
  "expected": { "address": "0x2222222222222222222222222222222222222222" },
  "method": "addr", "params": { "coinType": 60 } }
```

Verified end-to-end against live mainnet — see the test plan below.

## Files

- `core/src/main/java/org/web3j/ens/UniversalResolverContracts.java` — new. ChainId → UR address (mainnet only for now).
- `core/src/main/java/org/web3j/ens/EnsResolver.java` — `resolve()` body rewritten to call UR; adds `UNIVERSAL_RESOLVER_BATCH_GATEWAY` sentinel constant.
- `core/src/test/java/org/web3j/ens/UniversalResolverTest.java` — new unit test asserting `resolve()` goes through UR (uses mocked `(bytes, address)` return).
- `core/src/test/java/org/web3j/ens/EnsResolverTest.java` — `testResolve` updated to mock UR response format. Two wildcard-specific tests removed (ENSIP-10 wildcard is now handled inside the UR contract, not client-side — there's no branching left to test).
- `integration-tests/src/test/java/org/web3j/ens/UniversalResolverIT.java` — new live-mainnet smoke test, gated on `WEB3J_MAINNET_RPC`.

## Test plan

- [x] `./gradlew :core:test` — **396 tests, 0 failures, 0 errors**.
- [x] `./gradlew :core:test --tests "org.web3j.ens.*" --tests "org.web3j.utils.EnsUtilsTest"` — **51 tests, 0 failures** (`EnsResolverTest` dropped from 34 → 32 after removing the two wildcard tests).
- [x] Live mainnet smoke via public RPC:
  ```
  WEB3J_MAINNET_RPC=https://ethereum-rpc.publicnode.com \
    ./gradlew :integration-tests:test -Pintegration-tests=true \
    --tests org.web3j.ens.UniversalResolverIT
  ```
  Result: `ur.integration-tests.eth` → `0x2222222222222222222222222222222222222222` ✓. The UR picked resolver `0xde0c1269547dba2233f833dc1d9a7e586b4b199c` for the name (surfaced in the eth_call return; unused by `resolve()` itself but a useful signal that the UR was actually hit).

## Out of scope / follow-ups

This PR migrates only the forward `resolve()` path — enough to pass the resolution-tests `universal-resolver` case, which is the baseline ENSv2-readiness gate.

These still use the legacy `ENSRegistry` path and are good follow-up PRs:
- `reverseResolve()` — should call `UR.reverseWithGateways(bytes reverseName, uint256 coinType, string[] gateways)` to pick up ENSv2 chain-aware reverse records (L2 primary names).
- `getEnsText()` / `setEnsText()` — text records should go through UR's `resolveWithGateways` with `text(node, key)` calldata so offchain text records resolve.
- Remaining 9 resolution-tests cases: text records, contenthash, CCIP-Read through UR-wrapped OffchainLookup, L2 addr (non-60 coinType), L2 reverse, DNS names. The CCIP-Read trampoline through UR is the one non-trivial piece (UR re-wraps the resolver's `OffchainLookup` with its own callback and `extraData`); the rest are mechanical.
- Non-mainnet chains: `UniversalResolverContracts` currently only knows mainnet. Sepolia/Holesky/Base/etc. just need their UR addresses added.


Dev10-sys:
Hey @Dhaiwat10 , thanks for working on the Universal Resolver integration this is a useful direction.

As it look like this PR is LLM generated.

Also, I went through the changes carefully and there are a few concerns that should be addressed before this can be considered safe to merge.

1. Zero-address handling (critical)
   Right now the validation relies on WalletUtils.isValidAddress. The issue is that when an ENS name does not exist, the resolver can return the zero address (0x0000...0000). This still passes isValidAddress, so resolve() may silently return a zero address instead of failing. That’s risky because it can lead to users unknowingly sending funds to a burn address. The previous behavior explicitly handled this case, so we should restore an explicit zero/empty address check.

2. Resolver method compatibility
   The implementation assumes resolveWithGateways is always available. Some deployments (especially older networks or test environments) still expose only resolve(bytes,bytes). Without a fallback or capability check, this can cause failures on those networks. It would be safer to add a fallback path.

3. Test coverage regression
   A noticeable portion of EnsResolverTest has been removed and replaced with a narrower test setup. Some important edge cases (like empty resolution / invalid names) are no longer covered. We should preserve those cases and extend them instead of removing them.

 It would be great to first fix the zero-address handling and restore test coverage for invalid resolution cases.

Happy to take another look once those are updated.

Dhaiwat10:
Hey @Dev10-sys, Thanks for the review. I pushed a new commit with fixes for all three.

For the zero address: you're right, `WalletUtils.isValidAddress` accepts `0x0000...0000`. Added an explicit `EnsUtils.isAddressEmpty` check so resolve() throws EnsResolutionException instead of returning the burn address.

For the method compatibility: resolve() now tries `resolveWithGateways(bytes,bytes,string[])` first and falls back to `resolve(bytes,bytes)` if the first one reverts or returns no data. Detection uses `EthCall.isReverted()` plus an empty-result check, so it works whether the node returns a JSON-RPC error or just 0x on revert.

For test coverage, three new tests in EnsResolverTest:

`testResolveThrowsWhenZeroAddressReturned`
`testResolveFallsBackToLegacyResolveSignature`
`testResolveThrowsWhenBothSignaturesUnavailable`

One thing worth clarifying on the removed tests: only two came out (`testResolveWildCardSuccess` and `testResolveWildCardWhenResolvedAddressNotValid`). Both were testing client-side branching on supportsInterface(ENSIP_10_INTERFACE_ID), which doesn't exist in the new path because the Universal Resolver contract handles ENSIP-10 wildcard traversal internally. So the narrower test setup is really just tracking the simpler implementation. That said, you're right that the "invalid resolver result throws" case wasn't preserved, and that's what `testResolveThrowsWhenZeroAddressReturned` covers now.

Also yes, this PR is entirely generated by Claude like I disclosed on the Telegram group and in the PR description, since I don't write Java. The goal from my side here was to help you guys get started with the upgrade to ENS v2 :)) 

If you think the code quality is adequate, please feel free to merge.

core/src/main/java/org/web3j/ens/EnsResolver.java
```diff
@@ -183,57 +198,154 @@ protected OffchainResolverContract obtainOffchainResolver(
...
+     * Calls {@code UniversalResolver.resolveWithGateways(bytes, bytes, string[])} and
+     * decodes the inner {@code addr(bytes32)} return.
+     *
+     * @return resolved address, or {@code null} if the deployment does not expose this
+     *     signature (the call reverts or returns no data) — caller should fall back.
+     */
+    private String callUniversalResolverWithGateways(
+            String urAddress, String dnsEncoded, String addrCalldata, Function addrFn)
+            throws Exception {
+        Function fn =
```
NickSneo: It is better to create JAVA wrappers for UR contract and use that

core/src/main/java/org/web3j/ens/EnsResolver.java
```diff
@@ -183,57 +198,154 @@ protected OffchainResolverContract obtainOffchainResolver(
...
+            if (resolvedAddress == null) {
+                resolvedAddress =
+                        callUniversalResolverLegacy(urAddress, dnsEncoded, addrCalldata, addrFn);
+            }
 
-                String resolvedName;
-                if (supportWildcard) {
-                    String dnsEncoded = NameHash.dnsEncode(ensName);
-                    String addrFunction = resolver.addr(nameHash).encodeFunctionCall();
+            // ENS resolvers return 0x0000…0000 when a name has no addr record. Treat that
```
NickSneo: simplify comments and remove uncessary comments

NickSneo (changes requested):

Dhaiwat10:
@NickSneo thanks for the review. I've addressed your comments. Let me know if this looks ok now!

NickSneo (approved):
LGTM

gtebrean (approved):

PR #2286, opened 2026-05-02
Fix Codegen fails when trying to create wrapper from ABI compiled by …

### LINKED ISSUE #1925: Codegen fails when trying to create wrapper from ABI compiled by truffle for T-REX 4.0.0+
# _Codegen fails when trying to create wrapper from ABI compiled by truffle for T-REX 4.0.0+_

## Steps To Reproduce
Execute the codegen with ABI input with the following file: [ITREXFactory.json](https://pastebin.com/zDDYWp8z)

### Expected behavior
A wrapper being generated for the specified contract

### Actual behavior
The codegen fails with [this error message](https://pastebin.com/9ivYrTuk)

## Environment
- Web3j 4.10.0
- Java version:   openJDK17
- Operating System Win10
- Truffle version: 5.10.0
- Solidity version: 8.17.0  
  
  
## Additional context
_Add any other context about the problem here._
- Logs: error message: https://pastebin.com/9ivYrTuk
- Sample code and/or code snippets: Here are my full suite of ABIs https://www.dropbox.com/sh/08mw81jtq1ui04x/AADy6CFk3FbGrH8yr3nNn9BOa?dl=0

I am using the web3j maven plugin, but the issue is very clearly in the codegen module. It has to do with tuples with arrays or 2d arrays. I also had this same problem with JDK8, so 17 isn't the root cause either.

## Workaround  
  
If you encounter the same issue and you don't need the TREXFactory contract, then adding an exclusion to the maven plugin does allow for the rest of the build to proceed as expected.

DakshRJain737:
Fixes #1925 

### What does this PR do?
Codegen fails when trying to create wrapper from ABI compiled by truffle for T-REX 4.0.0+. The root cause is in `createMappedParameterTypes()` in `SolidityFunctionWrapper.java`, where the code blindly casts a `TypeName` to `ClassName` without checking its actual type first. When the inner type argument is a `ParameterizedTypeName` (which happens with nested arrays like `uint256[][]`), the cast fails.

Added an `instanceof ClassName` guard before the cast, and extracted the type argument into a variable to avoid redundant calls:
```java
TypeName firstTypeArg = typeNames.get(0);
if (firstTypeArg instanceof ClassName && structClassNameMap.values().stream()...)
```
If the type argument is not a `ClassName` (i.e. it's a nested array), it correctly falls through to the existing `else` block which handles `ParameterizedTypeName` properly.

The reviewer must just look into the `codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java` file.

### Why is it needed?
Because the code assumed every type inside an array is a simple type (`ClassName`), but for nested arrays like `uint256[][]`, the inner type is itself another array (`ParameterizedTypeName`).

## Checklist

### DIFF
--- a/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
+++ b/codegen/src/main/java/org/web3j/codegen/SolidityFunctionWrapper.java
@@ -1325,29 +1325,20 @@ public class SolidityFunctionWrapper extends Generator {
             if (typeNames.size() != 1) {
                 throw new UnsupportedOperationException(
                         "Only a single parameterized type is supported");
-            } else if (structClassNameMap.values().stream()
-                    .map(ClassName::simpleName)
-                    .anyMatch(
-                            name ->
-                                    name.equals(
-                                            ((ClassName)
-                                                            ((ParameterizedTypeName)
-                                                                            parameterSpec.type)
-                                                                    .typeArguments.get(0))
-                                                    .simpleName()))) {
+            }
+            TypeName firstTypeArg = typeNames.get(0);
+            if (firstTypeArg instanceof ClassName
+                    && structClassNameMap.values().stream()
+                            .map(ClassName::simpleName)
+                            .anyMatch(
+                                    name -> name.equals(((ClassName) firstTypeArg).simpleName()))) {
                 String structName =
                         structClassNameMap.values().stream()
                                 .map(ClassName::simpleName)
                                 .filter(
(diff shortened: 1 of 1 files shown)


NickSneo:
Is anyone still using Truffle?
Truffle was sunset by Consensys in September 2023 - https://consensys.io/blog/consensys-announces-the-sunset-of-truffle-and-ganache-and-new-hardhat

Issue #1904, opened 2023-05-06
When call a function with a parameter of two-dimensional array, it goes wrong


cylonggit:
# When call a function with a parameter of two-dimensional array, it goes wrong
_A clear and concise description of what the issue is._

## Steps To Reproduce
 the solidity code:
    function testParas(string[][] calldata array1 ) public view returns (string memory){
        return (array1[1][1]);
    }

the java wrapper:
    public RemoteFunctionCall<String> testParas(List<List<String>> updvo) {
        final org.web3j.abi.datatypes.Function function = new org.web3j.abi.datatypes.Function(FUNC_TESTPARAS3, 
                Arrays.<Type>asList(new org.web3j.abi.datatypes.DynamicArray<org.web3j.abi.datatypes.DynamicArray>(
                        org.web3j.abi.datatypes.DynamicArray.class,
                        org.web3j.abi.Utils.typeMap(updvo, org.web3j.abi.datatypes.DynamicArray.class,
                org.web3j.abi.datatypes.Utf8String.class))), 
                Arrays.<TypeReference<?>>asList(new TypeReference<Utf8String>() {}));
        return executeRemoteCallSingleValueReturn(function, String.class);
    }

the java code:
   List<String> u_0=new ArrayList<>();
   u_0.add("7300");
   u_0.add("7400");
   List<String> u_1=new ArrayList<>();
   u_1.add("7200");
   u_1.add("mark");
  
   List<List<String>> uAll = new ArrayList<>();
   uAll.add(u_0);
   uAll.add(u_1);
  
   String r=EthService.myContract_sp.testParas(uAll).send();

then it will go with an error:
org.web3j.tx.exceptions.ContractCallException: Contract Call has been reverted by the EVM with the reason: 'execution reverted'.

But I can call this function in Remix, and it's all ok.

## Environment
_Describe the environment in which the issue occurs_
- Web3j version : 4.9.8
- Java or Android version: 1.8
- Operating System: mac 13.1
- Solidity version: 0.8.7

EthanOK:
I have the same problem.

Can you help me get java code?
```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

contract TestArray {
    struct Order {
        address from;
        address to;
    }

    function setOrder(Order calldata) external {}

    function setOrders(Order[] calldata) external {}

    function setOrderss(Order[][] calldata) external {}

    // function get(uint256[][] calldata) external {}
}
```

[...]

PR #2287, opened 2026-05-03
fix: handle nested DynamicArray types correctly (string[][]) (#1904)

Dev10-sys:
Fixes issue #1904 ,
where nested arrays like string[][] were not getting the correct type and sometimes produced wrong values like dynamicarray[], causing contract call failures. 

Updated DynamicArray.getTypeAsString() to handle nested array types properly by checking for Array types instead of only specific cases, so multi-dimensional arrays resolve correctly. Verified with tests and full build.


abi/src/test/java/org/web3j/abi/datatypes/Issue1904ReproductionTest.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package org.web3j.abi.datatypes;
+
+import java.util.Collections;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class Issue1904ReproductionTest {
```
NickSneo: Add this test in `abi/src/test/java/org/web3j/abi/datatypes/DynamicArrayTest.java`

abi/src/test/java/org/web3j/abi/datatypes/NestedArrayTypeStringTest.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+import org.web3j.abi.datatypes.generated.StaticArray2;
+import org.web3j.abi.datatypes.generated.Uint256;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class NestedArrayTypeStringTest {
```
NickSneo: same, add this test in `abi/src/test/java/org/web3j/abi/datatypes/DynamicArrayTest.java`

abi/src/test/java/org/web3j/abi/datatypes/Issue1904ReproductionTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Test
+    @SuppressWarnings("unchecked")
```
NickSneo: Why are we adding  `@SuppressWarnings("unchecked")`?

Dev10-sys: I had added @SuppressWarnings("unchecked") initially because declaring the tests with fully parameterized nested types (like DynamicArray<DynamicArray<Utf8String>>) forced us to double-cast DynamicArray.class to satisfy the constructor which triggered compiler warnings.

As i have moved all the test cases into DynamicArrayTest.java and deleted the separate test files. Also , I refactored the declarations to use raw types for nested arrays matching the existing testMultidimensionalDynamicArray style. This allows the compiler to infer the types cleanly without casting. so the @SuppressWarnings("unchecked") annotations are no longer needed and have been removed.

abi/src/test/java/org/web3j/abi/datatypes/DynamicArrayTest.java
```diff
@@ -63,6 +65,81 @@ public void testMultidimensionalDynamicArray() {
         assertEquals("uint256[][][]", array.getTypeAsString());
     }
 
+    @Test
+    public void testString2DArrayTypeAsString() {
```
NickSneo, line 69: Add some tests for end to end encoding `TypeEncoder.encode()` and decoding

NickSneo:
@Dev10-sys Check if web3j works correctly with 3d,4d, ... etc dynamic arrays

Dev10-sys:
@NickSneo I have checked and added tests for 3D and 4D arrays as well. The fix handles multidimensional dynamic arrays correctly across all dimensions. Ready for review!

gtebrean:
@NickSneo anything else to add?

PR #2293, opened 2026-05-15
fix: prevent Transaction serialization failures when fee fields are null

### LINKED ISSUE #1830: org.web3j.protocol.core.methods.response.Transaction can't be serialized
# Transaction serialization bug
getMaxPriorityFeePerGas does not check null value. So if tx has no maxPriorityFeePerGas, exception throwed from Numeric.decodeQuantity

Dev10-sys:
This fixes a Transaction serialization issue caused by nullable numeric fields being decoded without null checks.

During Jackson serialization, getters such as `getMaxPriorityFeePerGas()` were calling `Numeric.decodeQuantity(...)` directly. If the underlying field was null, serialization failed with a decoding exception.

## Changes

Added explicit null checks to Transaction getters using `Numeric.decodeQuantity(...)`, including:

- getNonce()
- getBlockNumber()
- getTransactionIndex()
- getValue()
- getGasPrice()
- getGas()
- getMaxPriorityFeePerGas()
- getMaxFeePerBlobGas()

This keeps the behavior consistent with existing null-safe getters such as `getMaxFeePerGas()`.

## Testing

Added a regression test to verify that Transaction objects with null fee fields can be serialized successfully without throwing exceptions.

Verified locally with:

./gradlew :core:test --tests org.web3j.protocol.core.methods.response.TransactionTest
./gradlew :core:test --tests org.web3j.protocol.core.methods.response.EthBlockTest

Fixes #1830

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 ### Bug Fixes
 
--
+- Fix Transaction serialization failure when fee fields are null [#2293](https://github.com/LFDT-web3j/web3j/pull/2293)
+
 
 ### Features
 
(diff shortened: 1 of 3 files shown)


NickSneo (approved):
LGTM

PR #2304, opened 2026-07-17
fix(integration-tests): dynamically remove unstable legacy sonatype repositories

Dev10-sys:
The `integration-test` job has been failing on CI with gateway timeouts (504) while trying to download dependencies from `oss.sonatype.org`. Since Sonatype's legacy staging repository is unstable, Gradle gets blocked on the timeout and aborts before checking subsequent working repositories that actually contain the required artifacts.

This PR adds a dynamic `removeIf` filter in the subproject's repositories container to strip any legacy `oss.sonatype.org` repository. We do this dynamically because the parent build tools configuration applies these repositories to all subprojects via `allprojects`, so simply removing them from the local file alone doesn't clear them.

Tested locally by compiling the project and running the integration tests with `--refresh-dependencies` to ensure clean resolution.

### DIFF
--- a/integration-tests/build.gradle
+++ b/integration-tests/build.gradle
@@ -2,8 +2,9 @@ description 'web3j integration tests'
 
 
 repositories {
-    maven { url 'https://oss.sonatype.org/content/repositories/releases/' }
-    maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
+    repositories.removeIf { repo ->
+        repo instanceof MavenArtifactRepository && repo.url.toString().contains('oss.sonatype.org')
+    }
     maven { url "https://hyperledger.jfrog.io/artifactory/besu-maven/" }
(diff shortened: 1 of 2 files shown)


gtebrean (commented):
changelog update is missing

Dev10-sys:
> changelog update is missing

Added the changelog entry. 👍

integration-tests/build.gradle
```diff
@@ -2,8 +2,9 @@ description 'web3j integration tests'
 
 
 repositories {
-    maven { url 'https://oss.sonatype.org/content/repositories/releases/' }
-    maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
+    repositories.removeIf { repo ->
+        repo instanceof MavenArtifactRepository && repo.url.toString().contains('oss.sonatype.org')
+    }
```
gtebrean: remove this too as web3j build tools was updated, it should work now

